PR #7, opened 2019-10-29
Once A Day Frequency

jeffersonrabb:
- Introduces once-a-day frequency. 
- The transient is now an object that includes viewed count as well as the time stamp of the last viewing.
- The API call updating the transient is called when the popup is dismissed, replacing the `amp-access` pingback. This means a page view where the popup is loaded but not viewed will not count. This also means if the popup appears and the user exists without dismissing, it will also not be counted, and will probably appear on the next page view.

## Testing Instructions
- Create a new Popup, populate with content (e.g. Donate block), set Frequency to "once a day".
- Open any single page in an incognito window. The popup should display (based on delay or scroll position settings).
- Reload the page or navigate to another single page. This time the popup shouldn't appear.
- To test, you could either wait a day (!) or edit the transient.
- To edit the transient, find it in the Database with this query: `SELECT * FROM wp_options WHERE option_name LIKE '_transient_amp-%-%-popup' ORDER BY option_id DESC LIMIT 1; `. The `option_value` should look something like this `a:2:{s:5:"count";i:1;s:4:"time";i:1572383818;}`
- Use a tool like https://timestampgenerator.com/ to generate a date stamp that is over a day old, and replace the value of time in the transient.
- Reload the page and the popup should appear. Reload again and it should not.
- The logic has changed so it would be a good idea to also verify Never|Once|Every Page View frequency settings.


includes/class-newspack-popups-inserter.php
```diff
@@ -38,8 +38,6 @@ public static function popup( $content = '' ) {
 		if ( $popup ) {
 			$markup  = self::generate_popup( $popup );
 			$content = self::insert_popup( $content, $popup );
-			wp_enqueue_script( 'amp-animation' );
-			wp_enqueue_script( 'amp-position-observer' );
```
claudiulodro: Without these scripts, I don't get any popups on non-AMP pages (e.g. using AMP transitional mode).

jeffersonrabb: Fixed in  https://github.com/Automattic/newspack-popups/pull/7/commits/fff33261233a95c4a3201371baf5a70c81891b05. If the AMP plugin isn't installed at all we will need  to define the scripts  as well,  but this has proven to be a little  tricky and  I'd like to address in a separate PR.

Issue #1, opened 2019-10-28
Category Filtering


jeffersonrabb:
Add a way to show certain popups only on pages with specific taxonomies.

PR #15, opened 2019-11-03
Category Filtering

jeffersonrabb:
### All Submissions:

Adds category filtering to Pop-ups. The Pop-up CPT  now allows category selection, and these will be used to determine the rules for which pop-ups appear on which posts. 

- If a post has one or more categories assigned, the most recent Pop-up with any of those categories will be shown. 
- If no Pop-ups have the same categories as  the post, the most recent Pop-up with no categories will be shown. This is called the "Sitewide" pop-up.

Two questions:

1) To query for Pop-ups with no  categories I've used this logic:

```
$args['category__not_in'] = get_terms(
	'category',
		[
			'fields' => 'ids',
		]
);
```

This may be inefficient if there are a large number of categories. Is there a better way to structure this query?

2) In working on this, I noticed that the `the_content` hook fires twice. Is this problematic, and should anything be done to make sure the `popup()` function only executes once?

Closes https://github.com/Automattic/newspack-popups/issues/1.

1. Create a few pop-ups, some with multiple categories, some with one category, some with none.  To ease testing I suggest setting  the titles to describe their categories, and using a frequency of "Every page view."
2. In an incognito window, view  the site. Visit various pages with different categories and verify the filtering rules described above.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -135,12 +135,12 @@ final class Newspack_Popups_Inserter {
 		if ( count( $category_ids ) ) {
 			$args['category__in'] = $category_ids;
 		} else {
-			$args['category__not_in'] = get_terms(
-				'category',
+			$args['tax_query'] = [ // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_tax_query
 				[
-					'fields' => 'ids',
-				]
-			);
+					'taxonomy' => 'category',
+					'operator' => 'NOT EXISTS',
+				],
+			];
 		}
 		return self::retrieve_popup_with_query( new WP_Query( $args ) );
 	}
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -114,6 +114,7 @@ final class Newspack_Popups_Inserter {
 	/**
 	 * Retrieve popup CPT post.
 	 *
+	 * @param array $categories An array of categories to match.
 	 * @return object Popup object
 	 */
 	public static function retrieve_popup( $categories = [] ) {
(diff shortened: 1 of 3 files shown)


includes/class-newspack-popups-inserter.php
```diff
@@ -100,6 +125,22 @@ public static function retrieve_popup() {
 			'posts_per_page' => 1,
 		];
 
+		$category_ids = array_map(
+			function( $cagegory ) {
```
claudiulodro: `$cagegory` :P

includes/class-newspack-popups.php
```diff
@@ -174,9 +175,15 @@ public static function display_post_states( $post_states, $post ) {
...
-					'posts_per_page' => 1,
+					'post_type'        => self::NEWSPACK_PLUGINS_CPT,
+					'post_status'      => 'publish',
+					'posts_per_page'   => 1,
+					'category__not_in' => get_terms(
+						'category',
+						[
+							'fields' => 'ids',
+						]
+					),
```
claudiulodro: The `category__not_in` part is brutal performance-wise. This should work better IMO:
```
'tax_query' => [
    [
        'taxonomy' => 'category',
        'operator' => 'NOT EXISTS',
    ],
],
```

includes/class-newspack-popups-inserter.php
```diff
@@ -100,6 +125,22 @@ public static function retrieve_popup() {
...
+		);
+		if ( count( $category_ids ) ) {
+			$args['category__in'] = $category_ids;
+		} else {
+			$args['category__not_in'] = get_terms(
+				'category',
+				[
+					'fields' => 'ids',
+				]
+			);
```
claudiulodro: See my other comment about how to replace the `category__not_in` query.

includes/class-newspack-popups-inserter.php
```diff
@@ -33,22 +57,23 @@ public static function popup( $content = '' ) {
 		if ( is_user_logged_in() || ! is_single() ) {
 			return $content;
 		}
-		/* End */
```
claudiulodro: On line 45 you have `add_action( 'the_content', [ $this, 'popup' ] );` It should be `add_filter` instead.

claudiulodro:
> In working on this, I noticed that the the_content hook fires twice. Is this problematic, and should anything be done to make sure the popup() function only executes once?

I'm not sure exactly what you mean by this, but `the_content` is common and can potentially fire multiple times on the same post. For example, if my memory is right Yoast uses it to prep content so it can analyze it behind-the-scenes, etc. As long as someone isn't like `apply_filters( 'the_content', apply_filters( 'the_content' ), $content )`, and repeatedly applying `the_content` filter to the same piece of content it should be OK. If someone is doing something like that, all sorts of other stuff will break anyways.

includes/class-newspack-popups-inserter.php
jeffersonrabb: Oops https://github.com/Automattic/newspack-popups/pull/15/commits/59844684808b46005b3561db92f715b1a1590fa5

includes/class-newspack-popups.php
jeffersonrabb: When I try the `tax_query` approach I get the following warning: 

```
warning - Detected usage of tax_query, possible slow query. (WordPress.DB.SlowDBQuery.slow_db_query_tax_query)
```

claudiulodro: `category__not_in` is shorthand for the following tax query under the hood; it's just not caught by the linter:
```
[
    'taxonomy' => 'category',
    'operator' => 'NOT IN',
    'value' => [ 1, 2, 3, 4, 5, 6, 7, 8, 9, ... ] // Every single category on the site.
]
```

jeffersonrabb: Change made here: https://github.com/Automattic/newspack-popups/pull/15/commits/a73e69d1bee5f388e862020c0409d867a66b416a

includes/class-newspack-popups-inserter.php
jeffersonrabb: Change here: https://github.com/Automattic/newspack-popups/pull/15/commits/a73e69d1bee5f388e862020c0409d867a66b416a

jeffersonrabb:
Thanks for digging in on this one!

PR #28, opened 2019-12-06
Popups  On Non-Posts

jeffersonrabb:
### All Submissions:

This PR places Timed Pop-ups on non-posts, including Homepage, Pages, Category Archives, Author Archives, and Search Results. The hook used to place on Archive pages is specific to Newspack Theme, so for the moment, popups  will only appear on those pages if Newspack is the active theme. The  new logic applies only to Pop-ups with a timer trigger. Scroll triggers only work in posts, because the trigger point placement is calculated based on normal paragraph-based content.

1. With Newspack as the theme, create  a Pop-up, set Trigger: Timer, Delay: 1s, Frequency: Every Page View. 
2. In an incognito window, view homepage,  another page, a  category archive, an author archive, search results, and a post. The popup should appear on all of these.
3. Change the popup's Trigger to "Scroll Progress." Test on all of the above  again. It should appear only on the post.
4, Change theme to something other than Newspack, and set  the Trigger back to Timer. Re-check the above. Only the Post should work.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -43,12 +43,12 @@ final class Newspack_Popups_Inserter {
 	 */
 	public function __construct() {
 		add_filter( 'the_content', [ $this, 'popup' ] );
-		add_action( 'wp_head', [ __CLASS__, 'popup_access' ] );
+		add_action( 'after_header', [ $this, 'popup_after_header' ] ); // This is a Newspack theme hook. When used with other themes, popups won't be inserted on archive pages.
 		add_action( 'wp_head', [ __CLASS__, 'popup_access' ] );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


includes/class-newspack-popups-inserter.php
```diff
@@ -76,6 +81,29 @@ public static function popup( $content = '' ) {
 		return $content;
 	}
 
+	/**
+	 * Process popup and insert into archive pages if needed. Applies to Newspack Theme only.
+	 */
+	public static function popup_after_header() {
+		/* Posts and pages are covered by the_content hook */
+		if ( is_user_logged_in() || is_single() || is_page() ) {
```
claudiulodro: Just a fun tip: [is_singular](https://developer.wordpress.org/reference/functions/is_singular/) is an easier way of doing `is_single() || is_page()`

PR #34, opened 2020-01-15
Integration With Newspack Plugin

jeffersonrabb:
### All Submissions:

Integration points for the Newspack plugin Engagement Wizard. Test with https://github.com/Automattic/newspack-plugin/pull/365.

Complete testing instructions in https://github.com/Automattic/newspack-plugin/pull/365.
### Other information:

### DIFF
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -72,8 +72,9 @@ final class Newspack_Popups_Model {
 	 * @return object Popup object
 	 */
 	protected static function retrieve_popup_with_query( WP_Query $query ) {
-		$popup = null;
+		$popups = [];
 		while ( $query->have_posts() ) {
+			$popup = null;
 			$query->the_post();
 			$blocks = parse_blocks( get_the_content() );
 			$body   = '';
@@ -123,9 +124,17 @@ final class Newspack_Popups_Model {
 				$popup['options']['placement'] = 'center';
 			}
 			$popup['markup'] = self::generate_popup( $popup );
+
+			$popups[] = $popup;
 		}
 		wp_reset_postdata();
-		return $popup;
+		if ( 1 === count( $popups ) ) {
+			return $popups[0];
+		}
+		if ( count( $popups ) > 1 ) {
+			return $popups;
+		}
+		return null;
 	}
 
 	/**
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -25,11 +25,6 @@ final class Newspack_Popups_Model {
 		];
 
 		$popups = self::retrieve_popup_with_query( new WP_Query( $args ), true );
-
-		// Hacky check if this is a single popup object as opposed to an array of Pop-ups. TODO: replace.
-		if ( isset( $popups['id'] ) ) {
-			$popups = [ $popups ];
-		}
 		foreach ( $popups as &$popup ) {
 			if ( ! count( $popup['categories'] ) ) {
 				$popup['sitewide_default'] = true;
@@ -100,8 +95,6 @@ final class Newspack_Popups_Model {
 	 * @return object Popup object
 	 */
 	public static function retrieve_popup( $categories = [] ) {
-		$popup = null;
-
 		$args = [
 			'post_type'      => Newspack_Popups::NEWSPACK_PLUGINS_CPT,
 			'post_status'    => 'publish',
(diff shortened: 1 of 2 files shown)


includes/class-newspack-popups-model.php
```diff
@@ -12,6 +12,87 @@
...
+		$args = [
+			'post_type'      => Newspack_Popups::NEWSPACK_PLUGINS_CPT,
+			'post_status'    => 'publish',
+			'posts_per_page' => 100,
+		];
+
+		$popups = self::retrieve_popup_with_query( new WP_Query( $args ), true );
+
+		// Hacky check if this is a single popup object as opposed to an array of Pop-ups. TODO: replace.
+		if ( isset( $popups['id'] ) ) {
```
claudiulodro:
```suggestion	
		if ( $popups && ! is_array( $popups ) ) {
```

Here's the answer to your TODO :)

includes/class-newspack-popups-model.php
```diff
@@ -69,11 +150,13 @@ public static function retrieve_popup_by_id( $post_id ) {
...
 	 * @param WP_Query $query The query to use.
+	 * @param boolean  $include_categories If true, returned objects will include assigned categories.
 	 * @return object Popup object
 	 */
-	protected static function retrieve_popup_with_query( WP_Query $query ) {
-		$popup = null;
+	protected static function retrieve_popup_with_query( WP_Query $query, $include_categories = false ) {
+		$popups = [];
 		while ( $query->have_posts() ) {
+			$popup = null;
```
claudiulodro: I don't believe this line is needed at all?

includes/class-newspack-popups-model.php
```diff
@@ -12,6 +12,87 @@
...
+			'posts_per_page' => 100,
+		];
+
+		$popups = self::retrieve_popup_with_query( new WP_Query( $args ), true );
+
+		// Hacky check if this is a single popup object as opposed to an array of Pop-ups. TODO: replace.
+		if ( isset( $popups['id'] ) ) {
+			$popups = [ $popups ];
+		}
+		foreach ( $popups as &$popup ) {
```
claudiulodro: If I'm reading things correctly, it looks like `retrieve_popup_with_query` will return `null` if no popups exist. If so, this should not try and `foreach ( null )` because it will throw a notice. 

You could just put a `if ( empty( $popups ) ) { return []; }` somewhere above and it won't be an issue.

includes/class-newspack-popups-model.php
```diff
@@ -12,6 +12,87 @@
...
+		}
+		$time = current_time( 'mysql' );
+		wp_update_post(
+			[
+				'ID'            => $id,
+				'post_date'     => $time,
+				'post_date_gmt' => get_gmt_from_date( $time ),
+			]
+		);
+	}
```
claudiulodro: I just realized this sets the sitewide popup by making it the newest popup. I think that is super clever. 

If someone sets the sitewide default and then creates a new popup with no category, the new popup will take over as the sitewide default. Is this the intended behavior? (Both ways makes sense IMO, so just a heads-up)

includes/class-newspack-popups-model.php
jeffersonrabb: > Here's the answer to your TODO :)

I believe `is_array()` will return true if it's an associative array, so that won't catch this case.

includes/class-newspack-popups-model.php
jeffersonrabb: I reworked this a bit in https://github.com/Automattic/newspack-popups/pull/34/commits/a7ca443afb45b858255fdfccb81b4348e60ab376 so that `retrieve_popup_with_query()` will always return an array.

includes/class-newspack-popups-model.php
jeffersonrabb: Removed in https://github.com/Automattic/newspack-popups/pull/34/commits/8276805f65318a9e971857b68e43852817ca60ed

includes/class-newspack-popups-model.php
jeffersonrabb: The new logic in https://github.com/Automattic/newspack-popups/pull/34/commits/a7ca443afb45b858255fdfccb81b4348e60ab376 makes this bit unnecessary anyway...

includes/class-newspack-popups-model.php
jeffersonrabb: Yes, I think this behavior is OK. It might  be a  little unexpected, but it would be weirder if sitewide default was the newest Pop-up unless you had used the selection UI in the Newspack Plugin. That feels less predictable. Do you think we should iterate on this logic and try to improve it?

claudiulodro: No let's go with this first and see if anyone gets confused. I think it'll be fine.

jeffersonrabb:
Thanks for reviewing @claudiulodro!

PR #36, opened 2020-01-20
Add popup preview in the editor

### LINKED ISSUE #2: Audition Popup
Add feature to allow auditioning the current popup: a button that causes the popup as it currently exists to appear in an overlay for assessment.

adekbadek:
### All Submissions:

This adds a preview button which enables previewing a popup in the editor:

Screen Capture on 2020-01-27 at 12-29-59

Closes #2 

1. open popup editor
2. click "Preview" in sidebar's Pop-up Settings
3. a preview should be displayed

### Other information:


includes/class-newspack-popups.php
```diff
@@ -14,6 +14,8 @@ final class Newspack_Popups {
 
 	const NEWSPACK_PLUGINS_CPT = 'newspack_popups_cpt';
 
+	const NEWSPACK_POPUP_PREVIEW_QUERY_PARAM = 'newspack_popup_preview_id';
```
jeffersonrabb: I tend to prefix plugin options with the full slug of the plugin, so this would be `newspack_popups_preview_id`.

includes/class-newspack-popups.php
```diff
@@ -261,5 +269,14 @@ public static function display_post_states( $post_states, $post ) {
...
 		return $post_states;
 	}
+
+	/**
+	 * Get previewed popup id from the URL.
+	 *
+	 * @return number|null Popup id, if found in the URL
+	 */
+	public static function previewed_popup_id() {
+		return isset($_REQUEST[ Newspack_Popups::NEWSPACK_POPUP_PREVIEW_QUERY_PARAM ]) ? $_REQUEST[ Newspack_Popups::NEWSPACK_POPUP_PREVIEW_QUERY_PARAM ] : NULL;
```
jeffersonrabb: This line raises a bunch of PHPCS warnings and errors:

```
phpcs: error
error - Must use "self::" for local static member reference (Squiz.Classes.SelfMemberReference.NotUsed)

phpcs: error
error - Must use "self::" for local static member reference (Squiz.Classes.SelfMemberReference.NotUsed)

phpcs: error
error - TRUE, FALSE and NULL must be lowercase; expected "null" but found "NULL" (Generic.PHP.LowerCaseConstant.Found)

phpcs: warning
warning - Processing form data without nonce verification. (WordPress.Security.NonceVerification.Recommended)

phpcs: warning
warning - Processing form data without nonce verification. (WordPress.Security.NonceVerification.Recommended)

phpcs: warning
warning - Detected usage of a non-sanitized input variable: $_REQUEST[Newspack_Popups::NEWSPACK_POPUP_PREVIEW_QUERY_PARAM] (WordPress.Security.ValidatedSanitizedInput.InputNotSanitized)
```

Suggested alternative:

```suggestion
		return filter_input( INPUT_GET, self::NEWSPACK_POPUP_PREVIEW_QUERY_PARAM, FILTER_SANITIZE_STRING );
```

src/editor/index.js
```diff
@@ -37,9 +43,11 @@ class PopupSidebar extends Component {
 			trigger_delay,
 			trigger_type,
 			utm_suppression,
+			currentPost,
```
jeffersonrabb: Where is `currentPost` used?

src/editor/PopupPreview.js
```diff
@@ -0,0 +1,62 @@
...
+import { compose } from '@wordpress/compose';
+import { __ } from '@wordpress/i18n';
+
+/**
+ * External dependencies
+ */
+import { WebPreview } from 'newspack-components';
+
+const PopupPreviewSetting = ( {
+	content,
```
jeffersonrabb: Looks like `content`, `options`, and `embedPreview` are no longer in use?

src/editor/PopupPreview.js
```diff
@@ -0,0 +1,62 @@
...
+	isSavingPost,
+	embedPreview,
+	postId,
+} ) => {
+	const url = `/?newspack_popup_preview_id=${ postId }`;
+
+	return (
+		<WebPreview
+			url={ url }
+			isPrimary
```
jeffersonrabb: `isPrimary` has no meaning if `renderButton` is set.

thomasguillot:
The UI feels a bit strange to me.

Posts/Pages have a "Preview" button at the top. I think we should do the same here and keep it consistent.

Screenshot 2020-01-27 at 15 24 48

Having Newspack Components within the Editor is a bit odd and I would stick to Gutenberg.

jeffersonrabb:
> Having Newspack Components within the Editor is a bit odd and I would stick to Gutenberg.

Is there a Gutenberg equivalent for previewing a site in an iFrame?

src/editor/index.js
adekbadek: it's not, I'm so used to ESlint that I don't see unused variables by myself anymore 😬

thomasguillot:
> Is there a Gutenberg equivalent for previewing a site in an iFrame?

Not that I'm aware of.

If you look at home Preview works for Posts and Pages, it opens up in a new tab.

jeffersonrabb:
> Not that I'm aware of.

Me neither. Wouldn't that be an argument for using the Newspack component in this one case? Feels to me like a different case than if we were using Newspack components in the sidebar intermingled with Gutenberg ones. As a full-screen overlay this feels like a separate piece of UI that doesn't necessarily create dissonance.

adekbadek:
@jeffersonrabb : 
1. I've fixed the PHP formatting issues, sorry for not checking this before pushing
2. Had to add a class to the popup container to handle admin bar - the body class is there even if after `show_admin_bar( false )`: 1b901ec5b1d22e418e89f41a6a8daba4416c54bd
3. I've tweaked the `amp-access` bit like so: abbdebdd61408d0cb322d34c3d6f2388455975af

@thomasguillot : 
1. Gutenberg does not expose a way to insert a button next to "Publish…"/"Update", so this would be very hacky :/ I agree the placement of the button now is not perfect, but I'm not sure it's worth it
2. About using NP components in Gutenberg, I agree with Jeff here:
> As a full-screen overlay this feels like a separate piece of UI that doesn't necessarily create dissonance.

As a much smaller thing than posts, a popup is readable in a modal - opening a new tab by default seems like an overkill for me. Though that is certainly an option.

---

I've found another problem: when editing a published popup, triggering the preview will **save** it. This should not happen. The solution would be to use "autosaving", but this does not handle custom fields. I'll have to dig into that.

adekbadek:
> I suppose it's not possible to move the Preview button to the "Status & visibility" section either?

[Now that is possible](https://developer.wordpress.org/block-editor/developers/slotfills/plugin-post-status-info/) - a component can be placed right above the "Move to trash" link

thomasguillot:
> Now that is possible - a component can be placed right above the "Move to trash" link

Oh cool!

If we move it above the "Move to trash" I'd suggest we use an `isLink`.
Screenshot 2020-01-28 at 08 52 48

But below, an `isPrimary` is fine
Screenshot 2020-01-28 at 08 57 02

adekbadek:
@thomasguillot 👌 that's how the button is placed now: 

image

@jeffersonrabb As I [commented](https://github.com/Automattic/newspack-popups/pull/36#issuecomment-578907854), there was an issue with saving the post before displaying the preview. The solution was to autosave, but autosaving does not handle custom fields. I've worked around this by appending the options to the URL and reading them from there if it's a preview: c20fc4764833ee22f8f3476a6ca72ab967e583b8

One PHPCS warning I've suppressed was `WordPress.Security.NonceVerification.Recommended` here:

https://github.com/Automattic/newspack-popups/blob/4e1318212acd25db4c6a76f0537e4e7d5e70befa/includes/class-newspack-popups-model.php#L135-L140

I'm not sure of the security implication of this workaround.

src/editor/PopupPreview.js
```diff
@@ -0,0 +1,55 @@
...
+		// Autosave does not handle meta fields, so these will be passed in the URL
+		...metaFields,
+	} );
+
+	return (
+		<WebPreview
+			url={ `/?${ query }` }
+			renderButton={ ( { showPreview } ) => (
+				<Button
+					isLink
```
jeffersonrabb: When the autosave is in progress this button gets the in-progress state of the Update button, which looks peculiar on a link. I'd suggest changing it to `isPrimary` as discussed in https://github.com/Automattic/newspack-popups/pull/36#issuecomment-579145740. Is it possible to place it below `Move to Trash` as Thomas suggested or are we locked in to the current position?

src/view/style.scss
```diff
@@ -18,13 +18,18 @@ $color__secondary-variation: darken( $color__secondary, 10% );
...
 	z-index: 99999;
 
-	.admin-bar & {
-		@media only screen and (min-width: 601px) {
-			padding-top: 46px !important;
-		}
+	// In preview, there will be no admin bar. Unfortunately, the body classes seem to
+	// be added before the admin bar is rendered or not, so 'admin-bar' class
+	// will be there despite the actual bar not being rendered
+	&:not( &--preview ) {
```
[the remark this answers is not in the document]
adekbadek: What this is doing is preventing such state (when popup is top-aligned):

image

So it looks like this:

image

It appears that this margin applied to the `html` selector is yet another mean to accommodate the admin bar - I've un-margined it here: 0dba4abae0167f854c1fa98995d2706ee08ebd11

includes/class-newspack-popups-api.php
```diff
@@ -99,9 +104,31 @@ public function reader_get_endpoint( $request ) {
...
 		return rest_ensure_response( $response );
 	}
 
+	/**
+	 * Detect a popup preview request.
+	 *
+	 * @param  WP_REST_Request $request a request.
+	 * @return Boolean
+	 */
+	public function is_preview_request( $request ) {
```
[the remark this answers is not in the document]
adekbadek: Almost - `is_preview_request` handles the referer's URL while `previewed_popup_id` checks the current request. I've refactored it a bit, but now `previewed_popup_id` looks odd, with the two ways of query variable retrieval: 4d2d68c0fa8e4c979beb0cf63efc89cc3219e16a 🤔

src/editor/PopupPreview.js
adekbadek: That's true that the in-progress is odd… 

Screen Capture on 2020-01-30 at 16-42-28

How about keeping this as a link, but dimming on save? Otherwise, we can go back to using a `Button`. WDYT @thomasguillot ?

---

Different placement is not possible, we're locked in to this space:

Edit Pop-up ‹ newspack test — WordPress 2020-01-30 16-40-28

jeffersonrabb:
Here's an alternative approach to suppressing the admin bar: https://developer.wordpress.org/reference/hooks/show_admin_bar/. Since `previewed_popup_id()` is already based on the query param, it could be used to determine the visibility of the admin bar. 

- Around https://github.com/Automattic/newspack-popups/blob/24c68b340276f56da7ec9d3a0ea41e40f56eb324/includes/class-newspack-popups.php#L54 add:
```
add_filter( 'show_admin_bar', [ __CLASS__, 'hide_admin_bar_for_preview' ], 10, 2 );
```
- Add this function to the class:
```
public static function hide_admin_bar_for_preview() {
	return ! self::previewed_popup_id();
}
```
- Then you should be able to get rid of the special preview class and styling and 
```
show_admin_bar( false );
```

adekbadek:
That works, thanks! (5faca79396b5f07d89e82097a363cdf9844bb0e2)

includes/class-newspack-popups-model.php
```diff
@@ -222,6 +269,11 @@ public static function generate_popup( $popup ) {
 		$overlay_opacity = absint( $popup['options']['overlay_opacity'] ) / 100;
 		$overlay_color   = $popup['options']['overlay_color'];
 
+		// Add a class to indicate a preview.
+		if ( Newspack_Popups::previewed_popup_id() ) {
+			array_push( $classes, 'newspack-lightbox--preview' );
```
jeffersonrabb: With the non-CSS approach to admin bar suppression, is this still needed?

adekbadek: Nope, fixed: d1f90f6e827188218c54d6f05ebcca703eaf43a2

src/editor/PopupPreview.js
thomasguillot: Let's use a primary button.

It's ok if we can't move it underneath.

adekbadek: Updated: 80235dc

Issue #47, opened 2020-01-23
Sitewide Default Logic Ignores "Never" Frequency


jeffersonrabb:
The logic that determines which Popup to use as Sitewide Default doesn't take into account the possibility of a Frequency of "Never." If there are multiple candidates and the first in line is set to Never show, there will be no Pop-ups.

Issue #48, opened 2020-01-23
Display Logic Based on Referrer


jeffersonrabb:
Option to condition Pop-up display logic on the referring website, e.g. facebook.com or twitter.com. Customers may wish to show specific pop-ups to user arriving from these sites, or conversely may wish to show no pop-up under these conditions.

adekbadek:
Closed by #323

PR #50, opened 2020-02-02
Refactor Sitewide Default

jeffersonrabb:
### All Submissions:

"Sitewide Default" refers to the Pop-up that is shown on any page, assuming there is no better match based on category filters. Currently this will be the newest published Pop-up with no category assignments. This logic is awkward:

- Sometimes it is unclear why a particular Pop-up has this status. 
- Setting a different sitewide default requires changing the published date of the post, which is a bit janky. 
- As the display logic becomes richer (e.g. [referrer filters](https://github.com/Automattic/newspack-popups/issues/48)) this approach will quickly become untenable. 
- The current approach is effectively a blocker on https://github.com/Automattic/newspack-plugin/pull/409, and this change frees that work to proceed.

This PR clarifies this logic by storing the ID of the sitewide default Pop-up in a single option, rather than relying on post newness and category assignment. The editor explicitly picks which Pop-up is the sitewide default, which is a much more understandable and deliberate approach.

At the top of the Pop-up Settings panel there is a Sitewide Default checkbox. Checking this causes the current Pop-up to immediately become sitewide default, replacing any other that might have been previously. Unchecking will cause there to be no sitewide default at all. 

Screen Shot 2020-02-02 at 6 11 04 PM

Closes https://github.com/Automattic/newspack-popups/issues/47

1. Create at least three Pop-ups. Two should have no category assignments. Set the Trigger to Timer with a short Delay. Frequency should be Test Mode.
2. Edit one of the category-less ones, and check Sitewide Default. 
3. Return to the Pop-ups post list page, and observe that this post now has "Sitewide Default" after it's name. 
4. In a normal (not incognito) window, open the homepage of the site. Observe the Sitewide Default Pop-up appears.
5. Edit the other category-less one, and check Sitewide Default.
6. On the Pop-ups list page the new one should say Sitewide Default.
7. Refresh the homepage and you should be shown the new one.
8. Visit a page with a category matching the category chosen in the third Pop-up. Observe that this Pop-up is shown.
9. Edit the current sitewide default and uncheck the checkbox. 
10. Return to the Pop-ups list and observe no Pop-ups have Sitewide Default text.
11. Visit the homepage and observe that no Pop-up is shown.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -47,6 +47,9 @@ final class Newspack_Popups_Inserter {
 		$sitewide_default = get_option( Newspack_Popups::NEWSPACK_POPUPS_SITEWIDE_DEFAULT, null );
 		if ( $sitewide_default ) {
 			self::$popup = Newspack_Popups_Model::retrieve_popup_by_id( $sitewide_default );
+			if ( self::$popup ) {
+				return self::$popup;
+			}
 		}
 		return null;
 	}
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -35,13 +35,20 @@ final class Newspack_Popups_Inserter {
 		}
 
 		// Then try for pop-up with category filtering.
-		self::$popup = Newspack_Popups_Model::retrieve_popup( get_the_category() );
+		$categories = get_the_category();
+		if ( $categories && count( $categories ) ) {
+			self::$popup = Newspack_Popups_Model::retrieve_popup( get_the_category() );
+			if ( self::$popup ) {
+				return self::$popup;
+			}
+		}
 
 		// If nothing found, try for sitewide pop-up.
-		if ( ! self::$popup ) {
-			self::$popup = Newspack_Popups_Model::retrieve_popup();
+		$sitewide_default = get_option( Newspack_Popups::NEWSPACK_POPUPS_SITEWIDE_DEFAULT, null );
+		if ( $sitewide_default ) {
+			self::$popup = Newspack_Popups_Model::retrieve_popup_by_id( $sitewide_default );
 		}
-		return self::$popup;
+		return null;
 	}
 
 	/**
--- a/src/editor/index.js
+++ b/src/editor/index.js
@@ -31,6 +31,25 @@ import { optionsFieldsSelector } from './utils';
 import PopupPreview from './PopupPreview';
 
 class PopupSidebar extends Component {
+	state = {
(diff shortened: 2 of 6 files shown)


includes/class-newspack-popups-inserter.php
```diff
@@ -28,14 +28,22 @@ public static function popup_for_post() {
...
+				return self::$popup;
+			}
+		}
 
 		// If nothing found, try for sitewide pop-up.
-		if ( ! self::$popup ) {
-			self::$popup = Newspack_Popups_Model::retrieve_popup();
+		$sitewide_default = get_option( Newspack_Popups::NEWSPACK_POPUPS_SITEWIDE_DEFAULT, null );
+		if ( $sitewide_default ) {
+			self::$popup = Newspack_Popups_Model::retrieve_popup_by_id( $sitewide_default );
```
adekbadek: Missing return statement:
```suggestion
			self::$popup = Newspack_Popups_Model::retrieve_popup_by_id( $sitewide_default );
			if ( self::$popup ) {
				return self::$popup;
			}
```

src/editor/index.js
```diff
@@ -29,17 +31,32 @@ class PopupSidebar extends Component {
...
+					checked={ newspack_popups_is_sitewide_default }
+					onChange={ value => {
+						const params = {
+							path: '/newspack-popups/v1/sitewide_default/' + id,
+							method: value ? 'POST' : 'DELETE',
+						};
+						onSitewideDefaultChange( value );
+						apiFetch( params );
+					} }
+				/>
```
adekbadek: It feels a inconsistent that this input updates the popup settings _immediately_, while other inputs just change the "edited" state of the popup (awaiting publish). This will also result in a following scenario being possible:
1. user adds new popup
1. they fill in title, content, check "Sitewide Default", and publish
1. the popup is **not** Sitewide Default though, because at the time of checking the checkbox, the post was not existing yet

Maybe the option update logic could be handled [on post save (`didPostSaveRequestSucceed`)](https://developer.wordpress.org/block-editor/data/data-core-editor/#didPostSaveRequestSucceed)?

includes/class-newspack-popups-inserter.php
jeffersonrabb: Fixed in https://github.com/Automattic/newspack-popups/pull/50/commits/a4eb444c015aee7c61218384660fa82c5d5f2926

jeffersonrabb:
@adekbadek this should be ready for review again. Your alternate approach to updating the option on Post save is merged in, and I resolved the conflicts with `master`.

PR #53, opened 2020-02-03
save sitewide_default on component update

adekbadek:
Problem: the `newspack_popups_is_sitewide_default` on a post should be updated on save, like a meta field.  
How it works here - on save:
- the request is sent to `/sitewide_default/:id` endpoint
- post is saved

But `newspack_popups_is_sitewide_default` value in post save payload is ignored, so the response has stale value of `newspack_popups_is_sitewide_default`. Before, post save always happened after `/sitewide_default/:id` request has completed, so there was not such problem.


jeffersonrabb:
I tried a few other approaches but couldn't find a cleaner one, so I'm going to merge this in...

PR #59, opened 2020-02-12
Dont inject popups into feeds

claudiulodro:
### All Submissions:

This PR fixes an issue where popup content was getting injected into the feeds, so the first line of each item in a site's RSS feed would be the popup content.

1. Before applying this patch, create a popup:
Screen Shot 2020-02-12 at 2 21 01 PM

2. Visit `/feed` on your site. Observe the pop up content is in the description for each item:
Screen Shot 2020-02-12 at 2 21 10 PM

3. Apply this patch. Refresh. Observe the pop up content is not in the description for each item:
Screen Shot 2020-02-12 at 2 22 44 PM

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -70,7 +70,7 @@ final class Newspack_Popups_Inserter {
 	 * @return string The content with popup inserted.
 	 */
 	public static function popup( $content = '' ) {
-		if ( is_admin() ) {
+		if ( is_admin() || is_feed() ) {
 			return $content;
 		}
 		$popup = self::popup_for_post();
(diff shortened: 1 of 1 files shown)


includes/class-newspack-popups-inserter.php
```diff
@@ -70,7 +70,7 @@ public function __construct() {
 	 * @return string The content with popup inserted.
 	 */
 	public static function popup( $content = '' ) {
-		if ( is_admin() ) {
+		if ( is_admin() || is_feed() ) {
```
philipjohn: Is this hook only supposed to run for popups? If so, I'd recommend making this conditional check that the current post is of the right post type. E.g.:

```
if ( Newspack_Popups::NEWSPACK_PLUGINS_CPT !== get_post_type() ) {
    return;
}
```

claudiulodro: No, it injects the popup into post content.

claudiulodro:
Let me add a couple tweaks to prevent popups showing on all excerpts too before reviewing.

claudiulodro:
I've changed the logic a little so it will prevent popups from getting injected into excerpts on feeds and archives.

Issue #65, opened 2020-02-21
Inline Placement


jeffersonrabb:
Add support for inline placement of Pop-ups. This will cause the pop-up content to be automatically inserted into post content. A control allows determining how far into the content it should be placed. 

Bonus: add a Pop-up block to individually place Pop-up content directly into specific Posts.

PR #69, opened 2020-02-24
Inline Placement

jeffersonrabb:
### All Submissions:

This PR creates a new placement option: inline. When selected, the content is inserted inline in posts as opposed to being displayed in an overlay. For inline pop-ups the `trigger_scroll_progress` field is repurposed to determine the placement of the content.  

Unresolved issues (now resolved):

- [x] For overlays, clicking the close button is the event that records the campaign has been viewed. Inline campaigns don't need a close button, so we'll need an alternate approach.
- [x] How should inline campaigns be styled? cc: @thomasguillot 
- [x] Should inline campaigns use the same visibility rules as overlays (sitewide default, categories, etc)? 
- [x] Does it make sense to treat "inline" as a placement option? If this doesn't feel like a logical fit, there could be a new "type" control with options `overlay|inline`. Placement options would be shown only when overlay is selected.

Closes https://github.com/Automattic/newspack-popups/issues/65.

Screen Shot 2020-03-09 at 4 57 11 PM

### Testing Instructions

- Create a new Pop-up, add Mailchimp block. Set Placement to "Inline," and Frequency to "Test Mode."
- Preview the block, verify it appears inline. Note that the Preview for inline blocks will be the most recent Post, since Inline Pop-ups cannot be inserted on the Homepage.
- Try different Position settings. Note that the Pop-up will be placed immediately after the paragraph that appears closest to the position specified, so setting the value to 1% should place the block just after the first paragraph.
- Change Frequency to "Once."
- View a post in Incognito mode. Verify the Pop-up appears.
- Go to another post, and verify the Pop-up does not appear. 

After successful Mailchimp subscription, the Pop-up should never be shown again to the same user. This is difficult to test, though, since it would require waiting 24 hours and trying again. Here's how it can be tested:

- Clear the Database of all popup transients by executing this query: `DELETE FROM wp_options WHERE option_name LIKE '%amp%-popup'`
- View the Pop-up incognito, subscribe successfully.
- Query for the transient: `SELECT * FROM wp_options WHERE option_name LIKE '%amp%-popup'`
- Verify the `option_value` contains `mailing_list_status` with a value of 1.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-api.php
+++ b/includes/class-newspack-popups-api.php
@@ -229,6 +229,11 @@ final class Newspack_Popups_API {
 		}
 		$popup_id = isset( $request['popup_id'] ) ? $request['popup_id'] : false;
 		$url      = isset( $request['url'] ) ? esc_url_raw( urldecode( $request['url'] ) ) : false;
+		if ( ! $popup_id && ! $url ) {
+			$body     = json_decode( $request->get_body(), true );
+			$popup_id = isset( $body['popup_id'] ) ? $body['popup_id'] : false;
+			$url      = isset( $body['url'] ) ? esc_url_raw( urldecode( $body['url'] ) ) : false;
+		}
 		if ( $reader_id && $url && $popup_id ) {
 			return $reader_id . '-' . $popup_id . '-popup';
 		}
--- a/includes/class-newspack-popups-api.php
+++ b/includes/class-newspack-popups-api.php
@@ -188,7 +188,7 @@ final class Newspack_Popups_API {
 			if ( $request['suppress_forever'] ) {
 				$data['suppress_forever'] = true;
 			}
-			if ( $request['mailing_list_status'] && 'subscribed' === $request['mailing_list_status'] ) {
+			if ( $this->get_mailing_list_status( $request ) ) {
 				$data['mailing_list_status'] = true;
 			}
 			set_transient( $transient_name, $data, 0 );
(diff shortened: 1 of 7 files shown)


thomasguillot:
> Does it make sense to treat "inline" as a placement option? If this doesn't feel like a logical fit, there could be a new "type" control with options overlay|inline. Placement options would be shown only when overlay is selected.

I really don't mind. At least we see straight away all the options: center - top - bottom - inline.

I do think that the sidebar settings is getting very busy and hard to understand. I suggest that we reorganise the settings a bit, something like: 
Screenshot 2020-02-24 at 13 36 20

thomasguillot:
This inline placement also makes me wonder if we should rename "Pop-ups" to something else...

thomasguillot:
> For overlays, clicking the close button is the event that records the campaign has been viewed. Inline campaigns don't need a close button, so we'll need an alternate approach.

I'd actually suggest that we keep a close button, maybe smaller.

thomasguillot:
> How should inline campaigns be styled?

I think the only differentiation that we would need is when the background of the "pop-up" matches the background of the body: How do we differentiate the "pop-up" from the content? It could be with a 1px border around the `<div>`, based on the text colour (e.g. `rgba( text-color, 25%` ).

thomasguillot:
> Should inline campaigns use the same visibility rules as overlays (sitewide default, categories, etc)?

I'd say so.

We could also offer a "Pop-up Inline" block. To offer more placement control. It would then overwrite the visibility rules.

jeffersonrabb:
> This inline placement also makes me wonder if we should rename "Pop-ups" to something else...

I was wondering about "Campaigns" - What do you think?

thomasguillot:
> I was wondering about "Campaigns" - What do you think?

I like it but it feels quite specific. Using Campaigns for a donations block for example seem odd.

What about "Announcements". It's vague enough so it can be a pop-up, something inline, and still be tracked.

jeffersonrabb:
> I think the only differentiation that we would need is when the background of the "pop-up" matches the background of the body: How do we differentiate the "pop-up" from the content? It could be with a 1px border around the `<div>`, based on the text colour (e.g. `rgba( text-color, 25%` ).

Would it be bad to add this border always? It would be much easier than adding more color comparison logic, although that's certainly possible if needed.

thomasguillot:
> Would it be bad to add this border always?

No that's fine 😊

PR #94, opened 2020-03-31
fix: popup insertion logic

### LINKED ISSUE #92: Misplaced inline popups
The popup placement logic inserts the popup markup after a closing `p` tag. This can lead to unfortunate placement of the popup, such as here, where the popup markup was placed inside of a `blockquote`:

image

The placement logic should only be concerned with `p` elements which are direct descendants of `.entry-content`.

adekbadek:
### All Submissions:

By matching block end comments instead of paragraph end tags, the popup will always be inserted between blocks, and not inside of a block (which might have paragraphs inside).

Resolves #92

1. On `master`
1. Add a quote block to a post
2. Add an inline popup and try to set the "Approximate position" so that the popup placed is where the quote is
2. Observe in preview that the inline popup in inserted into the quote block
3. Switch to this branch
4. Observe in preview that the inline popup and quote block are not conjoined

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -184,23 +184,21 @@ final class Newspack_Popups_Inserter {
 		$total_length     = strlen( $content );
 		$precise_position = $total_length * $percentage;
 
-		$blocks = [];
-		// Match block closing comment.
-		// hyphens have to be escaped for some PHP versions: https://stackoverflow.com/a/55606868/3772847.
-		$block_close_pattern = '/<!\-\- \/wp:[\w\-\/]* \-\->/';
-		preg_match_all( $block_close_pattern, $content, $blocks, PREG_OFFSET_CAPTURE );
-		$insertion_position = $total_length;
-		foreach ( $blocks[0] as $index => $block ) {
-			$block_offest = $block[1];
-			if ( $block_offest >= $precise_position ) {
-				$insertion_position = $block_offest;
-				break;
+		$blocks      = parse_blocks( $content );
+		$pos         = 0;
+		$output      = '';
+		$is_inserted = false;
+		foreach ( $blocks as $block ) {
+			$block_content = render_block( $block );
+			$pos          += strlen( $block_content );
+			if ( ! $is_inserted && $pos >= $precise_position ) {
+				$output     .= '<!-- wp:shortcode -->' . $popup_markup . '<!-- /wp:shortcode -->';
+				$is_inserted = true;
 			}
+			$output .= $block_content;
 		}
 
-		$before_popup = substr( $content, 0, $insertion_position );
-		$after_popup  = substr( $content, $insertion_position );
-		return $before_popup . $popup_markup . $after_popup;
+		return $output;
 	}
 
 	/**
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -174,16 +174,14 @@ final class Newspack_Popups_Inserter {
 
 		$blocks = [];
 		// Match block closing comment.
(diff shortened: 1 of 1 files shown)


adekbadek:
@claudiulodro I've followed you suggestion: 77911ff720ba318b333fbb037b55827950135deb

includes/class-newspack-popups-inserter.php
```diff
@@ -180,27 +180,25 @@ public static function insert_popup( $content = '', $popup = [] ) {
...
+		$output      = '';
+		$is_inserted = false;
+		foreach ( $blocks as $block ) {
+			$block_content = render_block( $block );
+			$pos          += strlen( $block_content );
+			if ( ! $is_inserted && $pos >= $precise_position ) {
+				$output     .= '<!-- wp:shortcode -->' . $popup_markup . '<!-- /wp:shortcode -->';
+				$is_inserted = true;
 			}
+			$output .= $block_content;
```
claudiulodro, line 198: A site with this plugin active will have its blocks rendered at priority 1 in `the_content`. I'm not sure whether this will cause issues from other plugins that assume things, but I think there is definitely the potential for some weirdness. For example, a plugin that does something with raw block content and hooks into `the_content` at priority 4 will not have the raw block content it expects. The unknown unknowns here are my main concern. 

An approach that only inserts the shortcode block at the appropriate place and doesn't otherwise modify the content would solve this issue, e.g.
```
		$blocks      = parse_blocks( $content );
		$pos         = 0;
		$output      = '';
		$is_inserted = false;
		foreach ( $blocks as $block ) {
             $original_block_markup = '?????'; // Need to figure this out somehow I think.
			$block_content = render_block( $block );
			$pos          += strlen( $block_content );
			if ( ! $is_inserted && $pos >= $precise_position ) {
				$output     .= '<!-- wp:shortcode -->' . $popup_markup . '<!-- /wp:shortcode -->';
				$is_inserted = true;
			}
			$output .= $original_block_markup; // Let the block parser handle the block rendering at `the_content` priority ~9. 
```

It's a difficult problem to figure out the original block markup from the parsed block, though. I'm not sure how that would work.

adekbadek: I'm not that familiar with the potential implications of priority here. Could you please share some code that a plugin might introduce and that would break things?

claudiulodro: > Could you please share some code that a plugin might introduce and that would break things?

For example, if there were two plugins injecting popups into the content in the exact same way as this PR (Newspack Popups A and Newspack Popups B), only the first plugin would correctly insert the popup. Newspack Popups B wouldn't work correctly.

Here's a contrived example that demonstrates it further:
```
// Remove all image blocks from the content.
add_filter( 'the_content', function( $content ) {
	$regex = '#<!-- wp:image.*\/wp:image -->#sU';
	return preg_replace( $regex, '', $content );
}, 2 );
```

Without this PR, this will correctly remove all image blocks from the content. With this PR active, this will do nothing.

I fear I may be overly concerned with a "correct" logic for this, though. I've been thinking about this further. I think we'll likely be OK. Let's go ahead and release this version and see if anything surfaces. It seems pretty unlikely another plugin would do something similar.

claudiulodro: Leaving a note here about a better solution I've thought of because I think this did cause issues after all, in the edge case where 2+ inline campaigns are used.

We can combine the different approaches by injecting popups into the **fully rendered** content (run the filter at priority `99` instead of `1`) and **only counting top-level paragraph elements**. The tricky part is how to accurately count top-level paragraphs (it's not feasible with regex). [DOMDocument](https://www.php.net/manual/en/class.domdocument.php) is probably the most straightforward solution I think.

PR #161, opened 2020-06-29
Refactor/hooks firing too often

jeffersonrabb:
### All Submissions:

This PR cleans up a few aspects of the campaign insertion logic to avoid rendering campaign content more than once. In addition to performance benefits, multiple passes through the insertion logic can have other unexpected side effects. One of these concerns the Homepage Posts block, which excludes posts which have already been seen on the page. If this block is in a campaign it will not show the most recent posts since earlier render passes will have already taken a few posts off the table. 

As a refactor, this PR warrants a full round of functional testing:

1. Create five campaigns, all in test mode. Two inline campaigns at 30% and 70%. One Sitewide Default overlay. One overlay filtered to a category. One more overlay that isn't filtered to a category (this one should never be shown). I generally check the option to show the Campaign title, which makes testing a little easier. Publish them all.
2. View the homepage. Observe the Sitewide Default displays.
3. View a normal post (not in the filtered category). Observe the two inline campaigns and the Sitewide Default overlay. 
4. View a post that has the filtered category applied. Observe the two inline campaigns and the category overlay.
5. View the filtered category's archive page. Observe the category overlay. 
6. View any other archive page. Observe the Sitewide Default overlay.
7. Change Frequency to "Every page" for the inline campaigns and "Once a day" for the overlays.
8. Test all of the scenarios in an incognito browser. 
9. Verify the overlays are never shown more than once, and that dismissing the inline campaigns causes them to not show again.

Verify `insert_popups_in_content` is only being called once per page. This can be done by adding an `error_log()` statement [here](https://github.com/Automattic/newspack-popups/blob/7381e6b2442dea07b3e6bdcab3156fcaa6bd0428/includes/class-newspack-popups-inserter.php#L170) and watching the log. The same experiment on `master` should reveal multiple executions.

Also verify campaigns aren't rendered multiple times.

1. On `master`, add the Homepage Posts block to an inline Campaign. 
2. View a page with the campaign on it, observe you are not seeing the top posts.
3. Switch to this branch, observe you are seeing the top stories. Bonus: switch Newspack Blocks to https://github.com/Automattic/newspack-blocks/pull/539 (if it is still open) and observe Homepage Posts doesn't show the current page either. 

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -127,7 +127,38 @@ final class Newspack_Popups_Inserter {
 	 * @param string $content The content of the post.
 	 */
 	public static function insert_popups_in_content( $content = '' ) {
-		if ( is_admin() || ! is_singular() || self::assess_has_disabled_popups( $content ) ) {
+		// Not Frontend.
+		if ( is_admin() ) {
+			return $content;
+		}
+
+		// Content is empty.
+		if ( empty( trim( $content ) ) ) {
+			return $content;
+		}
+
+		// No campaign insertion in archive pages.
+		if ( ! is_singular() ) {
+			return $content;
+		}
+
+		// If not in the loop, ignore.
+		if ( ! in_the_loop() ) {
+			return $content;
+		}
+
+		// Ignore on front page.
+		if ( is_front_page() ) {
+			return $content;
+		}
+
+		// Campaigns disabled for this page.
+		if ( self::assess_has_disabled_popups() ) {
+			return $content;
+		}
+
+		// If the current post is a Campaign, ignore.
+		if ( Newspack_Popups::NEWSPACK_PLUGINS_CPT == get_post_type() ) {
 			return $content;
 		}
 
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -147,11 +147,6 @@ final class Newspack_Popups_Inserter {
 			return $content;
 		}
 
-		// Ignore on front page.
-		if ( is_front_page() ) {
-			return $content;
-		}
-
 		// Campaigns disabled for this page.
 		if ( self::assess_has_disabled_popups() ) {
 			return $content;
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -13,6 +13,13 @@ defined( 'ABSPATH' ) || exit;
 final class Newspack_Popups_Model {
 
 	/**
+	 * An array of all retrieved campaigns, to avoid double-rendering.
+	 *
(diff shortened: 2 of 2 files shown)


jeffersonrabb:
> 1. I never saw an overlay prompt appear on the homepage, even the Sitewide Default one. This was true while both logged in and not, in an Incognito window and a normal one. Maybe this is due to [this line](https://github.com/Automattic/newspack-popups/pull/161/files#diff-9c9a6b7ce934e14992cebfb9a2b61470R150)?

Yes you're absolutely right. Not sure what I was thinking, but once removed the overlay displays on the home: https://github.com/Automattic/newspack-popups/pull/161/commits/ebba0cdd61e753f1e7755867f794e064459ee60c

jeffersonrabb:
> 2\. After setting the campaigns to Active with frequency "Once a day", and viewing the pages in an Incognito window while not logged in: the inline campaigns only appeared once, even if I didn't click "I'm not interested" to dismiss them. (They also didn't appear again if I clicked "I'm not interested".)

This is correct behavior. When inline campaigns scroll into the viewport they are marked as viewed. Typically inline campaigns' frequency is set to "Every Page."

jeffersonrabb:
> 3\. The Sitewide Default overlay kept appearing on refresh or new posts until I clicked "I'm not interested".

When an overlay campaign is in Test Mode, it will be shown on every page while logged in, and never if logged out. So, if in your testing you were still in test mode and logged in you should have seen the overlay every time. If the campaign was not in Test Mode and you viewed it incognito, you should have seen the campaign only once. However, the event that marks the campaign as viewed is clicking the background, the close button, or "I'm Not Interested." So, if you saw the overlay incognito then refreshed the page without dismissing first, you would see it again. Given these clarifications, is what you saw in testing correct or is there still something amiss?

jeffersonrabb:
> 4\. Testing the Homepage Posts block as described, on `master` I **only** saw the homepage posts in an unstyled view like the screenshot below. On this branch, I saw the homepage posts within the Inline Campaign as described, so I'd say this branch fixes even more than what you're describing.

Hmmm I couldn't reproduce this, but glad to hear the change fixed it!

dkoo:
> This is correct behavior. When inline campaigns scroll into the viewport they are marked as viewed. Typically inline campaigns' frequency is set to "Every Page."

Noted, thanks for clarifying.

> However, the event that marks the campaign as viewed is clicking the background, the close button, or "I'm Not Interested." So, if you saw the overlay incognito then refreshed the page without dismissing first, you would see it again. 

I think this is what was happening—I viewed the page, saw the overlay appear, hit refresh or back, and saw it again. When I explicitly click close, the overlay background, or "I'm not interested," the campaign doesn't appear again.

PR #164, opened 2020-07-01
refactor: render campaign content immediately before injection

jeffersonrabb:
### All Submissions:

This PR is a new approach in the ongoing saga of `the_content` filter being called by all kinds of posts besides the main one which has side effects related to `amp-analytics`, Homepage Posts block, and others. Here, querying for Campaign CPTs is separated from the code that renders them. The rendering occurs only just before inserting the campaign into content. This solves the problems that arise from blocks within a Campaign being rendered even if the campaign may not end up being used. 

This is a refactor and should not change the behavior of the plugin. To test, create multiple campaigns of different types and verify that they render properly. 

### Other information:


dkoo:
Pulling in the actions before and after campaign rendering from #163. These are necessary to be able to get campaign details from outside of this plugin.

PR #168, opened 2020-07-07
Make popupVisible non-interactive event

claudiulodro:
### All Submissions:

This PR should fix an issue with artificially low bounce rates reported in Google Analytics. Since the popupVisible event wasn't marked as a non-interaction event, anyone that saw a campaign would be marked as a user that didn't bounce. 

I'm not sure how you'd really test this, since a local environment is going to have a weird bounce rate if you've hooked it up to GA, and bounce rate analytics aren't available in real-time AFAIK. I think we should find a willing participant and make the change on their site for a couple days to see how it affects things.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -518,6 +518,9 @@ final class Newspack_Popups_Model {
 								"event_name": "<?php echo esc_html__( 'Seen', 'newspack-popups' ); ?>",
 								"event_label": "<?php echo esc_attr( $event_label ); ?>",
 								"event_category": "<?php echo esc_attr( $event_category ); ?>"
+							},
+							"extraUrlParams": {
+								"ni": 1
 							}
 						}
 					}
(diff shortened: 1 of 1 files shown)


adekbadek:
@claudiulodro I've updated the branch – recently handling of visibility events was added in the plugin. I've removed the custom-analytics-reporting code and replaced it with and event for the plugin filter, making sure the non-interaction is handled: acfbc39435fad5b65e08656208908107599a95b8

adekbadek:
@claudiulodro I've updated the branch – recently handling of visibility events was added in the plugin. I've removed the custom-analytics-reporting code and replaced it with and event for the plugin filter, making sure the non-interaction is handled: acfbc39
It turned out that the `visibility` event omits the non-interaction bit in non-AMP setting, this PR fixes that: https://github.com/Automattic/newspack-plugin/pull/581

claudiulodro:
Since bounce rate is an average of how traffic behaves, I don't think it's possible to test on a local environment. We should have solid data from live sites a couple days after we deploy. 

Thanks, Adam and Derrick!

PR #173, opened 2020-07-10
Add option to suppress newsletter campaigns if visiting from email

adekbadek:
### All Submissions:

Adds plugin settings page and an option to suppress newsletter campaigns for visitors coming from email. Next step will be to set a transient so that the site will "remember" that newsletter campaigns should be hidden after user navigates.

Related: https://github.com/Automattic/newspack-newsletters/pull/265

1. Create two campaigns – one with a newsletter form (let's call it A) and one without (B)
2. Visit the site to confirm that both are showing up
3. Append `?utm_medium=email` to URL – observe that only campaign B is visible
4. Visit the Campaigns plugin settings page (link is in the sidebar), oberve the option is on by default 
5. Unset the option
4. Re-visit the site (with the `utm_medium`), observe that both campaigns are visible
4. Remove the param from URL, observe that both campaigns are visible

### Other information:

### DIFF
--- a/includes/class-newspack-popups-api.php
+++ b/includes/class-newspack-popups-api.php
@@ -48,18 +48,7 @@ final class Newspack_Popups_API {
 			[
 				'methods'             => \WP_REST_Server::EDITABLE,
 				'callback'            => [ $this, 'set_sitewide_default_endpoint' ],
-				'permission_callback' => function() {
-					if ( ! current_user_can( 'manage_options' ) ) {
-						return new \WP_Error(
-							'newspack_rest_forbidden',
-							esc_html__( 'You cannot use this resource.', 'newspack' ),
-							[
-								'status' => 403,
-							]
-						);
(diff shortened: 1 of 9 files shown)


adekbadek:
@dkoo I think this is a part of a larger issue of Newspack plugin & other Newspack suite plugins inter-dependence. For now, I'd keep this plugin independent, so that it's usable without Newspack plugin.

dkoo:
> I think this is a part of a larger issue of Newspack plugin & other Newspack suite plugins inter-dependence. For now, I'd keep this plugin independent, so that it's usable without Newspack plugin.

I'm okay with this with the caveat that the settings page looks incomplete to me from a product standpoint. I'd almost rather we keep the standard WP admin UI (PHP-based form) as a policy for non-wizard-based admin features. More for future discussion, though. Thanks for starting the discussion!

PR #194, opened 2020-08-18
Fix bounce rates: dismissals as non-interaction

adekbadek:
### All Submissions:

Because of hos events are set up, a user dismissing a campaign and leaving the site would now count as a bounce. This PR fixes that. 

1. Create a campaign
2. Visit page, dismiss the campaign
3. Observe the event is sent as non-interaction to GA (`ni` parameter of value `1` in the request)

### Other information:

### DIFF
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -565,24 +565,26 @@ final class Newspack_Popups_Model {
 		}
 		if ( $has_dismiss_form ) {
 			$analytics_events[] = [
-				'id'             => 'popupDismissed-' . $popup_id,
-				'amp_on'         => 'amp-form-submit-success',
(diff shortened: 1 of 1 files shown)


dkoo:
@adekbadek Just to clarify—with this patch, we DO want this case to count as a bounce, right? e.g. a user visits the site for the first time, dismisses a campaign, and leaves without visiting any other pages. The dismissal is a non-interaction, so the session is a bounce.

adekbadek:
@dkoo 

> Just to clarify—with this patch, we DO want this case to count as a bounce, right? e.g. a user visits the site for the first time, dismisses a campaign, and leaves without visiting any other pages. The dismissal is a non-interaction, so the session is a bounce.

Yes, the user did not _really_ interact with the site, so it's a bounce

PR #208, opened 2020-09-04
feat: lightweight GET API for campaigns

jeffersonrabb:
### All Submissions:

Refactors the endpoint that determines whether a Campaign should be shown as a lightweight, standalone application that uses WordPress' database and cacheing but doesn't incur the full weight of a WP REST API request. On a site with campaigns this endpoint will often be hit at least once for each anonymous page load, so the lighter and faster it is, the better. 

Set-up instructions are in https://github.com/Automattic/newspack-popups/blob/7e7f5df3a9a5057f6b6d5bb203fec63099591773/api/README.md. When testing on an Atomic site, the config can omit everything except for `DB_PREFIX`. 

To test cacheing locally, install `object-cache.php` from https://wordpress.org/plugins/memcached/ in `wp-content/`. 

1. Verify that the new endpoint is being used. If the config file is not found, the plugin will use the REST API endpoint instead.
2. Verify that all Campaign visibility and suppression flows work correctly.
3. With `memcached` installed in `wp-content/` and the `NEWSPACK_POPUPS_DEBUG` constant set to true in the config, observe the API responses. They will include information about the number of DB reads/writes and cache hits. When testing with a brand new client ID (i.e. in an incognito window) the first responses will make database queries but with each page load these will be replaced by cache hits. 
4. Verify the lightweight API works on an Atomic instance.
5. Verify the response speeds are faster than the REST API endpoints.

### Next steps:

1. Rewrite the `POST`endpoint in the same style.
2. Rework the`amp-access` pattern so the endpoint is hit only once even when there are multiple campaigns. This is the approach @adekbadek took in https://github.com/Automattic/newspack-audience
3. Auto-generate the config file when the plugin is installed.

### Other information:


api/classes/class-lightweight-api.php
```diff
@@ -0,0 +1,303 @@
...
+	 */
+	public function get_transient( $transient_name ) {
+		$query                    = sprintf(
+			"SELECT option_value FROM wp_options WHERE option_name = '%s'",
+			$this->db->real_escape_string( $transient_name )
+		);
+		$this->debug['queries'][] = $query;
+		$result                   = $this->db->query( $query );
+		$row                      = $result->fetch_assoc();
+		$transient                = maybe_unserialize( $row['option_value'] );
```
claudiulodro, line 235: This should check to make sure something was actually returned from the DB before referencing `$row['option_value']`

api/classes/class-lightweight-api.php
```diff
@@ -0,0 +1,303 @@
...
+	}
+
+	/**
+	 * Insert or update a transient.
+	 *
+	 * @param string $transient_name THe transient's name.
+	 * @param string $transient THe transient's value.
+	 */
+	public function set_transient( $transient_name, $transient ) {
+		$exists                   = sprintf( "SELECT option_id FROM wp_options WHERE option_name = '%s'", $transient_name );
```
iuravic, line 246: This is a great opportunity to use a prepared statement for security reasons, I think something like this should work out of the box:
```
$exists = sprintf( "SELECT option_id FROM wp_options WHERE option_name = '%s'", $transient_name );
$exists = $this->db->prepare( "SELECT option_id FROM wp_options WHERE option_name = ?" );
$exists->bind_param( "s", $transient_name );
$exists->execute();
$result = $exists->get_result();
```

api/classes/class-lightweight-api.php
```diff
@@ -0,0 +1,303 @@
...
+	}
+
+	/**
+	 * Get data from transient.
+	 *
+	 * @param string $transient_name The transient's name.
+	 */
+	public function get_transient( $transient_name ) {
+		$query                    = sprintf(
+			"SELECT option_value FROM wp_options WHERE option_name = '%s'",
```
claudiulodro, line 229: I assume won't be necessary in the future when this has a dedicated table but just noting: we should handle different DB prefixes. It's not always going to be `wp_`. Maybe another constant in `newspack-popups-config.php`?

api/classes/class-lightweight-api.php
```diff
@@ -0,0 +1,303 @@
...
+	 */
+	public function set_transient( $transient_name, $transient ) {
+		$exists                   = sprintf( "SELECT option_id FROM wp_options WHERE option_name = '%s'", $transient_name );
+		$this->debug['queries'][] = $exists;
+		$result                   = $this->db->query( $exists );
+		$row                      = $result->fetch_assoc();
+		if ( $row ) {
+			$update                 = sprintf(
+				"UPDATE wp_options SET option_value = '%s' WHERE option_name = '%s'",
+				$this->db->real_escape_string( $transient ),
```
iuravic, line 253: This is one step towards security, but it's strongly suggested to use `prepare` instead, because `escape` doesn't actually prevent SQL injection.

Now that we mention this, I think I might be guilty of using this same thing in the Converter plugin at one point, I'll have to check that...

api/classes/class-maybe-show-campaign.php
```diff
@@ -0,0 +1,167 @@
...
+	 * @return bool Should campaign be shown.
+	 */
+	public function should_suppress_because_utm_suppression() {
+		if ( ! $this->utm_suppression ) {
+			return false;
+		}
+		// Suppressing based on UTM Source parameter in the URL.
+		// If the visitor came from a campaign with suppressed utm_source, then it should not be displayed.
+		$utm_source_transient_name = $this->get_suppression_data_transient_name( 'utm_source' );
+		$utm_source_transient      = $this->get_transient( $utm_source_transient_name );
```
claudiulodro, line 100: Maybe I missed how this is handled, but don't all the `get_transient` calls in this file need checks like this?
```
$utm_source_transient      = $this->get_transient( $utm_source_transient_name );
if ( ! $utm_source_transient || ! is_array( $utm_source_transient ) ) {
    $utm_source_transient = [];
}
```

api/classes/utils/wp-functions.php
```diff
@@ -0,0 +1,88 @@
+<?php
```
iuravic, line 1: I think we could actually simply `require_once( 'wp-includes/functions.php' );` instead of duplicating these, couldn't we? I'm also thinking not just about duplication, but if we end up using this in a WP-loaded context that we'd end up with `function already exists` errors.

api/index.php
```diff
@@ -0,0 +1,15 @@
+<?php
+/**
+ * Route Newspack Campaigns API requests based on method.
+ *
+ * @package Newspack
+ */
+
+switch ( $_SERVER['REQUEST_METHOD'] ) { //phpcs:ignore
```
iuravic, line 8: This should be good enough for a 0.1 version, very simple and efficient. Bat later on, we could improve it with some light-weight and more functional routing.

claudiulodro (changes requested):
This works well, and I think the architecture is solid. Good idea using transients for now. I've left a bunch of small nitpicks mostly focused around hardening.

api/README.md
```diff
@@ -0,0 +1,18 @@
...
+## DB Connection Environment Variables
+If `DB_NAME`, `DB_USER`, and `DB_PASSWORD` environment variables are set, the lightweight API is used and database connection is made using these credentials (DB_HOST is assumed to be `localhost`).
+
+## Special Config File
+If the environment variable approach isn't possible. include a config file named `newspack-popups-config.php` at the root of the WordPress installation. The file must contain these four constants which should be identical to the values in `wp-config.php`: `DB_NAME`, `DB_USER`, `DB_PASSWORD`, `DB_HOST`. Example of the complete file:
+
+```
+<?php
+
+define( 'DB_NAME', 'wordpress' );
```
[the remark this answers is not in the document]
claudiulodro: That's a good point. I like your solution. We definitely don't want someone to be able to do this:
```
// malicious-script.php
include newspack-popups-config.php;
do_shady_stuff_with_access_to_the_db();
```

api/README.md
```diff
@@ -0,0 +1,18 @@
+# Lightweight API
+
+The lightweight API is a stripped-down alternative to the WordPress REST API. Campaigns will often be present on every page of a site, which can lead to performance issues as a result of a large number of uncacheable API requests. The lightweight API will be used in two scenarios:
+
+## DB Connection Environment Variables
+If `DB_NAME`, `DB_USER`, and `DB_PASSWORD` environment variables are set, the lightweight API is used and database connection is made using these credentials (DB_HOST is assumed to be `localhost`).
+
+## Special Config File
```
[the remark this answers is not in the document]
claudiulodro: I retract this statement after reading Ivan's convincing argument. :)

api/classes/class-lightweight-api.php
claudiulodro: If the plugin does the approach Ivan recommends where we parse the wp-config file, this could get the prefix from there.

api/README.md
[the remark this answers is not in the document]
jeffersonrabb: Much nicer solution! Done in https://github.com/Automattic/newspack-popups/pull/208/commits/61c2bf9a120399e40bbbbb79a00c9f49d97b0bd3

api/classes/class-lightweight-api.php
```diff
@@ -0,0 +1,303 @@
...
+	 * @var db
+	 */
+	public $db;
+
+	/**
+	 * Unique client ID.
+	 *
+	 * @var rid
+	 */
+	public $rid;
```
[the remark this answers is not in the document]
jeffersonrabb: Sure: https://github.com/Automattic/newspack-popups/pull/208/commits/2f95572f28acf0f3c82b285caaaa77c06b8a0766

api/classes/class-lightweight-api.php
```diff
@@ -0,0 +1,303 @@
...
+	 *
+	 * @var debug
+	 */
+	public $debug;
+
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		$this->debug = [
```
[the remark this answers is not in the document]
jeffersonrabb: Will come back to this later...

api/classes/class-lightweight-api.php
```diff
@@ -0,0 +1,303 @@
...
+		}
+	}
+
+	/**
+	 * Verify referer is valid.
+	 */
+	public function verify_referer() {
+		$http_referer = ! empty( $_SERVER['HTTP_REFERER'] ) ? parse_url( $_SERVER['HTTP_REFERER'] , PHP_URL_HOST ) : null; // phpcs:ignore
+		$valid_referers = [
+			$http_referer,
```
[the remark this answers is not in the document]
jeffersonrabb: > or take them from config?

What do you mean?

api/index.php
jeffersonrabb: Yes, although this particular endpoint is extremely simple - there are only two variants and should never be more.

includes/class-newspack-popups-inserter.php
```diff
@@ -327,9 +331,20 @@ public static function insert_popups_amp_access( $popups ) {
...
+				esc_url( $endpoint ) .
+				'?popup_id=' . esc_attr( $popup['id'] ) .
+				'&rid=READER_ID&url=CANONICAL_URL&RANDOM&' .
+				'frequency=' . esc_attr( $popup['options']['frequency'] ) .
+				'&placement=' . esc_attr( $popup['options']['placement'] ) .
+				'&utm_suppression=' . esc_attr( $popup['options']['utm_suppression'] ) .
+				'&suppress_newsletter_campaigns=' . esc_attr( $settings['suppress_newsletter_campaigns'] ) .
+				'&suppress_all_newsletter_campaigns_if_one_dismissed=' . esc_attr( $settings['suppress_all_newsletter_campaigns_if_one_dismissed'] ) .
+				'&has_newsletter_prompt=' . esc_attr( \Newspack_Popups_Model::has_newsletter_prompt( $popup ) );
+
```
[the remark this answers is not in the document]
jeffersonrabb: Good idea https://github.com/Automattic/newspack-popups/pull/208/commits/a980acbd7ad6927d303cf38aefd26c797f2c5d9d

api/classes/class-lightweight-api.php
```diff
@@ -0,0 +1,303 @@
...
+		}
+		$config = $_SERVER['DOCUMENT_ROOT'] . '/newspack-popups-config.php'; //phpcs:ignore
+		if ( file_exists( $config ) ) {
+			require $config;
+			return [
+				'db_host'     => DB_HOST,
+				'db_name'     => DB_NAME,
+				'db_user'     => DB_USER,
+				'db_password' => DB_PASSWORD,
+			];
```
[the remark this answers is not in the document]
jeffersonrabb: No longer relevant after the changes in https://github.com/Automattic/newspack-popups/pull/208/commits/61c2bf9a120399e40bbbbb79a00c9f49d97b0bd3

dkoo (changes requested):
Tried to be fairly comprehensive in my functional testing. @adekbadek / @jeffersonrabb, am I missing any important suppression scenarios in the tests below?

TL;DR: Most tests worked as expected, except 

1. the "Suppress Newsletter campaigns if visitor is coming from email" option did not seem to work
2. the "display once" and "display once a day" settings don't seem to work across AMP vs. non-AMP sessions (but this only results in at most one extra campaign display as it is remembered within each session)
3. the `utm_source` suppression did not seem to be remembered after navigating away from the initial page.

Full test results (checked items are working as expected):

## Test 1

### Test 1 conditions
- Campaign A: Overlay campaign, sitewide default, contains newsletter prompt
- Campaign B: Inline campaign, displays every page, contains newsletter prompt
- Campaign C: Inline campaign, displays every page, contains newsletter prompt
- Campaign D: Inline campaign, displays every page, does NOT contain newsletter prompt
- Suppress Newsletter campaigns if visitor is coming from email: ENABLED
- Suppress all Newsletter campaigns if at least one Newsletter campaign was permanently dismissed: ENABLED

### Test 1 results
- [x] When viewing the site for the first time in a new incognito window: all campaigns are displayed, as expected.
- [x] After refreshing the page, without interacting with any campaigns: all campaigns still show, as expected.
- [x] After dismissing Campaign D (no newsletter prompt), it does not reappear after refreshing the page. Campaigns A-C do appear as expected.
- [x] NOT WORKING: After dismissing Campaign A, B, or C (contains a newsletter prompt): I encounter the `utm_suppression` notice highlighted below on subsequent page refreshes, and the other newsletter prompt campaigns are not suppressed.
- [x] Loading a page in a new incognito window with the `utm_medium=email` param: Campaigns A-C are suppressed, while Campaign D is not, as expected.
- [x] After navigating to another page, without the `utm_medium` param: Campaigns A-C are still not displayed, as expected.


## Test 2

### Test 2 conditions
- Campaigns A-D: Same as in Test 1
- Suppress Newsletter campaigns if visitor is coming from email: DISABLED
- Suppress all Newsletter campaigns if at least one Newsletter campaign was permanently dismissed: ENABLED

### Test 2 results
- [x] Loading a page in a new incognito window with the `utm_medium=email` param: All campaigns are displayed, as expected.

## Test 3

### Test 3 conditions
- Campaigns A-D: Same as in Tests 1-2
- Suppress Newsletter campaigns if visitor is coming from email: ENABLED
- Suppress all Newsletter campaigns if at least one Newsletter campaign was permanently dismissed: DISABLED

### Test 3 results
- [x] Loading a page in a new incognito window with the `utm_medium=email` param: Campaigns A-C are not displayed, while Campaign D, as expected.
- [x] After navigating to another page, without the `utm_medium` param: Campaigns A-C are still not displayed, while Campaign D is, as expected.
- [x] Loading a page in a new incognito window without the `utm_medium=email` param: All campaigns are displayed, as expected.
- [x] After dismissing Campaign A, B, or C (contains a newsletter prompt): Only the dismissed campaign is suppressed on subsequent page loads, as expected.

## Test 4

### Test 4 conditions
- Campaigns A-C: Same as in Tests 1-3
- Campaign D: Inline campaign, displays once a day, no newsletter prompt

### Test 4 results
- [x] In a new incognito window, all campaigns are displayed, as expected.
- [x] After navigating to a new page: Campaigns A-C are displayed, Campaign D is not, as expected.
- [ ] NOT WORKING: If the first page visited is an AMP page and the second page is not (or vice versa), Campaign D is still displayed on the second page (but is not on subsequent pages—seems to be that the suppression is just not remembered across AMP vs. non-AMP sessions). 

## Test 5

### Test 5 conditions
- Campaign A: Overlay campaign, sitewide default, contains newsletter prompt, TEST MODE
- Campaigns B-D: Same as in Tests 1-3

### Test 5 results
- [x] In a new incognito window: Campaign A is not displayed, Campaigns B-D are displayed, as expected.
- [x] After logging into WP Admin: Campaign A is displayed, Campaigns B-D are not, as expected.

## Test 6

### Test 6 conditions:
- Campaigns A-C: Same as in Tests 1-3
- Campaign D: UTM Suppression field set to "suppress this"

### Test 6 results:
- [x] In a new incognito window, all campaigns are displayed, as expected.
- [x] When visting the page with the `utm_source=suppress+this` or `utm_source=suppres%20this` param, Campaigns A-C are displayed, while Campaign D is not, as expected.
- [ ] NOT WORKING (?): After navigating to a new page, without the `utm_source` param: Campaign D is displayed again. Expected: the suppression should be remembered across pages once it happens the first time. Marked (?) as I'm not sure whether my expectation is correct? cc @adekbadek

api/classes/class-lightweight-api.php
adekbadek: I know this is supposed to be bypassing WP, but the DB prefix is `$wpdb->prefix` – it could be read once and stored for later perhaps?

api/classes/class-maybe-show-campaign.php
```diff
@@ -0,0 +1,167 @@
...
+				$should_show = false;
+			}
+			if ( $should_show && $this->should_suppress_because_newsletter_campaign_dismissed() ) {
+				$show_show = false;
+			}
+		}
+
+		$this->response['displayPopup'] = $should_show;
+
+		// TODO: Preview handling.
```
adekbadek, line 38: Previews do not have to call the API at all – that's the approach on the exploratory [add/external-reader-service](https://github.com/Automattic/newspack-popups/tree/add/external-reader-service) branch. Just a heads up, I'll be happy to contribute to this branch :)

api/classes/class-lightweight-api.php
```diff
@@ -0,0 +1,303 @@
...
+
+	/**
+	 * Establish DB connection.
+	 */
+	public function connect() {
+		$credentials = $this->get_credentials();
+		if ( ! $credentials ) {
+			$this->error( 'no_credentials' );
+		}
+		$db = mysqli_init(); //phpcs:ignore
```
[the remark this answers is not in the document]
jeffersonrabb: Good suggestion, I've switched to PDO in https://github.com/Automattic/newspack-popups/pull/208/commits/56329eee53138e6d92df4199407984050a2e11f6

api/classes/class-lightweight-api.php
jeffersonrabb: Switched to prepared statements (along with PDO) in https://github.com/Automattic/newspack-popups/pull/208/commits/56329eee53138e6d92df4199407984050a2e11f6

api/classes/class-lightweight-api.php
jeffersonrabb: https://github.com/Automattic/newspack-popups/pull/208/commits/56329eee53138e6d92df4199407984050a2e11f6

api/classes/class-lightweight-api.php
```diff
@@ -0,0 +1,303 @@
...
+	public function get_credentials() {
+		if ( getenv( 'DB_NAME' ) && getenv( 'DB_USER' ) && getenv( 'DB_PASSWORD' ) ) {
+			return [
+				'db_host'     => 'localhost',
+				'db_name'     => getenv( 'DB_NAME' ),
+				'db_user'     => getenv( 'DB_USER' ),
+				'db_password' => getenv( 'DB_PASSWORD' ),
+			];
+		}
+		$config = $_SERVER['DOCUMENT_ROOT'] . '/newspack-popups-config.php'; //phpcs:ignore
```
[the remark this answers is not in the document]
jeffersonrabb: This was done in https://github.com/Automattic/newspack-popups/pull/208/commits/61c2bf9a120399e40bbbbb79a00c9f49d97b0bd3

api/classes/utils/wp-functions.php
jeffersonrabb: >  if we end up using this in a WP-loaded context

I don't think this work would ever be used in a WP-loaded context. The goal of this branch is to be as lean as possible — to reduce queries and code down to the bare necessities. If we used it in a WordPress context we'd negate this.

jeffersonrabb: > I think we could actually simply `require_once( 'wp-includes/functions.php' );`

I'm wary of including any WordPress files because they will create messy dependency issues. For example, if we include `functions.php` we immediately get warnings about `ABSPATH` and `WPINC`. We won't need many of these functions so I'd prefer to keep it ultra-lean, even if that means a verbatim copy of a function or two.

api/classes/class-lightweight-api.php
```diff
@@ -0,0 +1,303 @@
...
+	 * Verify referer is valid.
+	 */
+	public function verify_referer() {
+		$http_referer = ! empty( $_SERVER['HTTP_REFERER'] ) ? parse_url( $_SERVER['HTTP_REFERER'] , PHP_URL_HOST ) : null; // phpcs:ignore
+		$valid_referers = [
+			$http_referer,
+			// TODO: Add AMP Cache.
+		];
+		$http_host = ! empty( $_SERVER['HTTP_HOST'] ) ? $_SERVER['HTTP_HOST'] : null; // phpcs:ignore
+		return in_array( strtolower( $http_host ), $valid_referers, true );
```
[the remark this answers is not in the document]
jeffersonrabb: > null will be in $valid_referers

Why would `null` be in that array? I did a quick test forcing `$http_host` to be `null` and the API errors out (correctly).

api/classes/class-lightweight-api.php
```diff
@@ -0,0 +1,303 @@
...
+			$credentials['db_password'],
+			$credentials['db_name']
+		);
+		return $db;
+	}
+
+	/**
+	 * Get database credentials from environment variables (Atomic) or special config file.
+	 */
+	public function get_credentials() {
```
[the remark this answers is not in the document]
jeffersonrabb: Yeah, definitely. Added charset in https://github.com/Automattic/newspack-popups/pull/208/commits/f18332f87e59c973a0441c5b1440bab40d8a6009

api/classes/class-lightweight-api.php
jeffersonrabb: Done in https://github.com/Automattic/newspack-popups/pull/208/commits/aac241f2d83fc5688b32f5d286fa5b4931fb91a7

api/classes/class-lightweight-api.php
jeffersonrabb: I took a shot at extracting the prefix using a regex in https://github.com/Automattic/newspack-popups/pull/208/commits/cc16d6d0eaa5e015711ba4ea2cb2624010e25525. I'm becoming more and more concerned about extracting the DB credentials from `wp-config.php` by string manipulation though. To give one example of the potential brittleness, if there are commented-out constants the script gets confused, e.g.:

```
/*
define( 'DB_NAME', 'notreal' );
*/
define( 'DB_NAME', 'wordpresssimple' );
```

jeffersonrabb: > it could be read once and stored for later perhaps?

@adekbadek how would this work?

api/classes/class-maybe-show-campaign.php
jeffersonrabb: This should be taken care of as of https://github.com/Automattic/newspack-popups/pull/208/commits/aac241f2d83fc5688b32f5d286fa5b4931fb91a7#diff-bf20f4bd9a3d050c28b9648d1c89c0d3R242. I don't think we need the `is_array` check because the DB result could only be an array or `null`.

api/classes/class-maybe-show-campaign.php
```diff
@@ -0,0 +1,167 @@
...
+		// If the visitor came from a campaign with suppressed utm_source, then it should not be displayed.
+		$utm_source_transient_name = $this->get_suppression_data_transient_name( 'utm_source' );
+		$utm_source_transient      = $this->get_transient( $utm_source_transient_name );
+		if ( $this->referer_url && stripos( urldecode( $this->referer_url ), 'utm_source=' . $this->utm_suppression ) ) {
+			$utm_source_transient[ $this->utm_suppression ] = true;
+			$this->set_transient( $utm_source_transient_name, $utm_source_transient );
+			return true;
+		}
+
+		if ( ! empty( $utm_source_transient[ $utm_suppression ] ) ) {
```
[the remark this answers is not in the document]
jeffersonrabb: Oops, yes you're correct! https://github.com/Automattic/newspack-popups/pull/208/commits/71e5c7fc78c2cacf2b59e08e0424fcd2d38921af

jeffersonrabb:
> * NOT WORKING: After dismissing Campaign A, B, or C (contains a newsletter prompt): I encounter the `utm_suppression` notice highlighted below on subsequent page refreshes, and the other newsletter prompt campaigns are not suppressed.

@dkoo thanks for such comprehensive functional testing! This one should be fixed in https://github.com/Automattic/newspack-popups/pull/208/commits/e55e2d04b170f130330ed9a6aa5606748052c98b.

jeffersonrabb:
> * NOT WORKING: If the first page visited is an AMP page and the second page is not (or vice versa), Campaign D is still displayed on the second page (but is not on subsequent pages—seems to be that the suppression is just not remembered across AMP vs. non-AMP sessions).

I suspect this issue existed before this branch. @dkoo could you try this out on `master` and see if it can be reproduced?

jeffersonrabb:
> * NOT WORKING (?): After navigating to a new page, without the `utm_source` param: Campaign D is displayed again. Expected: the suppression should be remembered across pages once it happens the first time. Marked (?) as I'm not sure whether my expectation is correct?

@dkoo you're correct, the suppression should be remembered. This was a bug, fixed in https://github.com/Automattic/newspack-popups/pull/208/commits/e55e2d04b170f130330ed9a6aa5606748052c98b

api/classes/class-lightweight-api.php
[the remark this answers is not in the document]
iuravic: Sorry if I was unclear. What I meant was since the value of `$_SERVER['HTTP_REFERER']` [can be set by the client](https://thisinterestsme.com/php-spoof-http-referer-field/) as they wish, instead of just trusting the value of `$_SERVER['HTTP_REFERER']`, should we have a hard-coded list of valid referers (somewhere in config, or option, or else)?

api/classes/class-lightweight-api.php
iuravic: Perfect !

api/classes/class-lightweight-api.php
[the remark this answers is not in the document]
jeffersonrabb: I wouldn't want to put it in the Database because this would add a query. If we put it in a config, this would suggest we need to go back to the original approach where the plugin has it's own custom config file.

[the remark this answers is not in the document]
iuravic: I understand, and also agree. I think that my suggestion to store valid HTTP_REFERERs maybe wasn't good at all. 

Thinking once again about it -- this is simple a form of authentication, right? We could probably elaborate and use something else.

Pinging @adekbadek -- Adam, since you have experience with implementing WPCOM authentication, do you think it would be feasible to implement the same kind of oAuth verification here? Briefly put -- this is a self-standing light-weight API with a couple of GET and POST endpoints, meant to run on Newspack Atomic sites. We're also looking for something fast, and easy on the resources.

api/classes/class-lightweight-api.php
adekbadek: The plugin could generate a file – unless there's already one – with some initial WP-related data. That could be the only point of contact with WP.

api/classes/class-lightweight-api.php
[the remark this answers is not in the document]
claudiulodro:
```		
$http_referer = ! empty( $_SERVER['HTTP_REFERER'] ) ? parse_url( $_SERVER['HTTP_REFERER'] , PHP_URL_HOST ) : null; // phpcs:ignore
$valid_referers = [
	$http_referer,
];
```
`$http_referer` is always in the `$valid_referers` array, and `$http_referer` can be `null`, so you could end up with `$valid_referers = [ null ]`

api/classes/class-lightweight-api.php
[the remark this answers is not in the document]
adekbadek: In other works, the proposition is to use a self-issued authentication token to verify that the API caller is in fact the site, right? That still would mean an additional query, as the token would have to be stored somewhere.

dkoo:
> I suspect this issue existed before this branch. @dkoo could you try this out on master and see if it can be reproduced?

@jeffersonrabb Tested on master and the suppression does seem to be remembered across AMP vs. non-AMP. Steps to test:

* Site is set to AMP-first.
* Disable AMP for Post 1. Enable AMP for Post 2.
* Visit Post 1 in a new incognito window. Observe that all campaigns are shown
* Visit Post 2 in the same window. Campaigns A-C (set to show every page) are shown, but Campaign D (set to once per day) is not.

api/classes/class-maybe-show-campaign.php
jeffersonrabb: Here's a quick fix - previews will use the REST API instead: https://github.com/Automattic/newspack-popups/pull/208/commits/c42783ce339efe7f3aa85947543ed50f26915bfa. The approach in `add/external-reader-service` looks good but I think we should address separately after this ships.

api/classes/class-lightweight-api.php
jeffersonrabb: Take a look at this [draft PR](https://github.com/Automattic/newspack-popups/pull/210), which changes back to using  a custom config file. We could look at auto-generating this file when the plugin is activated.

api/classes/class-lightweight-api.php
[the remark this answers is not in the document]
jeffersonrabb: This should do it: https://github.com/Automattic/newspack-popups/pull/208/commits/33042c6bcd9b8b82458155ffdc361894b4484801

api/classes/class-lightweight-api.php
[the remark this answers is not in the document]
jeffersonrabb: https://github.com/Automattic/newspack-popups/pull/210 brings back the idea of a custom config file. If we go this way we could hardcode the hostname in the config, as @iuravic suggests.

jeffersonrabb:
> @jeffersonrabb Tested on master and the suppression does seem to be remembered across AMP vs. non-AMP. Steps to test:

@dkoo I can't seem to repro this one.

jeffersonrabb:
> * NOT WORKING: If the first page visited is an AMP page and the second page is not (or vice versa), Campaign D is still displayed on the second page (but is not on subsequent pages—seems to be that the suppression is just not remembered across AMP vs. non-AMP sessions).

@dkoo I still can't seem to reproduce this one. I'm consistently seeing the campaign hidden on the second page, regardless of whether I go from AMP->non AMP or vice-versa. One possible cause would be if the page is exited before the API call that marks the popup as viewed. Could you try this testing approach and see if you are still able to reproduce it?

- Open new incognito window.
- Open Network tab, click ALL filter, type "reader" in the filter field.
- Slowly scroll down the page until the once-a-day campaign is fully in view. 
- You should see a `POST` request to `/wp-json/newspack-popups/v1/reader`. 
- Stay on the page until it completes.
- View the other page, see if the campaign is shown.

jeffersonrabb:
Sorry, closed this by mistake! Reopening.

dkoo:
> One possible cause would be if the page is exited before the API call that marks the popup as viewed.

This may be the cause, actually. I kept navigating away from the page before the 3-second delay for the overlay to appear, because I wanted to make sure the overlay would appear again on the next page. Waiting for the `reader` requests to complete before navigating away did ensure that the "once a day" campaign wasn't shown on the next page, AMP or not. I'd say this "issue" is resolved, then.

api/classes/class-lightweight-api.php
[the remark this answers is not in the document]
claudiulodro: > In other words, the proposition is to use a self-issued authentication token to verify that the API caller is in fact the site, right? That still would mean an additional query, as the token would have to be stored somewhere. But other than that, it's feasible.

Won't someone just be able to copy the token by examining the API request?

Assuming all the other security bits are done well, I think the main "danger" with the API would be an attacker making hundreds of thousands of requests to fill up the database. I assume the system-level checks that prevent DoS attacks would catch that sort of attack, though.

api/classes/class-lightweight-api.php
```diff
@@ -0,0 +1,339 @@
...
+
+		if ( isset( $empty_transients[ $name ] ) ) {
+			$this->debug['cache_count'] += 1;
+			return null;
+		}
+
+		$value = wp_cache_get( $name, 'newspack-popups' );
+		if ( false === $value ) {
+			$this->debug['read_query_count'] += 1;
+			$row  = $wpdb->get_row( $wpdb->prepare( "SELECT option_value FROM $wpdb->options WHERE option_name = %s LIMIT 1", $name ) ); // phpcs:ignore
```
claudiulodro, line 303: Tiny nitpick: [I think `wpdb::get_var` may be simpler here](https://developer.wordpress.org/reference/classes/wpdb/get_var/).

api/classes/class-lightweight-api.php
```diff
@@ -0,0 +1,339 @@
...
+				wp_cache_set( 'empty_transients', $empty_transients, 'newspack-popups' );
+			}
+		} else {
+			$this->debug['cache_count'] += 1;
+		}
+		return maybe_unserialize( $value );
+	}
+
+	/**
+	 * Upsert transient.
```
claudiulodro, line 318: I don't know what "Upsert" is even supposed to mean. :)

api/classes/utils/wp-functions.php
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * Some useful functions taken directly from the WordPress codebase.
```
claudiulodro, line 3: Since we're already including `wp-db.php`, it's probably worth it to also include `wp-includes/functions.php` instead of copy+pasting the functions.

includes/class-newspack-popups-api.php
```diff
@@ -409,6 +431,7 @@ public function set_utm_source_transient( $utm_source ) {
 	public function set_utm_medium_transient() {
 		$transient_name = self::get_suppression_data_transient_name( 'utm_medium' );
 		if ( $transient_name ) {
+			$this->update_cache( $transient_name, true );
 			set_transient( $transient_name, true, 0 );
```
claudiulodro, line 435: Tiny nitpick: Is there ever a situation where `$this->update_cache` is not followed by `set_transient`? If not, you could move `set_transient` into `update_cache` (or something like that) and save a bunch of repeating yourself.

claudiulodro (approved):
This works great in my testing. The new endpoint is way faster than the old REST API. Good work.

I left a handful of tiny nitpicks and questions, but none are really blockers.

api/classes/class-lightweight-api.php
claudiulodro: Config file solves this problem nicely.

api/classes/class-lightweight-api.php
```diff
@@ -0,0 +1,339 @@
...
+	}
+
+	/**
+	 * Get database credentials from environment variables (Atomic) or special config file.
+	 */
+	public function get_credentials() {
+		if ( $this->credentials ) {
+			return $this->credentials;
+		}
+		$config_path = $_SERVER['DOCUMENT_ROOT'] . '/wp-config.php'; //phpcs:ignore
```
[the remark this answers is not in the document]
jeffersonrabb: Actually the whole `get_credentials` function is now unused. Removed in https://github.com/Automattic/newspack-popups/pull/208/commits/eebe11e6842ffe3074996b6d8a761fcbcb6ed66b. Good catch!

api/classes/class-lightweight-api.php
```diff
@@ -0,0 +1,339 @@
...
+		exit;
+	}
+
+	/**
+	 * Return a 500 code error.
+	 *
+	 * @param string $code The error code.
+	 */
+	public function error( $code ) {
+		http_response_code( 500 );
```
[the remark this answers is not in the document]
jeffersonrabb: Done in https://github.com/Automattic/newspack-popups/pull/208/commits/216551a149b44114bde0e4b19acd3afe24163a96

api/classes/utils/wp-functions.php
jeffersonrabb: I was initially wary of including `functions.php` since that file requires `options.php`, but it seems to be working well. I had to rework the `ABSPATH` handling a bit to get this change working correctly on Atomic sites. https://github.com/Automattic/newspack-popups/pull/208/commits/7d987634230e192af15e298564b97ec6e6b74d3c

api/classes/class-lightweight-api.php
jeffersonrabb: I stand by it! https://docs.mongodb.com/manual/reference/method/Bulk.find.upsert/

claudiulodro: TIL!

api/classes/class-lightweight-api.php
jeffersonrabb: Good call https://github.com/Automattic/newspack-popups/pull/208/commits/f143157fc087b59f96c8aac4802416ad1696a219

includes/class-newspack-popups-api.php
jeffersonrabb: Good idea. https://github.com/Automattic/newspack-popups/pull/208/commits/9e902337a0d46df83e7fc6341db56baa0e347566

api/index.php
```diff
@@ -0,0 +1,66 @@
...
+wp_cache_init();
+
+// phpcs:enable
+
+switch ( $_SERVER['REQUEST_METHOD'] ) { //phpcs:ignore
+	case 'GET':
+		include 'classes/class-maybe-show-campaign.php';
+		break;
+	case 'POST':
+		include 'classes/class-dismiss-campaign.php';
```
adekbadek, line 64: Note: the `POST` endpoint is not only for dismissals, but also for reporting a campaign has been seen, for instance.

api/classes/class-lightweight-api.php
```diff
@@ -0,0 +1,283 @@
...
+		if ( defined( 'NEWSPACK_POPUPS_DEBUG' ) && NEWSPACK_POPUPS_DEBUG ) {
+			$this->response['debug'] = $this->debug;
+		}
+		http_response_code( 200 );
+		print json_encode( $this->response ); // phpcs:ignore
+		exit;
+	}
+
+	/**
+	 * Return a 500 code error.
```
adekbadek, line 215: It's `400` now ;)

api/index.php
jeffersonrabb: True, but even more significantly this route shouldn't be present at all until the `POST` endpoint is implemented 😬 . Sorry I missed this! Removed in https://github.com/Automattic/newspack-popups/pull/208/commits/d66c9d58902a1511e74ff8314612d3da0bd37af4

api/classes/class-lightweight-api.php
jeffersonrabb: 😞 https://github.com/Automattic/newspack-popups/pull/208/commits/588c6c2367c8f43aa91a3cd1750c529623728b78

api/classes/class-lightweight-api.php
```diff
@@ -231,26 +233,19 @@ public function get_transient( $name ) {
...
 
-		$empty_transients = wp_cache_get( 'empty_transients', 'newspack-popups' );
-
-		if ( ! is_array( $empty_transients ) ) {
-			$empty_transients = array();
-		}
-
-		if ( isset( $empty_transients[ $name ] ) ) {
+		$value = wp_cache_get( $name, 'newspack-popups' );
+		if ( -1 === $value ) {
```
iuravic, line 237: This is wonderful ! , exactly what we needed here IMHO.

api/classes/class-lightweight-api.php
```diff
@@ -231,26 +233,19 @@ public function get_transient( $name ) {
...
-		}
-
-		$value = wp_cache_get( $name, 'newspack-popups' );
-		if ( false === $value ) {
+		} elseif ( false === $value ) {
 			$this->debug['read_query_count'] += 1;
 			$value = $wpdb->get_var( $wpdb->prepare( "SELECT option_value FROM $wpdb->options WHERE option_name = %s LIMIT 1", $name ) ); // phpcs:ignore
 			if ( $value ) {
-				wp_cache_add( $name, $value );
+				wp_cache_set( $name, $value );
```
iuravic, line 245: Do we want to use the Group Name param, too?
```
wp_cache_set( $name, $value, 'newspack-popups' );
```

iuravic (approved):

api/classes/class-lightweight-api.php
jeffersonrabb: Eek, you're absolutely right. Fixed inhttps://github.com/Automattic/newspack-popups/pull/208/commits/7b229574adcefb5a578980b254b0cf294a283e61

api/classes/class-lightweight-api.php
jeffersonrabb: Great suggestion!

jeffersonrabb:
@dkoo could I get your approval on this so I can merge?

PR #211, opened 2020-09-09
Handle false get_transient return value

### LINKED ISSUE #179: Missing transient causes PHP warning
https://github.com/Automattic/newspack-popups/pull/175#issuecomment-658327896

Occasionally, I see this PHP warning for a campaign: 

```
PHP Notice:  Trying to access array offset on value of type bool in /Users/dkoo/Local Sites/newspack/app/public/wp-content/plugins/newspack-popups/includes/class-newspack-popups-api.php on line 162
```

This seems to happen when `get_transient` in the following line returns `false`: https://github.com/Automattic/newspack-popups/blob/master/includes/class-newspack-popups-api.php#L160

The warning does not seem to affect campaign behavior. The early Returning the `$response` early suppresses the warning but also causes the campaigns to not display.

Haven't figured out reliable replication steps yet, but if I do I'll add them here.

adekbadek:
### All Submissions:

It appears that something about get_transient has changed when PHP 7.4 is running.

Closes #179 

1. Switch to PHP 7.4
2. On `master`, observe a notice logged when Campaigns POST endpoint is hit
3. Switch to this branch, observe no notice logged

### Other information:

### DIFF
--- a/includes/class-newspack-popups-api.php
+++ b/includes/class-newspack-popups-api.php
@@ -159,6 +159,11 @@ final class Newspack_Popups_API {
 			return rest_ensure_response( $response );
 		}
 		$data = get_transient( $transient_name );
+		if ( false === $data ) {
+			$data = [
+				'count' => 0,
+			];
+		}
 
 		$response['currentViews'] = (int) $data['count'];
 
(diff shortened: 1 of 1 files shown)


jeffersonrabb (approved):
👍

PR #212, opened 2020-09-14
feat: create config if not exists

adekbadek:
A proposition: create config file on plugin initialisation, if it does not exist.

### DIFF
--- a/includes/class-newspack-popups.php
+++ b/includes/class-newspack-popups.php
@@ -402,13 +402,15 @@ final class Newspack_Popups {
 		file_put_contents(
 			$config_path,
 			'<?php' .
-			( defined( 'DB_USER' ) ? "\ndefine( 'DB_USER', '" . DB_USER . "' );" : '' ) .
-			( defined( 'DB_PASSWORD' ) ? "\ndefine( 'DB_PASSWORD', '" . DB_PASSWORD . "' );" : '' ) .
-			( defined( 'DB_NAME' ) ? "\ndefine( 'DB_NAME', '" . DB_NAME . "' );" : '' ) .
-			( defined( 'DB_HOST' ) ? "\ndefine( 'DB_HOST', '" . DB_HOST . "' );" : '' ) .
-			( defined( 'DB_CHARSET' ) ? "\ndefine( 'DB_CHARSET', '" . DB_CHARSET . "' );" : '' ) .
-			( defined( 'WP_CACHE_KEY_SALT' ) ? "\ndefine( 'WP_CACHE_KEY_SALT', '" . WP_CACHE_KEY_SALT . "' );" : '' ) .
-			( defined( 'DB_PREFIX' ) ? '' : "\ndefine( 'DB_PREFIX', '" . $wpdb->prefix . "' );" ) .
+			// Insert these only if they are defined, but not in the as environment variables.
+			// This way only variables which are already declared in wp-config.php should be inserted in this config file.
+			( ! getenv( 'DB_USER' ) && defined( 'DB_USER' ) ? "\ndefine( 'DB_USER', '" . DB_USER . "' );" : '' ) .
+			( ! getenv( 'DB_PASSWORD' ) && defined( 'DB_PASSWORD' ) ? "\ndefine( 'DB_PASSWORD', '" . DB_PASSWORD . "' );" : '' ) .
+			( ! getenv( 'DB_NAME' ) && defined( 'DB_NAME' ) ? "\ndefine( 'DB_NAME', '" . DB_NAME . "' );" : '' ) .
+			( ! getenv( 'DB_HOST' ) && defined( 'DB_HOST' ) ? "\ndefine( 'DB_HOST', '" . DB_HOST . "' );" : '' ) .
+			( ! getenv( 'DB_CHARSET' ) && defined( 'DB_CHARSET' ) ? "\ndefine( 'DB_CHARSET', '" . DB_CHARSET . "' );" : '' ) .
+			( ! getenv( 'WP_CACHE_KEY_SALT' ) && defined( 'WP_CACHE_KEY_SALT' ) ? "\ndefine( 'WP_CACHE_KEY_SALT', '" . WP_CACHE_KEY_SALT . "' );" : '' ) .
+			"\ndefine( 'DB_PREFIX', '" . $wpdb->prefix . "' );" .
 			"\n"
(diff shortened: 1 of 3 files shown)


adekbadek:
Note: if this approach is to be followed, `api/README.md` should be updated.

claudiulodro (changes requested):
Great idea! I think this is [missing a few of the params from the latest version of the API](https://github.com/Automattic/newspack-popups/blob/7e7f5df3a9a5057f6b6d5bb203fec63099591773/api/README.md) then it should be good to go: the charset and WP_CACHE_KEY_SALT.

iuravic:
Great solution, taking all into account 👍

includes/class-newspack-popups.php
```diff
@@ -388,5 +389,28 @@ public static function popup_default_fields( $post_id, $post, $update ) {
...
 	}
+
+	public static function create_lightweight_api_config() {
+		$config_path = WP_CONTENT_DIR . '/../newspack-popups-config.php';
+		$has_config  = file_exists( $config_path );
+		if ( false === $has_config ) {
+			global $wpdb;
+			$config_file = fopen( $config_path, 'w' );
+			if ( $config_file ) {
+				fwrite(
```
iuravic, line 400: This is perfect as is and no need to change, just sharing that I usually prefer to use `file_put_contents()` instead, because it's a bit simpler -- doesn't need `fopen` and `fclose`, and takes file name string as argument.

includes/class-newspack-popups.php
```diff
@@ -388,5 +389,28 @@ public static function popup_default_fields( $post_id, $post, $update ) {
 		update_post_meta( $post_id, 'trigger_scroll_progress', 30 );
 		update_post_meta( $post_id, 'utm_suppression', '' );
 	}
+
+	public static function create_lightweight_api_config() {
+		$config_path = WP_CONTENT_DIR . '/../newspack-popups-config.php';
+		$has_config  = file_exists( $config_path );
+		if ( false === $has_config ) {
```
iuravic, line 396: Here sharing a general and simple coding style personal preference -- some fun stuff -- but not asking you to change anything here if you don't want to.

I generally always prefer using the early return pattern, instead of nesting code blocks into `if`s, whenever possible (in all coding languages). I'm sure this is just a personal preference, and therefore not very important, but for example. instead of:
```
function() {
  if ( something ) {
     do stuff
  }
}
```
I always try and use this:
```
function() {
  if ( ! something) {
    return
  }

  do stuff
}
```

This particular style:
- removes a whole nesting block and, makes code a bit more readable -- especially if there's another block/conditional/loop in the `do stuff` block
- it structures functions to immediately express when an exit/return is needed, which makes that flow condition a bit more readable, instead of having to read through the whole function -- this here is just a general observation, not related to your example at all

About this "pattern":
- [link 1](https://medium.com/swlh/return-early-pattern-3d18a41bba8#:~:text=Return%20early%20is%20the%20way,when%20conditions%20are%20not%20met.)
- [link 2](https://forum.freecodecamp.org/t/the-return-early-pattern-explained-with-javascript-examples/19364)
- [link 3](https://medium.com/better-programming/are-early-returns-any-good-eed4b4d03866)

adekbadek:
Thank you both for comments, I've addressed them all in 8a4f5986a76fbdd52b47a75da67b3b99429ba58a

claudiulodro (approved):
Latest version looks good 👍 I left one question about the hook this is hooked into, but I think the concern is low risk.

includes/class-newspack-popups.php
```diff
@@ -41,6 +43,7 @@ public static function instance() {
 	 * Constructor.
 	 */
 	public function __construct() {
+		add_action( 'init', [ __CLASS__, 'create_lightweight_api_config' ] );
```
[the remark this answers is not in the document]
adekbadek: I'd be concerned if that's fired on plugin update 🤔 Also following the advice [here](https://core.trac.wordpress.org/ticket/14170#comment:68), I've modified it so there's just one chance to create the config file: 3816b28e252c632b294de7ee9dc814a01bb699b0
Is that ok or too restrictive?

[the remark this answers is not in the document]
claudiulodro: That looks like a nice solution. 👍 I was thinking further, and since the activation hook wouldn't fire on plugin update we'd have had to deactivate then reactivate the plugin on all of our sites if we went with my suggestion. This latest approach should work nicely, and we won't have to do anything extra after deploying it. Good idea!

claudiulodro (approved):

jeffersonrabb (approved):
Before the next release we should do a follow-up PR that addresses the case of a hosting environment where the DB credentials are not stored in environment variables, or PHP cannot write to the file system. This can be addressed by adding detail to the README about creating the file manually and possibly displaying a `wp-admin` notification if the config is absent or incomplete.

adekbadek:
@jeffersonrabb 

>  This can be addressed by adding detail to the README about creating the file manually and possibly displaying a wp-admin notification if the config is absent or incomplete.

Added in c0f9fc3242aeafe58f33b78b58a35e8ddb8e5917

includes/class-newspack-popups.php
```diff
@@ -388,5 +392,50 @@ public static function popup_default_fields( $post_id, $post, $update ) {
...
+
+	/**
+	 * Add an admin notice if config is missing.
+	 */
+	public static function api_config_missing_notice() {
+		if ( file_exists( self::LIGHTWEIGHT_API_CONFIG_FILE_PATH ) ) {
+			return;
+		}
+		?>
+			<div class="notice notice-error">
```
jeffersonrabb, line 430: Slight rewrite:

```
<div class="notice notice-error">
				<p>
					<?php _e( 'Newspack Campaigns requires a custom configuration file, which is missing. Please create this file following instructions found ', 'newspack-popups' ); ?>
					<a href="https://github.com/Automattic/newspack-popups/blob/master/api/README.md">
						<?php _e( 'here.', 'newspack-popups' ); ?>
					</a>
				</p>
			</div>
```

jeffersonrabb (approved):
Commented on a slight rewrite of the notification, then we're good to go!

includes/class-newspack-popups.php
adekbadek: Done in d4d9a2b82c5f97ce009aa3b5fb5719d5e58b34c0

PR #223, opened 2020-09-18
Prevent a test category popup from taking the place of a sitewide default

### LINKED ISSUE #219: [bug] `retrieve_category_overlay_popup` fetches test campaign even if not logged in
If you have an overlay campaign with a matching category that is set to "test" frequency, the `retrieve_category_overlay_popup` method will still retrieve this campaign even if viewing the site as a non-logged-in user. If there is an active sitewide campaign that should be shown instead, this results in no overlay campaign being shown.

To replicate:

1. Create two overlay campaigns: 
  - Campaign A should be set to "Sitewide default" and "Once a day" frequency. 
  - Campaign B should be set to a category (e.g. "News") and "Test" frequency.
2. In a new incognito window, view a post with the category matching Campaign B.
3. Observe that no overlay campaign is displayed and that Campaign B markup is inserted into the post content instead of Campaign A.

### Expected behavior:

In this case, while viewing the post as a non-logged-in user, I would expect Campaign A to be inserted into the post content and displayed, since Campaign B is a test campaign.

adekbadek:
### All Submissions:

Closes #219.

1. Observe #219 is not reproducible

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -49,7 +49,7 @@ final class Newspack_Popups_Inserter {
 
 		// Check if there's an overlay popup with matching category.
 		$category_overlay_popup = Newspack_Popups_Model::retrieve_category_overlay_popup();
-		if ( $category_overlay_popup ) {
+		if ( $category_overlay_popup && self::should_display($category_overlay_popup) ) {
 			array_push(
 				$popups_to_maybe_display,
 				$category_overlay_popup


dkoo (approved):
Tested with two campaigns:

* Campaign A set to "Sitewide default" and "Once a day" frequency.
* Campaign B set to "News" category and "Test" frequency.

- [x] Visiting a News post in a new incognito window, Campaign A appeared as expected.
- [x] Visiting the same post while logged in, Campaign B appeared as expected.
- [x] After changing Campaign B to "active" and visiting a News post in a new incognito window, Campaign B appeared as expected.

Looks good! 🚢

PR #224, opened 2020-09-21
Set up segmentation features

adekbadek:
### All Submissions:

Sets up client visits tracking to enable campaign segmentation. For starters, adding a minimum posts read count option.

This is ready for review, but read on for some high-level questions which might influence large changes.

1. Create a new campaign, observe a new "Campaign Segmentation" panel in the sidebar
2. Set the "Min. no. of posts read until displayed" setting to 3
3. Visit the site as a non-logged in user, observe that the campaign is displayed only upon the third article reading

#### 1. Data stored in the DB

There are now three tables with which this plugin interacts – options, and the two custom tables set up in this PR. A single client visit (assuming there are campaigns on the page) will write to four rows (at most):

1. `<client-id>-popups` transient in options table – to store data about client's interaction with campaigns in general (e.g. has dismissed _a_ newsletter campaign)
2. `<client-id>-<campaign-id>-popup` transient in options table – to store information about a client's interaction with a specific campaign (e.g. last seen date)
3. a row in the clients custom table – the data about the client, which will serve later for segmentation, e.g. information that a client is a newsletter subscriber
4. a row in the visits custom table – data about a single visit

I was thinking about two ways to make it all lighter on the DB:

1. Move the transients data to the custom clients table, making them two additional columns on a client
2. Something suggested by @jeffersonrabb – using just **one custom table**, with visits, donations, newsletter subscriptions, single popup-related data – all as rows.  
One issue with this approach is the mapping of client IDs to user IDs – reconciliation of data from different clients. This is a different topic, so I won't get into details here, but I think that this mapping could be maintained on the WP side – storing the client ID(s) as user meta. The API would have to query the users table on a logged in user's visit, but maybe the cost of that is not an issue.
The "client meta" information, like "has dismissed a newsletter popup" information could be queried form this table, but I don't know what's the performance cost here (querying for all newsletter campaigns seen by a client to determine if at least one was dismissed vs. reading just one "client meta" row)
Update: this approach is pursued in #233

#### 2. DB config

The custom tables configuration – do the column types (which determine the size) make sense? 

#### 3. Should this be an opt-in feature?

The visits-reporting will require a lot of data to be stored, and some users may not need segmentation. But maybe we should revisit that question later if such problems arise.

### Other information:


dkoo (approved):
From a functional testing standpoint, this works exactly as described. Also works across AMP and non-AMP sessions. I don't think I'm qualified to comment on the DB questions, so I'll let one of the other reviewers respond to that.

api/campaigns/class-maybe-show-campaign.php
```diff
@@ -81,6 +88,20 @@ public function should_campaign_be_shown( $client_id, $campaign, $settings, $ref
 				break;
 		}
 
+		$min_posts_read = (int) $campaign->min;
```
jeffersonrabb, line 91: Even though the `min_posts` feature is a great proof of concept for this work, I think we should remove it before this merges, for two reasons:

1) UI thrash—we're still in the midst of considering UI for Audience features, so if we add this feature now it will probably be removed or changed very quickly, leading to confusion. 
2) Once we merge and deploy this branch we'll monitor closely for performance concerns. If we do encounter issues and need to revert and reconsider, it will be much tougher to do if real-world users have begun relying on the min posts feature. If the feature is effectively "silent" the only consequence of reversion will be less historical data. 

Removing this feature will also simplify the PR quite a bit, as you can drop all the React work and a few blocks of PHP code.

api/setup.php
```diff
@@ -1,16 +1,16 @@
 <?php
 /**
- * Route Newspack Campaigns API requests based on method.
+ * Setup Campaigns API.
  *
  * @package Newspack
  */
 
-$plugin_path = substr($_SERVER['SCRIPT_FILENAME'], 0, strrpos($_SERVER['SCRIPT_FILENAME'], 'wp-content/plugins/')); // phpcs:ignore
+$wp_root_path = substr($_SERVER['SCRIPT_FILENAME'], 0, strrpos($_SERVER['SCRIPT_FILENAME'], 'wp-content/plugins/')); // phpcs:ignore
```
jeffersonrabb, line 8: This issue is a holdover from past PRs but we should go more specific with the PHPCS ignore rules. Using blanket ignore rules makes it easy to miss basic formatting issues, like spaces after parentheses in this line.

includes/class-newspack-popups-segmentation.php
```diff
@@ -0,0 +1,185 @@
...
+			}
+			wp_enqueue_script( $script );
+		}
+	}
+
+	/**
+	 * Insert amp-analytics tracking code.
+	 * Has to be included on every page to set the cookie.
+	 */
+	public function insert_amp_analytics() {
```
jeffersonrabb, line 109: Sorry for being thick, but I'm still a bit unclear on why we need to make this `amp-analytics` `POST`, as opposed to using the `amp-access` `GET` to write the `post_read` data to the the DB (or log file after https://github.com/Automattic/newspack-popups/pull/233 is merged in)? Reducing these down to one would obviously be a big performance win.

src/editor/SegmentationSettings.js
```diff
@@ -0,0 +1,50 @@
+/**
```
jeffersonrabb, line 1: Noted elsewhere, I think all of the Minimum Number of Posts UI should be removed for now.

includes/class-newspack-popups-model.php
```diff
@@ -890,16 +892,12 @@ public static function container_style( $popup ) {
 	}
 
 	/**
-	 * Endpoint to report Pop-up data.
+	 * Endpoint to handle Pop-up data.
 	 *
 	 * @return string Endpoint URL.
 	 */
 	public static function get_reader_endpoint() {
-		return str_replace(
```
jeffersonrabb, line 898: To be honest I don't recall why we had this replacement to the double-slash scheme, but there probably was a good reason. Any particular reason why you removed?

jeffersonrabb (changes requested):
I tried to limit this review to code questions that aren't present in 233, hopefully I didn't make this more complicated than necessary.

includes/class-newspack-popups-model.php
adekbadek: I've removed it because it does not have any effect on sites served over HTTPS, this code would only change a URL with `http://`. It just looked weird and I'm not sure AMP would even talk to an endpoint not served over HTTPS. So I don't see a good reason for this replacement, but maybe I'm missing something here.

src/editor/SegmentationSettings.js
```diff
@@ -0,0 +1,50 @@
...
+			},
+		};
+	} ),
+] )( SegmentationSettings );
+
+registerPlugin( 'newspack-popups-segmentation-settings', {
+	render: () => (
+		<PluginDocumentSettingPanel
+			name="popup-segmentation-settings-panel"
+			title={ __( 'Campaign Segmentation' ) }
```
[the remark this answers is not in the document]
adekbadek: Done in 80e2e975028a1629d2951e175691aa7683ce1770

api/campaigns/class-maybe-show-campaign.php
adekbadek: Agreed, done in 4c9b9149a365c44db8c156af6672fd7af2ce69d4

api/setup.php
adekbadek: Done in e5a08a65cce36564e863cd12cc05c85404139681

src/editor/SegmentationSettings.js
adekbadek: Done in 4c9b9149a365c44db8c156af6672fd7af2ce69d4

includes/class-newspack-popups-model.php
jeffersonrabb: > I've removed it because it does not have any effect on sites served over HTTPS, this code would only change a URL with `http://`. It just looked weird and I'm not sure AMP would even talk to an endpoint not served over HTTPS. So I don't see a good reason for this replacement, but maybe I'm missing something here.

Yeah, I reviewed and couldn't find a good reason. Good call.

includes/class-newspack-popups-segmentation.php
adekbadek: While trying to reduce it to one request I've found our that we have a problem here anyway. It gets complicated so here's a tl;dr:

> When navigating from proxied site (AMP Cache) to origin site, the **first request** to `amp-access`' authorisation endpoint will receive a **wrong client ID**. Subsequent requests will be ok.

In practice, this means that some segmentation features will not kick in on this single request, because `amp-access` will be querying wrong data (no data, to be exact).

It boils down to:

## Cross-context linking and interoperation between `amp-access` and `amp-analytics`. 

Considering the following scenario:
1. User visits the page served from AMP Cache (proxied site)
2. User clicks a link and thus navigates to the origin site

For `amp-access` these will be visits from **two separate clients**, because a proxied site's cookie cannot be read by the origin site. Now, this same problem is resolved by `amp-analytics` with [`linker` configuration](https://amp.dev/documentation/components/amp-analytics/?format=websites#linkers). With it, any link on the proxied site gets a special param appended, with an encoded client ID. On loading the origin site, `amp-analytics` reads this param and sets it in a cookie. 

The issue with interoperability between these two AMP components I've run into manifests itself like this:

1. Page is loaded from AMP Cache. Client is assigned a client id, let's call it `id-A`. This is encoded and appended to URLs leading to origin site.
2. User navigates to the origin site. 
3. `amp-access`, seeing no client ID in site's cookies, assigns a client ID `id-B` and uses it to call the authorisation endpoint
4. `amp-analytics` reads the linker URL param, sets the `id-A` in cookie
5. Subsequent calls to `amp-access`' endpoint are made with `id-A` client ID (correctly), because it's read from the cookie

I _guess_ this issue could be resolved by either: 
1. Reversing the order of events – 4. first, 3. then. However I was not able to do that, because `amp-analytics` is placed in the document body, while `amp-access` in document head
2. Having that `linker` feature in `amp-access`

## What would change with a single request

This first page load after navigating from AMP Cache – now serving in place of analytics – will assign this visit to a wrong client ID. This means that this page load will be "lost", and because this client ID will not be used anymore, create a superfluous row in the DB. 

---

I'll research it a bit more, but if my conclusions here are correct, we have a situation where we have to choose an extra 
API request or extra rows in the DB. From a reader's perspective, the way it is now is better, because the data reported will be correct, even though they might get a different experience on that first-origin-page-load-after-amp-cache then intended by the publisher.

adekbadek: Added visit-reporting handling to the amp-access endpoint in e2b1f30de7841e1063c77fb18c578cde1b964482

iuravic (approved):
Folks, I haven't been commenting this PR very much, but I've been reading it, considering many aspects, commented with some privately, and tested it.

The underlying concepts seem really good to me, especially since we actually tried several (e.g. config, DB schema, read/write ideas, etc), and I believe we're on the right path with this. Congrats to @adekbadek and @jeffersonrabb for all the work, really nicely done! 👍 

I think that someone here needs to approve this, so I'm going to do the honors. Let's give it a spin in real life, and start monitoring first installs, particularly the log, the parsing, the DB size, and the API traffic and bandwidth.

api/segmentation/class-segmentation.php
```diff
@@ -0,0 +1,50 @@
...
+			},
+			$clients_events
+		);
+	}
+
+	/**
+	 * Get log file path.
+	 */
+	public static function get_log_file_path() {
+		return get_temp_dir() . 'newspack-popups-events.log';
```
jeffersonrabb, line 40: If there are multiple WordPress installations on the same server there will be a conflict. Suggest prepending the database prefix.

jeffersonrabb: https://github.com/Automattic/newspack-popups/pull/224/commits/1aee6100a63b2646d858541bd84618fbb2a735b3

dkoo (approved):
Tested once again on local and a new JN site. Verified with the log file that the reader client ID gets set and persists across AMP and non-AMP pages, and article views are logged correctly in both cases.

jeffersonrabb (approved):

PR #228, opened 2020-09-22
API: config creation fix

adekbadek:
### All Submissions:

Fixes two issues related to config:

1. The path to the config assumed the plugin folder is named `newspack-popups` – it can be named differently, e.g. when it's an alpha build it will have the version appended after plugin name
2. On certain environments for some reason PHP does not have access to the environment, so here the `DB_USER` etc. checks are removed

1. Remove the config file of present
2. Verify that the config file is created (you might need to remove the `newspack_has_tried_to_create_lightweight_api_config` option)

### Other information:

### DIFF
--- a/includes/class-newspack-popups.php
+++ b/includes/class-newspack-popups.php
@@ -403,10 +403,9 @@ final class Newspack_Popups {
 	 * Create the config file for the API, unless it exists.
 	 */
 	public static function create_lightweight_api_config() {
-		if ( get_option( 'newspack_has_tried_to_create_lightweight_api_config' ) ) {
+		if ( ! ( defined( 'ATOMIC_SITE_ID' ) && ATOMIC_SITE_ID ) ) {
 			return;
-		};
-		add_option( 'newspack_has_tried_to_create_lightweight_api_config', true );
+		}
 		if ( file_exists( self::LIGHTWEIGHT_API_CONFIG_FILE_PATH ) ) {
 			return;
 		}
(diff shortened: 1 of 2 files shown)


includes/class-newspack-popups.php
```diff
@@ -407,8 +407,7 @@ public static function create_lightweight_api_config() {
 			return;
 		};
 		add_option( 'newspack_has_tried_to_create_lightweight_api_config', true );
-		$has_db_config_in_env = getenv( 'DB_USER' ) && getenv( 'DB_PASSWORD' ) && getenv( 'DB_NAME' ) && getenv( 'DB_HOST' );
-		if ( ! $has_db_config_in_env || file_exists( self::LIGHTWEIGHT_API_CONFIG_FILE_PATH ) ) {
+		if ( file_exists( self::LIGHTWEIGHT_API_CONFIG_FILE_PATH ) ) {
```
jeffersonrabb, line 410: Doesn't this change effectively mean the config file will  be installed in all environments, not just Atomic?

includes/class-newspack-popups.php
```diff
@@ -407,8 +407,7 @@ public static function create_lightweight_api_config() {
 			return;
 		};
 		add_option( 'newspack_has_tried_to_create_lightweight_api_config', true );
-		$has_db_config_in_env = getenv( 'DB_USER' ) && getenv( 'DB_PASSWORD' ) && getenv( 'DB_NAME' ) && getenv( 'DB_HOST' );
```
jeffersonrabb, line 410: I'm finding `getenv( 'DB_USER' )` returns nothing but `getenv()['DB_USER']` properly returns the environment variable.

jeffersonrabb (changes requested):

jeffersonrabb (approved):
Looks good. Tested locally and on an Atomic site and the config was created only on Atomic.

PR #233, opened 2020-09-24
Segmentation: add log parsing

adekbadek:
This PR adds logs parsing to the #224 . In essence, the reported visits will not be inserted in the DB at once, but saved in a log file. In a cron job (running every minute here, but it should be increased once caching is added) the log file will be parsed, data added to the DB in bulk, and the log file emptied.

Testing should be done in the same way as in #224, but at least a minute should pass before completing the last step (visiting a post with read-count-restricted campaign). 

---

Also, changes the DB structure – the segmentation feature uses just one table ("events"). The separate clients table is not needed for the feature in #224, and the subsriber/donor segmentation can be achieved using this single events table.


includes/class-newspack-popups-segmentation.php
```diff
@@ -181,5 +184,36 @@ function( $cat ) {
...
+	 */
+	public static function create_database_table() {
+		global $wpdb;
+		$events_table_name = Segmentation::get_events_table_name();
+
+		// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
+		if ( $wpdb->get_var( $wpdb->prepare( 'SHOW TABLES LIKE %s', $events_table_name ) ) != $events_table_name ) {
+			$charset_collate = $wpdb->get_charset_collate();
+
+			$sql = "CREATE TABLE $events_table_name (
```
jeffersonrabb, line 199: The table structure should include indexes for all the fields that will appear in `WHERE` clauses, and if we do decide to enforce one row per `client_id`/`post_id` combo we'll need a multi-column `UNIQUE` key. Some good examples can be found in the `CREATE` statements for WP taxonomy tables: https://github.com/WordPress/WordPress/blob/d5b8d282e8c03929aafb1fa660181b7db9ad55d3/wp-admin/includes/schema.php#L65-L91

includes/class-newspack-popups-segmentation.php
```diff
@@ -181,5 +184,36 @@ function( $cat ) {
...
+				id bigint(20) NOT NULL AUTO_INCREMENT,
+				created_at date NOT NULL,
+				-- type of event
+				type varchar(20) NOT NULL,
```
jeffersonrabb, line 203: It would make sense to add a default of `post_read` for the `type` column to avoid adding that to all `INSERT`s.

jeffersonrabb (changes requested):

includes/class-newspack-popups-segmentation.php
```diff
@@ -181,5 +184,36 @@ function( $cat ) {
...
+				-- Unique id of a device/browser pair
+				client_id varchar(100) NOT NULL,
+				-- Article ID
+				post_id bigint(20),
+				-- Article categories IDs
+				categories_ids varchar(100),
```
jeffersonrabb, line 209: Suggest changing column name to `category_ids` or `categories`.

jeffersonrabb (changes requested):

api/segmentation/class-segmentation-report.php
```diff
@@ -19,83 +19,47 @@ class Segmentation_Report extends Lightweight_API {
...
-			$wpdb->prepare( "SELECT * FROM $visits_table_name WHERE post_id = %s AND client_id = %s", $post_id, $client_id ) // phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared
-		);
-		if ( null === $existing_post_visits ) {
-			$wpdb->insert( // phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery
-				$visits_table_name,
+		$is_post = $payload['is_post'];
+		if ( '1' === $is_post ) {
+			// Add line to log file.
+			$line = implode(
+				';',
```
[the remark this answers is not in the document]
adekbadek: It is – AMP client IDs are `base64url`-encoded ([AMP docs](https://amp.dev/documentation/guides-and-tutorials/optimize-and-measure/configure-analytics/analytics_basics/#user-identification), [encoding docs on "base64" vs "base64url"](https://tools.ietf.org/html/rfc4648#section-5))

includes/class-newspack-popups-parse-logs.php
```diff
@@ -0,0 +1,176 @@
...
+	 */
+	public static function bulk_db_insert( $table_name, $rows, $column_names, $placeholder ) {
+		if ( 0 === count( $rows ) ) {
+			return;
+		}
+
+		global $wpdb;
+
+		$column_names = implode( ', ', $column_names );
+		$query        = "INSERT INTO $table_name ($column_names) VALUES ";
```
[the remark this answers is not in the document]
adekbadek: According to this SO post, we can insert `13107` rows (there are 5 placeholders per row). With cron running every 15 minutes, that gives us ~14 events per second. With one event per pageview, our upper limit is about `1.2M` PVs per day (with cron running every 5 minutes, it's ~43 events, `3.7M` PVs/day). 
Note that not all pageviews will log events (only articles), so that gives us an even bigger margin of error. If these calculations are correct, I think we're safe.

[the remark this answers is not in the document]
jeffersonrabb: Sounds good. Perhaps as a later enhancement we can insert them in groups of 10k max to be ultra-safe. Agreed we're safe for now.

includes/class-newspack-popups-segmentation.php
adekbadek: True, but in the future there will be other event types, and this would introduce a lot of forking in `parse_events_logs` and two separate calls to `bulk_db_insert` method – one for read posts (omitting the event type column) and one for other events. I don't think adding that complexity is worth it.

api/segmentation/class-segmentation.php
```diff
@@ -12,11 +12,14 @@
...
 	 */
-	public static function init() {
-		self::create_segmentation_tables();
-	}
+	const IS_PARSING_FILE_PATH = WP_CONTENT_DIR . '/../.is-parsing';
+
+	/**
+	 * The log file path.
+	 */
+	const LOG_FILE_PATH = WP_CONTENT_DIR . '/../newspack-popups-events.log';
```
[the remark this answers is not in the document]
adekbadek: Fixed in 197f1dc6fbbdbc0967fe2d47ba44a9cc5f49fbcc

api/campaigns/class-report-campaign-data.php
```diff
@@ -22,7 +22,7 @@ class Report_Campaign_Data extends Lightweight_API {
 	 */
 	public function __construct() {
 		parent::__construct();
-		if ( ! $_POST && ! $_GET ) { // phpcs:ignore
+		if ( empty( $this->get_post_payload() ) ) {
```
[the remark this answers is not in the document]
adekbadek: Fixed in 44ab1f329630f7a47ca29d0e96c914b64081fd35

api/segmentation/class-segmentation-report.php
```diff
@@ -19,83 +19,47 @@ class Segmentation_Report extends Lightweight_API {
...
-				'updated_at' => gmdate( 'Y-m-d' ),
-			];
-			// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
-			$wpdb->update(
-				$clients_table_name,
-				$updates,
-				[ 'ID' => $found_client->id ]
-			);
+	public static function api_handle_post_read( $payload ) {
+		if ( file_exists( Segmentation::IS_PARSING_FILE_PATH ) ) {
```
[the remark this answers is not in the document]
adekbadek: > Couldn't we just use PHP locking?

Now that I've learned about file locking, we certainly can :D 09868736672cc79251f423feeefae674863b9f06

includes/class-newspack-popups-parse-logs.php
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		add_filter( 'cron_schedules', [ __CLASS__, 'add_cron_interval' ] ); // phpcs:ignore WordPress.WP.CronInterval.CronSchedulesInterval
+		add_action( 'newspack_popups_segmentation_cron_hook', [ __CLASS__, 'parse_events_logs' ] );
+		if ( ! wp_next_scheduled( 'newspack_popups_segmentation_cron_hook' ) ) {
+			wp_schedule_event( time(), 'every_minute', 'newspack_popups_segmentation_cron_hook' );
```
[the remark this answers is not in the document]
adekbadek: Good idea, done in d961bf8358f102f3096ec5db321c636c708a5a7a

includes/class-newspack-popups-parse-logs.php
```diff
@@ -0,0 +1,176 @@
...
+		$lines    = [];
+
+		while ( ! feof( $log_file ) ) {
+			$line = trim( fgets( $log_file ) );
+			if ( ! empty( $line ) ) {
+				$lines[] = $line;
+			}
+		}
+
+		error_log( 'Parsing ' . count( $lines ) . ' lines of log file: ' . Segmentation::LOG_FILE_PATH ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
```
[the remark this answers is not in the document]
adekbadek: They were for debugging, removed in 736df267ec27252e93d73447d1650065c6e8bf2e

includes/class-newspack-popups-parse-logs.php
```diff
@@ -0,0 +1,176 @@
...
+
+		foreach ( $lines as $line ) {
+			$result     = explode( ';', $line );
+			$event_type = $result[0];
+			$client_id  = $result[1];
+			$date       = $result[2];
+			$post_id    = $result[3];
+			$categories = $result[4];
+
+			$existing_post_events = $wpdb->get_row( // phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
```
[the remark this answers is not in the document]
adekbadek: 1., 2. Seems like an edge case to me. This functionality has been following what Memberkit/NCI do so far – where the basic metric of brand loyalty is number of articles read in the last 30 days. I've [added this suggestion as an issue](https://github.com/Automattic/newspack-popups/issues/234) to maybe pursue it at a later date.
3. Done in b1c6e4a645776590017cb7e684dbc43db1b151fd

includes/class-newspack-popups-segmentation.php
adekbadek: Done in b1c6e4a645776590017cb7e684dbc43db1b151fd, along with `INSERT ON DUPLICATE KEY UPDATE`

includes/class-newspack-popups-segmentation.php
adekbadek: Done in 27d76d966b45a1806853eb1d5a98872d9678eb98

includes/class-newspack-popups-parse-logs.php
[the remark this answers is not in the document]
jeffersonrabb: Definitely an edge case. Let's get Brian's feedback on the issue and consider whether to change in the future.

api/segmentation/class-segmentation.php
```diff
@@ -12,11 +12,9 @@
  */
 class Segmentation {
 	/**
-	 * Initialize.
+	 * The log file path.
 	 */
-	public static function init() {
-		self::create_segmentation_tables();
-	}
+	const LOG_FILE_PATH = '/tmp/newspack-popups-events.log';
```
jeffersonrabb, line 17: I'm not sure we can reliably assume the existence of this directory. Also, this could create chaos in situations where there are multiple WordPress installations hosted on the same server, and multisite too. I think we need to [programmatically get the path to the log file](https://stackoverflow.com/questions/8955411/find-out-the-error-logs-path), then add elements to the filename to ensure uniqueness. I think the database prefix will do the trick. As for the multisite scenario, I'm not sure how best to address.

includes/class-newspack-popups-parse-logs.php
```diff
@@ -0,0 +1,169 @@
...
+		$wpdb->query( $sql ); // phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching, WordPress.DB.PreparedSQL.NotPrepared
+	}
+
+	/**
+	 * Parse the log file, write data to the DB, and remove the file.
+	 */
+	public static function parse_events_logs() {
+		global $wpdb;
+
+		$start_time        = microtime( true );
```
jeffersonrabb, line 111: Believe this was there for logging, and can now be removed?

includes/class-newspack-popups-segmentation.php
```diff
@@ -181,5 +184,37 @@ function( $cat ) {
...
+				created_at date NOT NULL,
+				-- type of event
+				type varchar(20) NOT NULL,
+				-- Unique id of a device/browser pair
+				client_id varchar(100) NOT NULL,
+				-- Article ID
+				post_id bigint(20),
+				-- Article categories IDs
+				category_ids varchar(100),
+				UNIQUE KEY client_id_post_id (client_id, post_id),
```
jeffersonrabb, line 210: Won't this table also conceivably be queried based on `client_id` alone, `type` alone, and even `post_id` alone? If so, we may need more indexes.

jeffersonrabb (changes requested):

api/segmentation/class-segmentation.php
adekbadek: I've used [WP function `get_temp_dir`](https://developer.wordpress.org/reference/functions/get_temp_dir/) in f66afff4ed2a4fe91091fee9a3895e3f69bbc298 – does that look ok?

includes/class-newspack-popups-parse-logs.php
adekbadek: Yup, done in ed28d25aafa38f8704bf5db65935a502acbb238e

includes/class-newspack-popups-segmentation.php
adekbadek: This should do the trick: 7ac8d413e03537057fef2094a05452d95d9eaad6

jeffersonrabb (approved):
I think we're good to merge this into `add/segentation` and get that reviewed. Left one comment that I could kind of go either way on.

api/setup.php
```diff
@@ -37,6 +37,8 @@ function is_wp_error( $thing ) { return ( $thing instanceof WP_Error ); }
 if ( file_exists( ABSPATH . WPINC . '/wp-db.php' ) ) {
 	require_once ABSPATH . WPINC . '/wp-db.php';
 	require_once ABSPATH . WPINC . '/functions.php';
+	// Needed for get_temp_dir function.
+	require_once ABSPATH . WPINC . '/formatting.php';
```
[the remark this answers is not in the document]
adekbadek: That's true, added in 697cc1e8c4f02910347dafdc39a527f7618773cc

PR #240, opened 2020-09-30
Fix: scroll-triggered popups on non-AMP pages

adekbadek:
### All Submissions:

* [X] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?

AMP Components can be self-closing, but if page is served as non-AMP, in some cases this may not be handled. 
In the case of this bug, the `amp-position-observer` component can be treated as not-closed, resulting in the page content being inserted as children of it. 

1. On `master`, create an overlay, scroll-triggered campaign, set to sitewide default 
2. Set your site to AMP transitional mode
3. Visit the (non-AMP) homepage, observe only the header and footer are displayed
3. Visit a (non-AMP) post, observe only the header and footer are displayed
4. Switch to this branch, reload site
5. Observe the homepage and post are rendered as expected, with the campaign triggered by scroll as expected
6. Also check the AMP version

### Other information:

### DIFF
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -789,7 +789,7 @@ final class Newspack_Popups_Model {
 			</form>
 		</amp-layout>
 		<div id="page-position-marker" style="position: absolute; top: <?php echo esc_attr( $popup['options']['trigger_scroll_progress'] ); ?>%"></div>
-		<amp-position-observer target="page-position-marker" on="enter:showAnim.start;" once layout="nodisplay" />
+		<amp-position-observer target="page-position-marker" on="enter:showAnim.start;" once layout="nodisplay"></amp-position-observer>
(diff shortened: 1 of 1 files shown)


jeffersonrabb (approved):
I verified the fix on all Newspack sites that were affected:

- https://rabberson.newspackstaging.com/
- https://hechingerreport.org/
- https://southernmarylandchronicle.com/

@adekbadek resolved a CSS issue leading to clipped header on https://sahanjournal.com/. This issue was actually due to the change in https://github.com/Automattic/newspack-popups/pull/236, not this PR, but I'm noting it here because all of these issues initially appeared to be related and will be resolved by the release which will contain this PR.

PR #266, opened 2020-10-22
Fix previews if latest post has campaigns disabled

### LINKED ISSUE #110: Preview vs disabled popups
If the most recent post has disabled popups, the inline popup preview will be broken – as it uses the most recent post. 
[Reported](https://github.com/Automattic/newspack-popups/pull/108#pullrequestreview-399857949) on #108

adekbadek:
### All Submissions:

Closes #110.

1. Disable campaigns on the latest post
2. Create or edit an inline campaign, observe that preview uses the lastest post which does not have disabled campaigns

### Other information:

### DIFF
--- a/includes/class-newspack-popups.php
+++ b/includes/class-newspack-popups.php
@@ -272,20 +272,34 @@ final class Newspack_Popups {
 			filemtime( dirname( NEWSPACK_POPUPS_PLUGIN_FILE ) . '/dist/editor.js' ),
 			true
 		);
-		$recent_posts = wp_get_recent_posts(
+
+		$query                  = new WP_Query(
 			[
-				'numberposts' => 1,
-				'post_status' => 'publish',
-			],
-			OBJECT
+				'posts_per_page' => 1,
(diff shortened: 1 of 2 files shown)


dkoo (changes requested):
Due to [this logic](https://github.com/Automattic/newspack-popups/blob/master/src/editor/PopupPreview.js#L22-L27), if you disable campaigns on the homepage and attempt to preview a time-triggered overlay campaign, you will still see the homepage in the preview but will not be able to see the campaign. Should we consider showing the most recent post without the `newspack_popups_has_disabled_popups` option in this case?

adekbadek:
@dkoo I've simplified it so the preview always uses the latest post (21190bc6b9de9f2748fcf0072962b4426e9bd159). My thinking is that since more users will visit an article than homepage, this is after all more representative.

dkoo (approved):
> My thinking is that since more users will visit an article than homepage, this is after all more representative.

That works for me. 👍

PR #267, opened 2020-10-22
Add tag filtering

### LINKED ISSUE #109: Allow filtering based on tag
A request (360139597972-zen) has come in to add tag-based filtering alongside the category filtering already available.

Different publishers use tags and categories in different ways. Overall though, categories are more likely to be limited, because of their prominent display at the top of posts and in homepage posts blocks. Tags on the other are less visible and given very little prominence so using them for things that may not make a lot of sense to normal visitors is much more viable.

Having tag-based filtering would allow more targeted engagement strategies.

adekbadek:
### All Submissions:

Closes #109.

1. Create or edit a campaign, observe that tags can be chosen in the sidebar
2. Set a tag on the campaign, observe that it's displayed only on posts with matching tag (just like with category filtering)
3. Observe a campaign with no tags set behaves as before

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -478,6 +478,24 @@ final class Newspack_Popups_Inserter {
 	}
 
 	/**
+	 * If Pop-up has tags, it should only be shown on posts/pages with those.
+	 *
+	 * @param object $popup The popup to assess.
+	 * @return bool Should popup be shown based on tags it has.
+	 */
+	public static function assess_tags_filter( $popup ) {
+		$post_tags  = get_the_tags();
+		$popup_tags = get_the_tags( $popup['id'] );
+		if ( $popup_tags && count( $popup_tags ) ) {
+			return array_intersect(
+				array_column( $post_tags, 'term_id' ),
+				array_column( $popup_tags, 'term_id' )
+			);
+		}
+		return true;
+	}
+
+	/**
 	 * Should Popup be rendered, based on universal conditions.
 	 *
 	 * @param object $popup The popup to assess.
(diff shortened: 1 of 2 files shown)


dkoo (changes requested):
Works as described, but getting a PHP warning when viewing a post that lacks any tags:

```php
PHP Warning:  array_column() expects parameter 1 to be array, bool given in /Users/dkoo/Local Sites/newspack/app/public/wp-content/plugins/newspack-popups/includes/class-newspack-popups-inserter.php on line 491
```

includes/class-newspack-popups-inserter.php
```diff
@@ -477,6 +477,24 @@ public static function assess_categories_filter( $popup ) {
...
+	 *
+	 * @param object $popup The popup to assess.
+	 * @return bool Should popup be shown based on tags it has.
+	 */
+	public static function assess_tags_filter( $popup ) {
+		$post_tags  = get_the_tags();
+		$popup_tags = get_the_tags( $popup['id'] );
+		if ( $popup_tags && count( $popup_tags ) ) {
+			return array_intersect(
+				array_column( $post_tags, 'term_id' ),
```
dkoo, line 491: Maybe this will avoid the error?

```suggestion
				array_column( ! empty( $post_tags ) ? $post_tags : [], 'term_id' ),
```

adekbadek: Good catch! Fixed in 5e6d0490a5ab3d81fd09bcc29b5636d3309fe88f
Added a check for categories, too. I was unable to make a post category-less, but this is just for consistency.

dkoo (approved):
That did the trick!

PR #283, opened 2020-10-28
Fix non-AMP issues

### LINKED ISSUE #200: Setting form submission status does not work on non-AMP pages
When a newsletter signup form (Mailchimp) block is a part of a Campaign, submitting the form will send an event to the plugin, which will mark this Campaign and prevent it from showing up for this client later (because they've already subscribed). This does not work on non-AMP pages, though. 

---

I've tried setting the `on` value of the `formSubmitSuccess` trigger to `amp-form-submit`, but no dice, because `amp-form` is not enqueued on the front end (for a good reason!). This should be handled [as in the `newspack-plugin`](https://github.com/Automattic/newspack-plugin/blob/master/includes/class-analytics.php#L528-L555), but perhaps we can avoid repetition somehow.

### LINKED ISSUE #257: Dismissals not working on non-AMP pages
Because the `amp-form` script is not loaded (on purpose), this form value substitution will not work:

https://github.com/Automattic/newspack-popups/blob/bc26e7b0d57d10bf8027c7cad740333e305b1af7/includes/class-newspack-popups-model.php#L928

Because of this, the dismissal will not be attributed to a client ID, since the ID will not be sent. 

## How to reproduce

1. Visit a non-AMP page with an inline campaign with "Every page" frequency
2. Dismiss permanently ("I'm not interested" button)
3. Reload the page, observe the campaign is still visible

adekbadek:
### All Submissions:

Closes #200.
Closes #257.

1. Create a campaign with a newsletter signup form
2. On `master`, 
3. observe that this campaign cannot be permanently dismissed on a non-AMP page
3. observe that signing up for the newsletter does not dismiss the campaign permanently
4. Switch to this branch
5. Observe both issues resolved

### Other information:

### DIFF
--- a/src/view/index.js
+++ b/src/view/index.js
@@ -82,7 +82,7 @@ const manageForms = container => {
 			}, {} );
 			performXHRequest( {
 				url: form.attributes[ 'action-xhr' ].value,
-				data,
+				data: processFormData( data, form ),
 			} );
 			event.preventDefault();
 		} );
--- a/src/view/index.js
+++ b/src/view/index.js
@@ -9,20 +9,81 @@ import domReady from '@wordpress/dom-ready';
 import './style.scss';
 import './patterns.scss';
 
+const values = object => Object.keys( object ).map( key => object[ key ] );
+
+const performXHRequest = ( { url, data } ) => {
+	const XHR = new XMLHttpRequest();
+	XHR.open( 'POST', url );
+	XHR.setRequestHeader( 'Content-Type', 'application/x-www-form-urlencoded' );
+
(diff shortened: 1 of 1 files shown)


dkoo (approved):
This works as described:

1. On a non-AMP page, dismissing a campaign causes the campaign to not appear on subsequent refreshes or other page loads.
2. On a non-AMP page, submitting a Mailchimp newsletter signup form suppresses newsletter campaigns on future page loads according to the "Suppress all Newsletter campaigns if at least one Newsletter campaign was permanently dismissed" setting.

PR #286, opened 2020-10-28
fix: allow only one overlay campaign per page

jeffersonrabb:
### All Submissions:

Recent changes to the handling of Test Mode campaigns introduced two issues related to overlays:

- If there is a sitewide default campaign a logged in user will get two instances of the same popup layered on top of each other.
- If there are multiple popups a logged in user may get two overlapping popups. 
- Anonymous users won't have either of these problems, so this is purely an issue for site editors.

This PR addresses the problem by retaining only the first overlay campaign from all candidates. This may not be a perfect solution, as an admin may see an overlay that isn't the expected one, but for now this resolves the obvious problem. 

1. On `master`, create and publish a single overlay campaign, Sitewide default.
2. View a page on the site (logged in). Observe two versions of the overlay are stacked on top of each other (you will need to click the dismissal button twice). 
3. Create a second overlay campaign assigning a Category. 
4. Visit a post in that category (logged in). Observe both popups appear.
5. Switch to this branch and repeat. Observe only one overlay appears in either case.
6. Verify no change to the test mode display of Inline campaigns. 

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -80,10 +80,28 @@ final class Newspack_Popups_Inserter {
 			}
 		}
 
-		$popups_to_display = array_filter(
+		// Allow only one overlay campaign.
+		$has_overlay                     = false;
+		$popups_to_maybe_display_deduped = array_filter(
 			$popups_to_maybe_display,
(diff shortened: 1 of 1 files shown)


adekbadek (approved):
👍

PR #291, opened 2020-10-30
Improve preview post link

adekbadek:
### All Submissions:

* [X] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?

Adds a password check to preview post getting. 

1. Make the most recent post password-protected
2. On `master`, observe that the campaign preview will use homepage for previews
3. Switch to this branch, observe that the campaign preview will use the most recent not-password-protected post

### Other information:

### DIFF
--- a/includes/class-newspack-popups.php
+++ b/includes/class-newspack-popups.php
@@ -246,6 +246,34 @@ final class Newspack_Popups {
 	}
 
 	/**
+	 * Get preview post permalink.
+	 */
+	public static function preview_post_permalink() {
+		$query        = new WP_Query(
+			[
+				'posts_per_page' => 1,
+				'post_status'    => 'publish',
+				'has_password'   => false,
+				'orderby'        => 'post_date',
+				'order'          => 'DESC',
+				'meta_query'     => [ // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
+					'relation' => 'OR',
+					[
+						'key'     => 'newspack_popups_has_disabled_popups',
+						'compare' => 'NOT EXISTS',
+					],
+					[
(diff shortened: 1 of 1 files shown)


jeffersonrabb (changes requested):
I'm having trouble reproducing the error. 

- Created a new post, turned on password protection, published.
- Edited an overlay campaign and previewed. The preview was in the password-protected post.
- Tried the same for an inline campaign, same result.
- In the Newspack Campaigns Wizard, the password protected post is used for the inline campaign BUT the home is used for the overlay. 

Is this expected behavior?

adekbadek:
> Edited an overlay campaign and previewed. The preview was in the password-protected post.

That's not how it should be, and different than on my machine. It's odd since the WP query parameter is pretty simple:
https://github.com/Automattic/newspack-popups/pull/291/files#diff-c7efd1b032b601286ed42b644beb0443eaa33ce019c42dc068c443adaf1028fdR256

> In the Newspack Campaigns Wizard, the password protected post is used for the inline campaign BUT the home is used for the overlay. Any idea what might be wrong with the WP query here?

This is expected unless `newspack-popups` is on this PR's branch and `newspack-plugin` is on https://github.com/Automattic/newspack-plugin/pull/690 's branch

dkoo (approved):
This worked for me—created a password-protected post, edited both inline and overlay campaigns, the preview shown for both was the next (non-password-protected) post.

jeffersonrabb (approved):
Tried again, worked this time. I must have done something wrong the first time.

Issue #260, opened 2020-10-22
On-page inline campaign placements


adekbadek:
Inline campaigns can only be placed in an article. It was suggested that two placements outside of the article would be beneficial:

1. [x] Above the site header (#292)
2. [x] ~In the middle of the homepage~ _now possible with a shorcode_
2. [ ] On article page, before the headline ([requested here](https://github.com/Automattic/newspack-popups/issues/209#issuecomment-689032193))

PR #292, opened 2020-10-30
Add above header placement option

adekbadek:
### All Submissions:

Adds an "above header" campaign placement, as requested in #260

Additional changes in Theme (https://github.com/Automattic/newspack-theme/pull/1131) and the plugin (https://github.com/Automattic/newspack-plugin/pull/691)

1. Create a new campaign, observe a new option in "Placement" select – "Above site header"
2. Use this option
3. Verify that the campaign is placed above site header
4. Filtering (category, tag, etc.) and everything else should work as with a regular inline campaign

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -42,8 +42,11 @@ final class Newspack_Popups_Inserter {
 			return [ Newspack_Popups_Model::retrieve_preview_popup( Newspack_Popups::previewed_popup_id() ) ];
 		}
 
-		// 1. Get all inline popups.
-		$popups_to_maybe_display = Newspack_Popups_Model::retrieve_inline_popups();
+		// 1. Get all inline, above header popups.
+		$popups_to_maybe_display = array_merge(
+			Newspack_Popups_Model::retrieve_inline_popups(),
+			Newspack_Popups_Model::retrieve_above_header_popups()
+		);
 
 		// 2. Get the overlay popup/s. There can be only one displayed, unless in test mode.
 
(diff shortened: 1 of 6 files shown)


laurelfulford:
I happened to notice while testing #1131 that the above header popup seems to appear twice on pages like the search results and archives, once above the header and once below: 

image

I haven't tested extensively, but the above is in test mode and appears with and without AMP enabled; it displays only once on single posts and pages.

dkoo (changes requested):
Looks like the `before_header` hook is firing twice on category archive pages, resulting in the "above header" campaigns appearing 2x each.

dkoo:
Another weird thing I noticed is that when not logged in, I'm seeing extra `<br>` tags being added inside certain label/button elements in the Homepage Posts block. I haven't figured out where they're coming from, but it only happens when this branch is active.

adekbadek:
> the above header popup seems to appear twice on pages like the search results and archives

Thanks for catching that, should be fixed by 2d0ede60d0e41e6546a94c0855981262f66f34fd 👍

adekbadek:
> Another weird thing I noticed is that when not logged in, I'm seeing extra <br> tags being added inside certain label/button elements in the Homepage Posts block.

Whoa, that's a good catch! Looks like re-building the post content without any campaigns caused this issues. Fixed in 35ff3ac66099443468886242c83628d5b7582ade

dkoo (approved):
Looks good to me now!

adekbadek:
So now this waits for theme changes – https://github.com/Automattic/newspack-theme/pull/1131

adekbadek:
(Should be release in tandem with https://github.com/Automattic/newspack-plugin/pull/691, awaiting review there)

dkoo (approved):

PR #294, opened 2020-11-02
Handle subscription, donation segmentation

### LINKED ISSUE #249: Segmentation Feature: Donor Status (WC)
Segmentation based on WC donor status. 3rd party donors handling extracted to #293

### LINKED ISSUE #250: Segmentation Feature: Mailing List Status
There is some campaign suppression logic in place that identifies anonymous IDs as mailing list subscribers but there are some pieces missing.

adekbadek:
### All Submissions:

Closes #249
Closes #250

_Follow testing steps in https://github.com/Automattic/newspack-plugin/pull/693_

### Other information:

* [X] Have you added an explanation of what your changes do and why you'd like us to include them?

### DIFF
--- a/api/campaigns/class-maybe-show-campaign.php
+++ b/api/campaigns/class-maybe-show-campaign.php
@@ -173,6 +173,8 @@ class Maybe_Show_Campaign extends Lightweight_API {
 		if ( ! empty( $campaign_segment ) ) {
 			$client_data      = $this->get_client_data( $client_id );
 			$posts_read_count = count( $client_data['posts_read'] );
+			$is_subscriber    = ! empty( $client_data['email_subscriptions'] );
+			$is_donor         = ! empty( $client_data['donations'] );
 			if (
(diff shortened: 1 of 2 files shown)


dkoo (changes requested):
I'm getting an error from the amp-access request which prevents all campaigns from displaying:

```
Notice: Undefined property: stdClass::$is_subscribed in /Users/dkoo/Local Sites/newspack/app/public/wp-content/plugins/newspack-popups/api/campaigns/class-maybe-show-campaign.php on line 189
```

adekbadek:
> I'm getting an error from the amp-access request which prevents all campaigns from displaying […] The error is thrown because it can't find the deleted segment which is still associated with the campaign post.

I think it's something else – a not-found segment will be handled gracefully. The issue is that there is a legacy segment assigned – created before the changes here. Can you remove all segments and test again?
I don't think there's a need to handle legacy segments since these changes are not yet released.

dkoo (approved):
Nice, I think you're right—deleting all pre-existing segments fixes the issue.

PR #305, opened 2020-11-05
Retrieve email, donation data from Mailchimp

### LINKED ISSUE #304: Get client data from Mailchimp
When a link from MC email is followed, let's assume the user is the one who subscribed and:
1. update their client data,
2. ask MC API for other client data, e.g. about donations

According to [MC docs](https://mailchimp.com/developer/guides/marketing-api-conventions/), the `mc_eid` parameter is "a unique Mailchimp ID that identifies the recipient's email address":

> Use the mc_eid value from the tracking URL to associate orders with a subscribed contact. You can make an API call to the /lists/{list_id}/members endpoint and specify the member with the unique_email_id query param by the unique_email_id value from the mc_eid in the link to find the email address for that recipient, which you can then use to identify the customer for your order.

adekbadek:
### All Submissions:

* [X] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

#### Questions/issues:

1. This is using the Mailchimp API key provided for Newspack Newsletters plugin (`newspack_newsletters_mailchimp_api_key` option) – while concise, this introduces coupling between the two. Other way would be to add a similar option in this plugin's settings. WDYT?
2. Is donation data stored in Mailchimp as orders, and thus retrievable as the subscriber's total revenue? Assuming yes and waiting for more data.

Closes #304 

1. Create four inline campaigns and four matching segments: donors (D) and not donors (ND), subscribers (S) and not subscribers (NS)
2. Observe that by default only the two NS, ND campaigns are displayed in a new session
3. Create a newsletter (using Newspack Newsletters™ & Mailchimp) with a link to a post on the site
4. Visit that link in a new session, observe that only S, ND campaigns are visible
5. Now I have not found a simple way to create a dummy Mailchimp order, so I've tested that part by hardcoding a revenue value [in `api/segmentation/class-segmentation-client-data.php`](https://github.com/Automattic/newspack-popups/blob/c5460527612c460928f7fd600368e42887822c4d/api/segmentation/class-segmentation-client-data.php#L78).
6. Now that the test reader has some donation data saved, only D, S campaigns should be visible – on subsequent visits, since the MC data fetching is done asynchronously

_Note: you might need this filter to bypass SSL check on a local environment:_
```
add_filter(
	'http_request_args',
	function ( $r ) {
		$r['sslverify'] = false;
		return $r;
	}
);
```

### Other information:

* [X] Have you added an explanation of what your changes do and why you'd like us to include them?


dkoo (changes requested):
I had trouble with these steps:

> 3. Create a newsletter (using Newspack Newsletters™) with a link to a post on the site
> 4. Visit that link in a new session, observe that only S, ND campaigns are visible

When I visited the link from the email in a new session, the ND campaign was visible, but I couldn't get _inline_ S campaigns to display. Overlay S campaigns DID display, however.

dkoo:
Also had trouble with this step:

> 6. Now that the test reader has some donation data saved, only D, S campaigns should be visible

After hardcoding `$revenue = 100` at the specified line, in a new session, ND campaigns were still displayed, and D campaigns were not (same for both overlay and inline). I verified that both this issue and the issue with S campaigns happen after running `composer install`, and tried both with and without the SSL filter you provided above.

api/segmentation/class-segmentation-client-data.php
```diff
@@ -50,6 +54,38 @@ public function report_client_data( $request ) {
 			$client_data_update['email_subscriptions'][] = $email_subscription;
 		}
 
+		// Fetch Mailchimp data.
+		$mailchimp_campaign_id   = $this->get_request_param( 'mc_cid', $request );
+		$mailchimp_subscriber_id = $this->get_request_param( 'mc_eid', $request );
+		if ( $mailchimp_campaign_id && $mailchimp_subscriber_id ) {
+			$mailchimp_api_key_option_name = 'newspack_newsletters_mailchimp_api_key';
+			global $wpdb;
+			$mailchimp_api_key = $wpdb->get_row( // phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
```
dkoo, line 63: @adekbadek Just wanted to ask, are we accessing the options table directly for performance reasons, as opposed to just using `get_option`?

dkoo (approved):
This seems to be working now, thanks for fixing.

api/segmentation/class-segmentation-client-data.php
adekbadek: This is because in the API part of this plugin, WordPress is not loaded. Just the bare minimum is, and some functions are mocked – see `api/setup.php`.

adekbadek:
Not merging yet, as I'm awaiting data about how is the donations data stored in Mailchimp.

adekbadek:
Updated to use MC's merge fields as source of truth about donations data in 4fe4af629b31fbd110c4a99000d0335dfe9875a9

dkoo (commented):
> Updated to use MC's merge fields as source of truth about donations data in 4fe4af6

Code looks good, any advice on how to test?

adekbadek:
> Code looks good, any advice on how to test?

You'd need a MC account with the merge fields configured, and a subscriber with appropriate values set there.

dkoo:
> You'd need a MC account with the merge fields configured, and a subscriber with appropriate values set there.

Here's what I did to try to test: I subscribed myself to my list in MC, then added a merge field called "DONATED" and set the value to `true` for my email address. Then sent a newsletter and visited the link from that email address in a new session. Should I expect to see the "Donated" segment only there? I'm still seeing campaigns with the non-donors segment only.

dkoo (approved):
I recreated the setup on jurassic.ninja and all is now working as expected, so let's chalk up the prior message to user/tester error.

Issue #193, opened 2020-08-12
AMP scripts polyfills for non-AMP pages


dkoo:
Currently, to avoid code splitting, [the AMP runtime and other AMP scripts are enqueued](https://github.com/Automattic/newspack-popups/blob/master/includes/class-newspack-popups-inserter.php#L547) by this plugin regardless of whether the page is being served as an AMP page. 

This solution has [caused](https://github.com/Automattic/newspack-popups/issues/151) [various](https://github.com/Automattic/newspack-popups/issues/153) [problems](https://github.com/Automattic/newspack-popups/issues/170) on the front-end.

To put these issues to rest and hopefully avoid any undiscovered or future issues, we should refactor to code split for non-AMP pages, rather than try to force AMP scripts on non-AMP pages.

---

Scripts:
- [X] amp-form
- [x] amp-analytics
- [ ] amp-access (https://github.com/Automattic/newspack-popups/pull/959)
- [x] amp-animation (https://github.com/Automattic/newspack-popups/pull/946)
- [x] amp-bind (https://github.com/Automattic/newspack-popups/pull/946)
- [x] amp-position-observer (https://github.com/Automattic/newspack-popups/pull/946)

aschweigert:
@dkoo it looks like this can be closed, right?

dkoo:
@aschweigert we can probably deprioritize this since I don't think any new issues have been reported as a result of the AMP scripts being loaded on non-AMP pages. #170 is still technically an open issue, but it's an edge case and there's an easy workaround (to disable Campaigns on non-AMP posts that use Jetpack slideshow blocks).

yogeshbeniwal:
These 4 additional JS files are adding to unused Javascript, leading to LCP in Web Vitals. IMHO we should prioritize removing these for overall page experience as @laurelfulford is doing for theme

dkoo:
Since we're moving toward code unification under AMP Plus/JS, we can probably close this as a #wontfix as it won't be relevant in the eventual future where we're not relying on AMP runtime scripts to control prompt display.

yogeshbeniwal:
@dkoo May I know if there is any tracker issue for new approach?

dkoo:
There isn't currently an open issue to track the non-AMP approach. It will likely be more of a milestone than a single issue when the work does happen.

adekbadek:
Just to add some clarity here: this issue is about moving away from AMP (a "JS approach"), but in a way that requires minimal changes. 

[...]

PR #306, opened 2020-11-09
Polyfill amp-analytics script

adekbadek:
### All Submissions:

Polyfill of `amp-analytics` AMP script (just the bits used by the plugin), part of #193.

A part that needs some explanation here is the linker parameter. Here's [a short overview](https://amp.dev/documentation/components/amp-analytics/#linkers), and here the [documentation about linker param receiving](https://github.com/ampproject/amphtml/blob/master/extensions/amp-analytics/linker-id-receiving.md) – the part that's actually polyfilled. Linker param _setting_ is not handled here, because it would be set only by a page served from AMP cache, which would be… an AMP page (not a version with polyfill).

Here's an example of a linker param:
```
?ref_newspack_cid=1*10zr7ce*cid*T2ptZDdJNU9EYndELWV0TlM1N0FwSGVwNHE3S240VkVVU0o3YlNsaFc2YVVPRXJhSWFhaVlsOTgtQXNsc21Eeg
```

The `T2pt…` part is base64-encoded client ID (of value `Ojmd7I5ODbwD-etNS57ApHep4q7Kn4VEUSJ7bSlhW6aUOEraIaaiYl98-AslsmDz`). You can this set the client ID to a specified value (in JS, e.g. `btoa('hello-world')` gives you a base64 encoded `hello-world`).

Note that the checksum part of the linker param is disregarded in this polyfill. I think the security implications of client ID spoofing in this use case are negligible.

1. Linker parameter
    1. Append a linker param to a post URL in non-AMP mode, visit the site in a fresh session (with no `newspack-cid` cookie set)
    2. Observe in the devtools that the `newspack-cid` cookie is set accordingly
    3. Observe that the linker param is stripped from the URL afterwards
    4. Confirm that this works in the same way on an AMP site – you can use [this AMP-cached site](https://campaigns--testing-newspackstaging-com.cdn.ampproject.org/c/s/campaigns-testing.newspackstaging.com/) to "generate" linker params with valid checksums
2. Reporting views to API 
    1. Create a campaign with "once" frequency
    2. Visit the site in a fresh session in non-AMP mode, observe the campaign is displayed (make sure you "see" it – that means at least 500ms with 90% of the campaign visible in the viewport)
    3. Reload the site or navigate to another post – observe the campaign is not displayed

### Other information:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,6 +70,15 @@ jobs:
             composer install
             ./vendor/bin/phpcs
 
+  test-js:
+    docker:
+      - image: circleci/node:12
+    steps:
+      - checkout_code
+      - run:
+          name: Run JS Tests
+          command: npm run test
+
   release:
     docker:
       - image: circleci/node:12
@@ -104,6 +113,9 @@ workflows:
       - build
       - lint-php
       - lint-js-scss:
+          requires:
+            - build
+      - test-js:
           requires:
             - build
       - release:
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,0 +1,10 @@
+module.exports = ( { env } ) => {
+	const isTest = env( 'test' );
+	return {
+		presets: [
+			...( isTest ? [ [ '@babel/preset-env', { targets: { node: 'current' } } ] ] : [] ),
+			'@automattic/calypso-build/babel/default',
+			'@automattic/calypso-build/babel/wordpress-element',
+		],
+	};
+};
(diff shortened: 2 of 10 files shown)


dkoo (approved):
This works as described and the polyfill is quite impressive. The new JS is pretty complex—do you think it's worth adding some JSDoc for the new functions? Otherwise, looks good to me and great job.

Tested in Chrome, Safari and Firefox and the linker functionality works as expected across non-AMP to AMP and vice versa.

adekbadek:
> The new JS is pretty complex—do you think it's worth adding some JSDoc for the new functions?

Good idea, added in effcfdc34a5a3ab99c45526dd7d695fe4369f597

Also, I've added handling of `ini-load` event to account for changes introduced in #305 in 6c43ab39a25bd81d6308d88d42c8fda4e2c0dc49

dkoo (approved):
:shipit:

PR #307, opened 2020-11-10
feat: reorganize editor sidebar and add notice when test mode is enabled

### LINKED ISSUE #97: Reorganise editor sidebar and add notice when Test mode is enabled
I suggest we differentiate the Inline pop-ups from the "Overlay" pop-ups using a toggle.

Whether it's enabled or not the settings will be slightly different.

### Overlay pop-up
overlay-popup-editor

### Inline pop-up
inline-popup-editor

### Details

I moved the test mode to its own control, and part of the "Status & visibility" section. And did the same for the Overlay's Sitewide default (we don't need this option for the Inline pop-ups anymore).

If test mode is enabled, I added a dismissible notice to the editor to avoid confusion.

I moved the "Inline" placement to its own control so we can toggle between the 2 different types of Pop-ups: "Inline" or "Overlay".

I changed the Trigger to a select just for visual consistency with the Placement setting above. 

Frequency and Color have now their own sections.

dkoo:
### All Submissions:

Reorganizes the editor sidebar according to design mockups by @thomasguillot 

Closes #97.

@thomasguillot, I extrapolated from your mockups in a few ways:

1. I replaced all instances of the word "popup" with "campaign" to follow the updated branding.
2. When showing the "test mode" warning/notice, I couldn't figure out how to format the text as designed. The component only takes plain, unformatted strings as content, so this is the closest I could get:

Screen Shot 2020-11-10 at 12 34 18 PM

3. I updated some help text for certain options, and added some where I felt it was missing. For example:

Screen Shot 2020-11-10 at 12 24 39 PM

4. I moved the new "Segment" option to the Frequency panel, since it felt like the most appropriate place. Let me know if you disagree:

Screen Shot 2020-11-10 at 12 48 41 PM

5. This is no longer applicable, but I'm keeping the examples here for reference. ~Rather than simply showing or hiding options based on whether the current campaign is inline vs. overlay, we now show disabled controls when an option is not applicable. I felt this was less confusing since the options are now split between various panels which may or may not be open at any one time. For example, when the "inline" option is enabled in the Campaign Settings panel, we show this for the "sitewide default" option in the Status & Visibility panel:~

Screen Shot 2020-11-10 at 12 41 37 PM

~and this when the "inline" option is disabled:~

Screen Shot 2020-11-10 at 12 41 42 PM

~Likewise for the color options, if the campaign is inline, we show disabled controls for the non-applicable options:~

Screen Shot 2020-11-10 at 12 41 55 PM

~For the placement/trigger settings, we still show and hide those based on the inline setting, since they're right underneath that setting (but we could do the disabled thing here, too, for consistency):~

Screen Shot 2020-11-10 at 12 21 56 PM
Screen Shot 2020-11-10 at 12 24 39 PM

~Let me know if you disagree with this approach, and I'll be happy to refactor as you prefer.~

1. Check out this branch and run `npm run build` to compile.
2. Edit a campaign or create a new one.
3. Verify that the layout of all sidebar panels and options are as described in #97.
4. Verify that the options are shown/hidden according to whether the "inline" option is enabled or disabled.
5. Verify that the options still function as before—there should be no changes in behavior on the front-end.

### Other information:


thomasguillot:
> Let me know if you disagree with this approach, and I'll be happy to refactor as you prefer.

I understand the idea but I disagree 😊
In a perfect world (e.g. with the new NP onboarding and enabling some "smart defaults"), overlay campaigns are going to be (almost?) non-existent as this is not good practice. Having a bunch of disabled options for something we don't really recommend feels unneeded.

thomasguillot:
> I moved the new "Segment" option to the Frequency panel, since it felt like the most appropriate place. Let me know if you disagree

Smart move.
I  do wonder if it should have its own panel though?
Maybe with a suggestion to create a segment if there aren't any?
What do you think?

dkoo:
> In a perfect world (e.g. with the new NP onboarding and enabling some "smart defaults"), overlay campaigns are going to be (almost?) non-existent as this is not good practice. Having a bunch of disabled options for something we don't really recommend feels unneeded.

Sounds good! I went back to hiding the controls when the campaign is inline in 463b272.

> I do wonder if it should have its own panel though?
> Maybe with a suggestion to create a segment if there aren't any?
> What do you think?

Good ideas! Also updated in 463b272.

thomasguillot:
This is great, thank you @dkoo!

Screenshot 2020-11-12 at 16 10 50
Screenshot 2020-11-12 at 16 11 20

Can we just drop the "Campaign" from the Settings? (except for Campaign Settings 😊). It feels very redundant.

dkoo:
> Can we just drop the "Campaign" from the Settings? (except for Campaign Settings 😊). It feels very redundant.

Done in 65c713b!

thomasguillot:
> Done in 65c713b!

Amazing. Thank you so much @dkoo

thomasguillot (approved):

adekbadek (commented):
Everything works as expected – I've left one suggestion about test mode notice.

Two nitpicks:
1. All these components are prefixed with `Popup`, but since that's the main entity this plugin deals with, I don't think it's needed
2. The choice to use class components instead of function ones is a bit odd – using the latter would be more readable, and we're generally moving away from class components

src/editor/PopupStatusSidebar.js
```diff
@@ -0,0 +1,74 @@
...
+					<ToggleControl
+						checked={ isTest }
+						label={ __( 'Test Mode', 'newspack-popups' ) }
+						help={ __(
+							'In "Test Mode" logged-in admins will see the campaign every time, and non-admins will never see them.',
+							'newspack-popups'
+						) }
+						onChange={ value => {
+							if ( value ) {
+								createNotice(
```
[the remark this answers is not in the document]
dkoo: Done in dec2b0e.

dkoo:
> 1. All these components are prefixed with Popup, but since that's the main entity this plugin deals with, I don't think it's needed

Removed the `Popup` prefixes in dec2b0e.

> 2. The choice to use class components instead of function ones is a bit odd – using the latter would be more readable, and we're generally moving away from class components

I was lazy and copy/pasted the existing `PopupSidebar` component. I refactored to functional components in dec2b0e.

src/editor/StatusSidebar.js
```diff
@@ -0,0 +1,78 @@
...
+				<ToggleControl
+					checked={ isTest }
+					label={ __( 'Test Mode', 'newspack-popups' ) }
+					help={ __(
+						'In "Test Mode" logged-in admins will see the campaign every time, and non-admins will never see them.',
+						'newspack-popups'
+					) }
+					onChange={ value => {
+						if ( value ) {
+							createTestNotice();
```
adekbadek, line 58: Nitpick: That's already handled in the `useEffect` hook, where the removal of the notice should also take place.

adekbadek (approved):
Works as it should; left one comment.

src/editor/StatusSidebar.js
dkoo: Good call—updated in 8fdc7d7.

adekbadek (approved):
:shipit:

PR #310, opened 2020-11-16
feat: add option for dismiss button alignment

### LINKED ISSUE #41: Add align option to the "Not interested" button
Currently the "Not interested" text is centred which can look out of place when the content is left aligned.

I suggest we add an option to change the text alignment { left / center / right }

dkoo:
### All Submissions:

Adds a new option/attribute for setting the alignment of the "I'm not interested" dismiss button. Looks like this:

Screen Shot 2020-11-16 at 9 54 06 AM

Default is center, so it shouldn't affect existing campaigns.

Depends on #307 because it was easier to slot this change into the reorganized sidebar options—I can rebase this PR to master if that one gets merged first.

Note that changing this setting has no visible effect in the editor, because we don't currently show any preview of the dismiss button in the block editor (though we could potentially add one if you feel it's warranted, @thomasguillot).

Closes #41.

1. Check out this branch and run `npm run build` to compile assets.
2. View an existing inline or overlay campaign on the front-end. Confirm that there's no difference in display compared to master.
3. Edit the existing campaign. Set the "Dismiss button alignment" option to left, center, and right and confirm that the button appears aligned as specified on the front-end.
4. Repeat with a new campaign.

### Other information:


thomasguillot:
> Note that changing this setting has no visible effect in the editor, because we don't currently show any preview of the dismiss button in the block editor (though we could potentially add one if you feel it's warranted, @thomasguillot).

How would this work? Can we have a non-removable block at the bottom? That would be great because we could remove all the "Dismiss Button" options: Use the inline-editor for the text, and use the native text alignment (like the Paragraph block) for the... text alignment.

thomasguillot:
Kapture 2020-11-17 at 14 46 04

The alignment doesn't seem to have any effect on the front-end?

dkoo:
> I think it's probably better if we switch to a dropdown.

Updated in fbc57ed.

dkoo:
> The alignment doesn't seem to have any effect on the front-end?

Ah, I missed one place where this would affect overlay campaigns. Fixed in 6c7e918. Thanks for catching!

dkoo:
> How would this work? Can we have a non-removable block at the bottom? That would be great because we could remove all the "Dismiss Button" options: Use the inline-editor for the text, and use the native text alignment (like the Paragraph block) for the... text alignment.

I wasn't thinking it would be a block, but that's a neat idea. My idea was much simpler, basically just rendering a non-interactive static preview of what it would look like at the bottom of the block content. Either way, I'd suggest we discuss and tackle in a new issue/PR since it sounds like a tangent from these changes? I opened #311 to continue the discussion.

thomasguillot (commented):
This is looking great, thank you @dkoo.

What do you think if we move the Dismiss Button settings to its own section and shorten the labels?

dismiss-settings

dkoo:
> What do you think if we move the Dismiss Button settings to its own section and shorten the labels?

Sounds good—updated in 5a41b31.

thomasguillot (approved):
Looking good!

PR #315, opened 2020-11-23
Improve configuration file handling

### LINKED ISSUE #314: Configuration: Not everyone has access to root
The [rewritten API](https://github.com/Automattic/newspack-popups/pull/208) needs a configuration file to sit in the root of the site to function. While the plugin attempts to create this, write access is often not allowed anywhere other than in the `wp-content` folder which means that many sites (e.g. those on shared/managed hosting) cannot use this newer method.

Generally, `wp-content` is accessible and writable and would be a better place to look for/place the configuration file.

philipjohn:
### All Submissions:

This PR does three things:

1. Closes #314 by moving the configuration file location to the `wp-content` directory rather than root
1. Fixes the conditional for checking whether to create a config file or not, as it wasn't actually working reliably on non-Atomic environments
1. Changes the hook used so that the plugin is not attempting to write to the disk from frontend requests, which could create performance issues

1. On a non-Atomic environment where no configuration file exists, observe there is no config file
2. Open any admin screen
3. Observe there is now a configuration file

### Other information:

### DIFF
--- a/api/setup.php
+++ b/api/setup.php
@@ -22,7 +22,7 @@ if ( ! defined( 'WP_DEBUG' ) ) {
 	define( 'WP_DEBUG', false );
 }
 
-require_once $wp_root_path . 'newspack-popups-config.php';
+require_once $wp_root_path . 'wp-content/newspack-popups-config.php';
 
 // phpcs:disable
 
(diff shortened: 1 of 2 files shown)


adekbadek (approved):
Fixed one thing in 8416fa8 & tested on an ephemeral site – works well 👍

PR #321, opened 2020-11-25
Shortcode (officially)

adekbadek:
### All Submissions:

Adds ability to place a campaign anywhere, using a shortcode.
And a way to restrict a campaign to shortcode-only (manual).

1. Create an inline campaign, with "Every page" frequency.
2. Publish the campaign, observe that it's displayed on the frontend as expected
3. Add a shortcode – via shortcode block in a page, in a post, and in a text widget outside of the content area
4. Observe that the campaign is shown in all the shortcoded places
5. Observe that the campaign is not duplicated in the post content
6. In a different post (without the shortcode), observe the campaign is inserted automatically

### Caveats

If a campaign is placed via shortcode in a text widget, it will be treated as visible regardless of whether the widget area is actually rendered. This means that preventing duplication of campaigns in a widget area and in content is not feasible.
If there's a way to determine whether widget areas are rendered (regardless of the theme), this could be improved.

### Other information:


includes/class-newspack-popups-inserter.php
```diff
@@ -127,7 +127,7 @@ public function __construct() {
 		add_filter( 'the_content', [ $this, 'insert_popups_in_content' ], 1 );
 		add_shortcode( 'newspack-popup', [ $this, 'popup_shortcode' ] );
 		add_action( 'after_header', [ $this, 'insert_popups_after_header' ] ); // This is a Newspack theme hook. When used with other themes, popups won't be inserted on archive pages.
-		add_action( 'wp_head', [ $this, 'insert_popups_amp_access' ] );
+		add_action( 'wp_footer', [ $this, 'insert_popups_amp_access' ] );
```
jeffersonrabb, line 130: In my testing the API request doesn't fire at all, unless I revert to the `wp_head` action. What's the reason for this change?

includes/class-newspack-popups.php
```diff
@@ -149,6 +149,18 @@ public static function register_meta() {
 			]
 		);
 
+		\register_meta(
```
jeffersonrabb, line 152: What's the reason for the `Manual Only` option? Couldn't a manual shortcode placement always take precedence over normal campaign logic, and avoid burdening the user with a difficult-to-understand new setting?

jeffersonrabb (changes requested):

jeffersonrabb (changes requested):
I'm having trouble testing this one. 

- Created an inline campaign, verified it appears normally in content. 
- Then I add it as a shortcode on the homepage, using Shortcode block with this content: `[newspack-popup id="8375"]`
- The markup is present on the home, but the campaign doesn't show. 
- Examining the API  request from the homepage there's nothing in popups

includes/class-newspack-popups-inserter.php
adekbadek: _Accidentally removed the comments by cancelling a review._

I've reverted this change, probably forgot to remove earlier. Works both ways for me, though.

includes/class-newspack-popups.php
adekbadek: _Accidentally removed the comments by cancelling a review._

> Couldn't a manual shortcode placement always take precedence over normal campaign logic

That's easy for shortcodes in post content, but I couldn't make it work for shortcodes in widgets. Looks like widgets are rendered after post content, so the shortcodes rendered in the post content cannot be made aware of their existence.

includes/class-newspack-popups-inserter.php
adekbadek: Ah no, it's intentional. The change to insert amp-access in the footer hook was to make amp-access aware of any inserted shortcodes: 

https://github.com/Automattic/newspack-popups/blob/8f024171fb4c95d3347397407ab188f16feb372a/includes/class-newspack-popups-inserter.php#L373-L379

This does not cause issues for me with API requests though, I'll investigate further.

adekbadek:
Turns out it's because of AMP validation – I was testing on non-AMP, so did not get the error. The issue is that `amp-access` has to be placed in document head – an AMP page will strip an `amp-access` script from the footer. 
The problem with it is that `insert_popups_amp_access`, which is ran in the `wp_head` has to be aware of any shortcode-inserted campaigns in the page/widgets content.

jeffersonrabb (approved):
This is working well and ready to go. A few ideas for future improvements, which I'll also capture in enhancement issues:

1) The `newspack_popups_widget_shortcode_popups_ids` option will contain IDs of all Text Widgets, whether or not they contain a campaign shortcode. This isn't damaging, it just means that the option may be larger than necessary. A site would have to have a huge number of Text Widgets for this to cause any real issues!

2) It might make sense to leverage core WordPress shortcode functions for detection and extraction, e.g. [get_shortcode_regex](https://developer.wordpress.org/reference/functions/get_shortcode_regex/). In my testing the regexes in the PR all work well, but using core functions would add a bit of confidence that we're accounting for all possible formatting quirks.

3) If campaign shortcodes turns out to be a common pattern it might make sense to create a campaign widget to make the flow easier.

PR #336, opened 2020-12-08
feat: render a preview of the dismiss button in the editor

### LINKED ISSUE #311: Show a preview of the dismiss button in the editor?
Two ideas to explore:

1. Create a custom block to edit the dismiss button text/styles, that always appears at the end of all campaign posts and is undeleteable. Remove the options from the sidebar and use the block attributes instead.

2. Render a simple non-interactive preview of the button as static HTML toward the bottom of the editor content.

Off the top of my head, I don't believe (1) will work because the dismiss button settings are currently stored and rendered as post metadata.

dkoo:
### All Submissions:

Uses `editor.BlockListBlock` filter to render a preview of the dismiss button text and alignment at the end of the editor block list.

Closes #311.

1. Check out this branch and run `npm ci && npm run build` (must install new NPM packages before building).
2. Create a new campaign or edit an existing one.
3. Observe that a non-interactive preview of the dismiss button now appears after the last block in the campaign post content. ~**Note:** If testing with a new campaign, the preview will not appear until the content contains at least one block, e.g. after clicking into the block editor area.~

Screen Shot 2020-12-07 at 6 55 30 PM

4. Test that when changing the dismiss button text and alignment, the preview updates itself accordingly.
5. Test that the preview only renders after the last block, even after adding and removing blocks. ~**Note:** On initial page render, the block appender button may appear underneath the preview, but the preview will move itself if more blocks are added to the content.~
6. Test that the preview does not render with any post type other than Campaigns.

### Other information:


adekbadek (approved):
🚢

thomasguillot (changes requested):
> If testing with a new campaign, the preview will not appear until the content contains at least one block, e.g. after clicking into the block editor area.

This is just very odd and I didn't understand at first ([I even did a screencast](https://user-images.githubusercontent.com/177929/101481062-7b11e480-394c-11eb-8bdc-ef51400357df.gif), then I saw your note). I think the button should always be present, even when there aren't any blocks, especially because I can preview an empty Campaign and the "Not interested" button [will be there](https://user-images.githubusercontent.com/177929/101481226-bdd3bc80-394c-11eb-8327-e4b117820481.png).

> On initial page render, the block appender button may appear underneath the preview, but the preview will move itself if more blocks are added to the content.

This is also very odd. Like above, I started taking screenshots of this before reading your note. The button has to be at the bottom.

Then this is probably theme-related (cc @laurelfulford) but the font in the editor doesn't match the preview.

Screenshot 2020-12-08 at 11 47 00
Screenshot 2020-12-08 at 11 47 15

dkoo:
@thomasguillot I completely understand your reaction to the two issues I mentioned (the preview not appearing until after a block exists, and the appender appearing beneath the button at first). These are both side effects of the hook we were using to render the preview, `edit.BlockListBlock`. Currently this is the only existing hook that lets us filter the block editor area to render arbitrary components like our preview, and it is only fired when a block list exists in the editor (which only happens when there's at least one block). It also gets fired separately from the appender.

So all this to say, I was trying to achieve this in a React/WP-friendly way, but I don't think it's possible yet with the limited filters the editor currently provides. So I've refactored this in a simpler way that addresses your two UX points (but not the mismatched typeface—that would need to be addressed in the theme). What do you think now?

@adekbadek I'll need another review since 2258275 changes the technical approach.

adekbadek (changes requested):
Works well, but only after the "Dismiss Button Settings" panel is opened. If the editor is loaded with the panel closed, the dismiss button preview will not be rendered. 
This can be fixed by firing the effect hook in the `render` function of `'newspack-popups-dismiss'` editor plugin.

thomasguillot (commented):
> @thomasguillot I completely understand your reaction to the two issues I mentioned (the preview not appearing until after a block exists, and the appender appearing beneath the button at first). These are both side effects of the hook we were using to render the preview, edit.BlockListBlock. Currently this is the only existing hook that lets us filter the block editor area to render arbitrary components like our preview, and it is only fired when a block list exists in the editor (which only happens when there's at least one block). It also gets fired separately from the appender.
> 
> So all this to say, I was trying to achieve this in a React/WP-friendly way, but I don't think it's possible yet with the limited filters the editor currently provides. So I've refactored this in a simpler way that addresses your two UX points (but not the mismatched typeface—that would need to be addressed in the theme). What do you think now?

This is great, thank you for implementing this! I know it's not ideal from a React point of view but for the user this is so much better 😊

And yes, like @adekbadek said, it needs to be loaded even when the sidebar is closed.

dkoo:
> If the editor is loaded with the panel closed, the dismiss button preview will not be rendered.

Ah, good catch. I think this is fixed in 8c6c2bb, can you confirm?

src/editor/index.js
```diff
@@ -110,14 +111,49 @@ registerPlugin( 'newspack-popups-segmentation', {
...
 registerPlugin( 'newspack-popups-dismiss', {
-	render: () => (
-		<PluginDocumentSettingPanel
-			name="popup-dismiss-panel"
-			title={ __( 'Dismiss Button Settings', 'newspack-popups' ) }
-		>
-			<DismissSidebarWithData />
-		</PluginDocumentSettingPanel>
-	),
+	render: () => {
```
adekbadek, line 114: As the [linter says](https://app.circleci.com/pipelines/github/Automattic/newspack-popups/717/workflows/82089462-e673-4724-8c7b-f174a262bb70/jobs/2123/parallel-runs/0/steps/0-103), this anonymous function should be a React component.

package-lock.json
```diff
@@ -11362,11 +11362,26 @@
...
+          }
+        },
+        "regenerator-runtime": {
+          "version": "0.13.7",
+          "resolved": "https://registry.npmjs.org/regenerator-runtime/-/regenerator-runtime-0.13.7.tgz",
+          "integrity": "sha512-a54FxoJDIr27pgf7IgeQGxmqUNYrcV338lf/6gH456HZ/PhX+5BcwHXG9ajESmwe6WRO0tAzRUrRmNONWgkrew=="
+        }
       }
     },
     "@wordpress/html-entities": {
```
adekbadek, line 11387: nitpick: since there are no changes to `package.json`, these changes are not needed here

adekbadek (changes requested):
> Ah, good catch. I think this is fixed in 8c6c2bb, can you confirm?

Confirmed 👍

package-lock.json
dkoo: Reverted those changes in d2d7a0e

src/editor/index.js
dkoo: Is this overengineered? I made it a higher-order component in d2d7a0e.

adekbadek (approved):

src/editor/index.js
adekbadek: A little bit – since it's a single-use HOC :D But it's all right

dkoo:
@thomasguillot Can you please take another look?

thomasguillot (approved):
> @thomasguillot Can you please take another look?

Yes, sorry. I thought I approved it earlier today... clearly forgot to press "submit review" 🤦 

🚢

PR #345, opened 2020-12-11
Add "view as segment" feature

adekbadek:
### All Submissions:

Adds a feature that enables admins to view site as a member of a particular segment would see it.
There's not UI for that yet – it's a 'stealth' feature. The way to control it it via a URL query param and cookies.

1. Create a bunch of campaigns, using different segments – to use all segmentation options.
2. Visit the site as a logged-in user. 
3. Observe that active campaigns are not visible, and test campaigns are – nothing new here
4. Append a `view_as` query param to the URL, with the value of `segment:<segment-id>` (e.g. `?view_as=segment:5fa056b4b94bc`). You can read the segment ID from URL in Campaigns Wizard -> Segments -> edit a segment.
5. Observe that now, as a logged in admin user, you can see the applicable active campaigns – as a member of the chosen segment would.
6. Navigate to another page ~(now the URL param will be gone) – observe the above is still the case~
7. ~Remove the `newspack_popups_view_as` cookie (Chrome devtools -> Application panel)~
8. Observe that now only test campaigns are visible on the site

Edit: 👆 removed cookie use

### Other information:

### DIFF
--- a/api/campaigns/class-campaign-data-utils.php
+++ b/api/campaigns/class-campaign-data-utils.php
@@ -39,18 +39,41 @@ class Campaign_Data_Utils {
 	}
 
 	/**
+	 * Add default values to a segment.
+	 *
+	 * @param object $segment Segment configuration.
+	 * @return object Segment configuration with default values.
+	 */
+	public static function canonize_segment( $segment ) {
+		return (object) array_merge(
+			[
+				'min_posts'         => 0,
+				'max_posts'         => 0,
+				'is_subscribed'     => false,
+				'is_not_subscribed' => false,
+				'is_donor'          => false,
+				'is_not_donor'      => false,
+				'referrers'         => '',
+			],
+			(array) $segment
+		);
+	}
+
+	/**
 	 * Given a segment and client data, decide if the campaign should be shown.
 	 *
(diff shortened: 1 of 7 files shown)


dkoo (changes requested):
This is great! Works as described while logged in and does nothing while not logged in (as expected). Has great potential for a nice new testing UI which can probably completely replace the current "test mode."

I did see the following warning when trying to set the `view_as` param on a non-AMP page:

```
PHP Warning:  Cannot modify header information - headers already sent by (output started at /Users/dkoo/Local Sites/newspack/app/public/wp-includes/general-template.php:3071) in /Users/dkoo/Local Sites/newspack/app/public/wp-content/plugins/newspack-popups/includes/class-newspack-popups-view-as.php on line 64
```

This doesn't seem to break the feature, though—I still see the campaigns matching the segment.

adekbadek:
@dkoo I've removed the cookie use, there should be no warning now.

dkoo (approved):
Confirmed that I'm no longer seeing the PHP warning. I think it makes sense to not use the cookie and just rely on the query param. We could pass the param from page to page when we implement a testing UI.

PR #376, opened 2021-01-08
fix: remove trailing question mark from cleaned non-AMP URLs

dkoo:
### All Submissions:

Due to [this clean function](https://github.com/Automattic/newspack-popups/pull/306/files#diff-961feb79bdd31646421b94dfdb35d9ea989a94b447987ea75e325e9665465241R86), when visiting a non-AMP page or post, the AMP polyfill cleans the URL of linker params but leaves a trailing `?` character at the end of the URL. It also creates extra browser history via `history.pushState`. This PR removes the trailing `?` and uses `history.replaceState` to avoid the extra browser history.

1. On `master`, visit a non-AMP post in a new Incognito window and observe that the page loads with a trailing `?` on the URL. Also observe that clicking the browser's back button after the the page loads just removes the `?` without actually going back to the prior page.
2. Check out this branch and run `npm run build`.
3. Visit the same post and observe that the trailing `?` no longer gets appended.
4. Open Dev Tools > Application > Cookies > [your test site URL] and confirm that the AMP cookie is still getting set correctly as `newspack-cid`.
5. Confirm that hitting the back button returns to the prior page as expected.

### Other information:

### DIFF
--- a/src/view/index.js
+++ b/src/view/index.js
@@ -47,7 +47,7 @@ const manageAnalyticsLinkers = () => {
 			document.cookie = cookieValue;
 		}
 		if ( cleanURL ) {
-			window.history.pushState( {}, document.title, cleanURL );
+			window.history.replaceState( {}, document.title, cleanURL );
 		}
 	} );
 };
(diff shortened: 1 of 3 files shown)


claudiulodro (approved):
Works great, code looks good! Thanks for the speedy fix! 🎖️

PR #356, opened 2020-12-18
fix: show pending and future campaigns in UI

dkoo:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Current campaign UI only shows campaign posts that have `publish` and `draft` statuses. This would cause posts with the `pending` or `future` status to be dropped before being rendered in the UI.

This PR, along with https://github.com/Automattic/newspack-plugin/pull/757, shows pending drafts under the "Draft" header and scheduled campaigns under the "Inactive" header, with a corresponding "Pending review" or "Scheduled" label for each.

Screen Shot 2020-12-18 at 3 36 46 PM

Closes #355.

### How to test the changes in this Pull Request:

Follow instructions in https://github.com/Automattic/newspack-plugin/pull/757.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #399, opened 2021-01-14
fix: in preview tab, allow previewing without choosing groups

dkoo:
### All Submissions:

Previews created from the preview tab does should be able to show any campaigns that match the selected segment, even if no groups are selected.

This PR can replace #356 as it contains all changes in that branch (required for properly testing that the Preview tab works with both published and unpublished campaigns). Leaving #356 open for now so that it can be tested separately if necessary.

1. On `master`, go to the Preview tab and select any segment while leaving the Groups field empty.
2. Observe that only published campaigns are displayed in the preview modal (basically only what can appear on the front-end rather than what matches the `view_as` param).
3. Check out this branch.
4. Select any segment while leaving Groups empty and "Show unpublished campaigns" OFF. Confirm that all published campaigns matching the selected segment, regardless of group, are shown in the preview modal.
5. Toggle "Show unpublished campaigns" ON, and confirm that all published and draft campaigns matching the selected segment, regardless of group, are shown in the preview modal.
6. Go back to the Campaigns tab and filter by All Campaigns (no group selected). Repeat steps 4-5 using the same segment you tested with in the Preview tab, and confirm that the campaigns you see are the same when previewing from both tabs.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -34,7 +34,7 @@ final class Newspack_Popups_Model {
 	public static function retrieve_popups( $include_unpublished = false ) {
 		$args = [
 			'post_type'      => Newspack_Popups::NEWSPACK_POPUPS_CPT,
-			'post_status'    => $include_unpublished ? [ 'publish', 'draft' ] : 'publish',
+			'post_status'    => $include_unpublished ? [ 'draft', 'pending', 'future', 'publish' ] : 'publish',
 			'posts_per_page' => 100,
 		];
 
@@ -211,12 +211,13 @@ final class Newspack_Popups_Model {
 	/**
 	 * Retrieve all inline popups.
 	 *
+	 * @param  boolean $include_unpublished Whether to include unpublished posts.
 	 * @return array Inline popup objects.
 	 */
-	public static function retrieve_inline_popups() {
+	public static function retrieve_inline_popups( $include_unpublished = false ) {
 		$args = [
 			'post_type'    => Newspack_Popups::NEWSPACK_POPUPS_CPT,
-			'post_status'  => 'publish',
+			'post_status'  => $include_unpublished ? [ 'draft', 'pending', 'future', 'publish' ] : 'publish',
 			'meta_key'     => 'placement',
 			// phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_value
 			'meta_value'   => self::$inline_placements,
@@ -269,12 +270,13 @@ final class Newspack_Popups_Model {
 	/**
 	 * Get overlay test popups.
 	 *
+	 * @param  boolean $include_unpublished Whether to include unpublished posts.
 	 * @return array Overlay test popup objects.
 	 */
-	public static function retrieve_overlay_test_popups() {
+	public static function retrieve_overlay_test_popups( $include_unpublished = false ) {
 		$args = [
 			'post_type'   => Newspack_Popups::NEWSPACK_POPUPS_CPT,
-			'post_status' => 'publish',
+			'post_status' => $include_unpublished ? [ 'draft', 'pending', 'future', 'publish' ] : 'publish',
 			'meta_query'  => [ // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
 				[
 					'key'     => 'placement',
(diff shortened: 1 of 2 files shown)


jeffersonrabb (approved):
🚢

adekbadek (approved):
This works as described, but I'm worried about the added complexity here – unless I'm missing something.

In both "Campaigns" and "Preview" tabs of the wizard, the user can trigger a preview. The preview content is conditioned by three settings: segment, groups, and whether to show unpublished campaigns. 
In these two instances, the code is a bit different though – in the "Campaigns" tab the preview is informed by a list of campaign IDs, while in "Preview" – by the chosen group. 

The `campaigns` part of the "view as" spec seems a bit in conflict with the `groups` part – the campaign IDs seem to always be from the same group, the way the wizard is set up.

dkoo:
> This works as described, but I'm worried about the added complexity here – unless I'm missing something.

> In both "Campaigns" and "Preview" tabs of the wizard, the user can trigger a preview. The preview content is conditioned by three settings: segment, groups, and whether to show unpublished campaigns.
In these two instances, the code is a bit different though – in the "Campaigns" tab the preview is informed by a list of campaign IDs, while in "Preview" – by the chosen group.

> The campaigns part of the "view as" spec seems a bit in conflict with the groups part – the campaign IDs seem to always be from the same group, the way the wizard is set up.

@jeffersonrabb Good point. Trying something new in c96454c, in response to that valid concern:

* We remove the functionality to preview by specific campaigns. Now, previewing all campaigns in the Campaigns tab is the same as previewing without specifying any groups in the Preview tab, and previewing a group in the Campaigns tab is the same as previewing with the specified group in the Preview tab.
* The above requires a small refactor in the plugin, which I've pushed as a draft [here](https://github.com/Automattic/newspack-plugin/pull/820).

While testing previewing all campaigns, I once again ran into the old question about what to do if multiple overlay campaigns match the selected preview params. Currently, we just show the first one and discard the rest, but this again seems confusing to me if I'm previewing a group that has two overlay campaigns, or previewing all campaigns and there's more than one overlay matching the selected segment. For example in my local site, I have two overlay campaigns, one published as a sitewide default and the other a draft that was created more recently. When previewing all campaigns with the "show unpublished" option enabled, the query gets both campaigns sorted by newest first. Then the inserter logic grabs just the first overlay (the more recent draft) and discards the sitewide default. If the draft overlay does not match the segment selected in the preview UI, no overlay is shown in the preview—even if the published sitewide default does match the segment!

On the front-end, we still show fetch a matching category overlay, or the Sitewide Default if none exists, so only one overlay will ever display. But while previewing, I would expect that any campaigns matching my preview options should show in the preview modal.

We have an open issue to implement a system of segment priority, which might solve this. But until we have that, I propose we bring back functionality to show all overlays in the preview. I've done this in c96454c, but can easily revert if you disagree.

includes/class-newspack-popups-model.php
```diff
@@ -211,12 +211,13 @@ public static function set_popup_options( $id, $options ) {
 	/**
 	 * Retrieve all inline popups.
 	 *
+	 * @param  boolean $include_unpublished Whether to include unpublished posts.
 	 * @return array Inline popup objects.
 	 */
-	public static function retrieve_inline_popups() {
+	public static function retrieve_inline_popups( $include_unpublished = false ) {
```
adekbadek, line 217: This method is no longer used with the argument

includes/class-newspack-popups-model.php
```diff
@@ -249,32 +250,16 @@ public static function retrieve_group_popups( $group_slugs, $include_unpublished
...
-	}
-
 	/**
 	 * Get overlay test popups.
 	 *
+	 * @param  boolean $include_unpublished Whether to include unpublished posts.
 	 * @return array Overlay test popup objects.
 	 */
-	public static function retrieve_overlay_test_popups() {
+	public static function retrieve_overlay_test_popups( $include_unpublished = false ) {
```
adekbadek, line 259: This method is no longer used with the argument

includes/class-newspack-popups-model.php
```diff
@@ -294,9 +279,10 @@ public static function retrieve_overlay_test_popups() {
 	/**
 	 * Retrieve first overlay popup matching post categries.
 	 *
+	 * @param  boolean $include_unpublished Whether to include unpublished posts.
 	 * @return object|null Popup object.
 	 */
-	public static function retrieve_category_overlay_popup() {
+	public static function retrieve_category_overlay_popup( $include_unpublished = false ) {
```
adekbadek, line 285: This method is no longer used with the argument

adekbadek (changes requested):
All good, just some cleanups are needed.

includes/class-newspack-popups-model.php
dkoo: Good catch, removed in a603104.

includes/class-newspack-popups-model.php
dkoo: Removed in a603104.

includes/class-newspack-popups-model.php
dkoo: Removed in a603104.

adekbadek (approved):

PR #403, opened 2021-01-19
fix: corrected logic for above header display

jeffersonrabb:
### All Submissions:

There's currently a bug where Above Header campaigns will only be shown if they happen to have a timer trigger, which is an unused feature for this type of campaign. This corrects the issue.

1. On `master`, create a new campaign. Switch it to be an overlay, and set the trigger to scroll progress. Switch Inline back on, and set placement to Above Site Header.
2. Observe the campaign will not display in previews or front end. 
3. Switch to this branch, observe the campaign displays normally.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -512,7 +512,7 @@ final class Newspack_Popups_Model {
 		if ( ! isset( $popup['options'], $popup['options']['placement'] ) ) {
 			return false;
 		}
-		return in_array( $popup['options']['placement'], [ 'inline', 'above_header' ] );
+		return in_array( $popup['options']['placement'], self::$inline_placements );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


includes/class-newspack-popups-model.php
```diff
@@ -512,7 +512,7 @@ public static function is_inline( $popup ) {
 		if ( ! isset( $popup['options'], $popup['options']['placement'] ) ) {
 			return false;
 		}
-		return 'inline' === $popup['options']['placement'];
+		return in_array( $popup['options']['placement'], [ 'inline', 'above_header' ] );
```
adekbadek, line 515:
```suggestion
		return in_array( $popup['options']['placement'], self::$inline_placements );
```

adekbadek (changes requested):

includes/class-newspack-popups-model.php
jeffersonrabb: Nice https://github.com/Automattic/newspack-popups/pull/403/commits/099ce49ac213a8d75666db921e1450e4be7b234e

adekbadek (approved):

PR #404, opened 2021-01-19
fix: view-as-segment should always take min count values

jeffersonrabb:
### All Submissions:

When using view-as-segment for a segment that has 0 values for read counts, the logic will not take these values from the segment and instead falls back to any legitimate views that have been logged in the current browser. This leads to confusing, sometimes incorrect results where previewing may work correctly for some users and incorrectly for other based on their history of viewing the site. This branch resolves the issue by always adopting the `min_posts` and `min_session_posts` verbatim from the segment, regardless of whether they are zero or not. `max_posts` and `max_session_posts` are ignored, since being precisely at the min values is sufficient for matching the segment.

A separate question this raises is whether segments with 0 counts should even be allowed, but that's for later.

1. Switch to `master`, open an incognito window, log in to test site. 
2. Create several segments with no counts (e.g. set a referrer only, or check one of the Donor checkboxes) and one with a read count min 1.
3. Create a few campaigns and assign to one of the non read-count segments. Assign one campaign to the read count 1 segment
4. Preview each segment, observe the campaigns are shown correctly.
5. Log out of WordPress and view several pages on the site.
6. Log back in and preview each segment. This time, observe that you see the campaign assigned to the read count 1 segment when previewing any segment, which is incorrect.
7. Switch to this branch, repeat step 6, and verify that you see only the campaigns you'd expect to see with each segment.
8. Test a variety of other segments and campaigns to ensure view-as-segment is working properly.

### Other information:

### DIFF
--- a/api/campaigns/class-campaign-data-utils.php
+++ b/api/campaigns/class-campaign-data-utils.php
@@ -124,7 +124,7 @@ class Campaign_Data_Utils {
 			$is_subscriber = $view_as_segment->is_subscribed;
 			$is_donor      = $view_as_segment->is_donor;
 			if ( ! empty( $view_as_segment->referrers ) ) {
-				$first_referrer = array_map( 'trim', explode( ',', $campaign_segment->referrers ) )[0];
+				$first_referrer = array_map( 'trim', explode( ',', $view_as_segment->referrers ) )[0];
 				if ( strpos( $first_referrer, 'http' ) !== 0 ) {
 					$first_referrer = 'https://' . $first_referrer;
 				}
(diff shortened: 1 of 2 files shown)


dkoo (approved):
This is working well for me. I created various segments using every type of config option, and previewing each I saw only the expected campaigns.

jeffersonrabb:
> This is working well for me. I created various segments using every type of config option, and previewing each I saw only the expected campaigns.

Great! I'm not going to merge until @adekbadek has had a chance to take a peek, just in case there's something about view-as I'm not thinking of.

adekbadek:
I've pushed some other fixes & tests to this branch, all regarding issues stemming from using view-as when some real client data is available. 

The bug in the PR description was not reproducible for me - [unless the segments in question had some read counts](https://github.com/Automattic/newspack-popups/pull/404/files#diff-0eb8371b72c3645321e39c8bae29c8c4ef73c133564e8d4e09cdd952aa6f5c00R1026-R1048).

adekbadek (approved):

dkoo (approved):
Re-tested and confirmed the referrer/category bugs and the fixes. Thanks!

PR #415, opened 2021-01-25
Referrer exclusion segmentation

### LINKED ISSUE #386: Segmentation By Referrer Suppression
In addition to the existing referrer logic, it would be useful to create a segment that does _not_ show to users who have come in from a particular domain, i.e. a pop-up that would be shown to everyone except traffic from google.com.

adekbadek:
### All Submissions:

Closes #386 

Follow testing steps in https://github.com/Automattic/newspack-plugin/pull/831

### Other information:

### DIFF
--- a/api/campaigns/class-campaign-data-utils.php
+++ b/api/campaigns/class-campaign-data-utils.php
@@ -34,11 +34,10 @@ class Campaign_Data_Utils {
 	 * @param string $referrers_list_string Comma-separated referrer domains list.
 	 */
 	public static function does_referrer_match( $page_referrer_url, $referrers_list_string ) {
-		$referer_domain = parse_url( $page_referrer_url, PHP_URL_HOST ); // phpcs:ignore WordPress.WP.AlternativeFunctions.parse_url_parse_url, WordPress.Security.NonceVerification.Recommended, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
-		// Handle the 'www' prefix – assume `www.example.com` and `example.com` referrers are the same.
-		$referer_domain_alternative = strpos( $referer_domain, 'www.' ) === 0 ? substr( $referer_domain, 4 ) : "www.$referer_domain";
-		$referrer_matches           = array_intersect(
-			[ $referer_domain, $referer_domain_alternative ],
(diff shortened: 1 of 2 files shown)


api/campaigns/class-campaign-data-utils.php
```diff
@@ -27,6 +27,23 @@ public static function is_donor( $client_data ) {
...
+	/**
+	 * Compare page referrer to a list of referrers.
+	 *
+	 * @param string $page_referrer_url Referrer to compare to.
+	 * @param string $referrers_list_string Comma-separated referrer domains list.
+	 */
+	public static function does_referrer_match( $page_referrer_url, $referrers_list_string ) {
+		$referer_domain = parse_url( $page_referrer_url, PHP_URL_HOST ); // phpcs:ignore WordPress.WP.AlternativeFunctions.parse_url_parse_url, WordPress.Security.NonceVerification.Recommended, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+		// Handle the 'www' prefix – assume `www.example.com` and `example.com` referrers are the same.
+		$referer_domain_alternative = strpos( $referer_domain, 'www.' ) === 0 ? substr( $referer_domain, 4 ) : "www.$referer_domain";
```
dkoo, line 39: This made me realize that any subdomain other than `www` won't be considered a match. e.g. if I set my segment to exclude `google.com`, and the referrer is `https://gmail.google.com`, it won't be considered a match and the campaign will still be displayed. Also true of the regular referrer option. Should we clarify that URLs with subdomains must be specified in the field if they want them to match the segment?

dkoo (changes requested):

api/campaigns/class-campaign-data-utils.php
adekbadek: Now I wonder if we should maybe disregard the subdomain - just check the domain. Referral from `images.example.com` is still a referral from `example.com`, right?

dkoo: Yeah, that's what I was wondering. My gut says we should just look at the root domain and ignore subdomains.

adekbadek: Right, done in d16574b

dkoo (approved):
👍

PR #416, opened 2021-01-25
feat: segment priority UI and logic

### LINKED ISSUE #371: Segment Precedence
With the introduction of segmentation it's possible to have multiple campaigns be candidates to show to a single user. For inline campaigns this isn't a big deal, but for overlays and above-header placements this creates a problem. To resolve, we need to determine campaign precedence — in case of a conflict, which campaign "wins?" 

After discussing with @dkoo, @adekbadek and @thomasguillot we settled on this proposed approach:

1) In the Segmentation tag, segment `ActionCard`s can be dragged to establish an order. The order represents precedence from top to bottom.
2) If there is a conflict, the campaign whose segment has the higher weight wins.
3) If there are two campaigns that are in direct conflict, a warning is shown to the admin. This would occur if two published overlay campaigns have the same segment assigned. If the two have category assignments which differ then there's no conflict. These rules also apply to Above Header campaigns.
4) Sitewide default can be removed.

dkoo:
### All Submissions:

Along with https://github.com/Automattic/newspack-plugin/pull/832, introduces the concept of segment priority and the UI for managing it

Closes #371.

Follow testing instructions in https://github.com/Automattic/newspack-plugin/pull/832.

### Other information:


api/campaigns/class-maybe-show-campaign.php
```diff
@@ -74,15 +74,49 @@ function ( $post_data ) use ( $visit ) {
...
 				$page_referer_url,
 				$view_as_spec
 			);
+
+			// If an overlay is already able to be shown, pick the one that has the higher priority.
+			if ( $campaign_should_be_shown && 'o' === $campaign->t ) {
+				if ( empty( $overlay_to_maybe_display ) ) {
+					$overlay_to_maybe_display = $campaign;
+				} else {
+					$higher_priority = self::get_higher_priority( $overlay_to_maybe_display, $campaign, $settings->all_segments );
```
adekbadek, line 96: A naming suggestion, to make it more understandable:
```suggestion
					$higher_priority_item = self::get_higher_priority_item( $overlay_to_maybe_display, $campaign, $settings->all_segments );
```

includes/class-newspack-popups-inserter.php
```diff
@@ -55,87 +55,49 @@ public static function popups_for_post() {
...
-					) {
-						array_push(
-							$popups_to_maybe_display,
-							$found_popup
-						);
-					}
-				}
-			}
-		}
+		// 2. Get all inline and above-header popups.
```
adekbadek, line 72: In other places in the code and in the UI, an "above header" popup _is_ an inline one:

image

```suggestion
		// 2. Get all inline popups.
```

includes/class-newspack-popups-model.php
```diff
@@ -210,52 +147,75 @@ public static function set_popup_options( $id, $options ) {
...
+				],
+			];
+		}
+
 		return self::retrieve_popups_with_query( new WP_Query( $args ) );
 	}
 
 	/**
-	 * Retrieve all popups from a given group.
+	 * Retrieve all inline and above-header popups.
```
adekbadek, line 181: Same nitpick as in the other place:
```suggestion
	 * Retrieve all inline popups.
```

includes/class-newspack-popups-segmentation.php
```diff
@@ -417,6 +453,34 @@ function ( $client_data ) use ( $segment_config ) {
...
+		$index = 0;
+		array_map(
+			function( $segment ) use ( $index ) {
+				$segment['priority'] = $index;
+				$index++;
+			},
+			$segments
+		);
+
+		return $segments;
```
adekbadek, line 481: This didn't work for me because the `array_map` is not actually updating the `$segments` variable. Here's a solution:

```php
return array_map(
	function( $segment ) use ( &$index ) {
		$segment['priority'] = $index;
		$index++;
		return $segment;
	},
	$segments
);
```

(note the `&` before `$index` - it's there to make make the variable mutable)

adekbadek (changes requested):
Impressive work! I've found just some small issues.

includes/class-newspack-popups-inserter.php
dkoo: True, but we're treating above-headers differently than other inlines in the code. Still, I get what you're saying.

includes/class-newspack-popups-segmentation.php
dkoo: Ah, I always wondered what the `&$` meant. Fixed in 6a5f9ed, thanks.

includes/class-newspack-popups-model.php
dkoo: Updated in 6a5f9ed.

api/campaigns/class-maybe-show-campaign.php
dkoo: Updated in 6a5f9ed.

includes/class-newspack-popups.php
```diff
@@ -422,28 +421,6 @@ public static function previewed_popup_id() {
...
-					if ( $value ) {
-						return Newspack_Popups_Model::set_sitewide_popup( $post->ID );
-					} else {
-						return Newspack_Popups_Model::unset_sitewide_popup( $post->ID );
-					}
-				},
-			]
-		);
-	}
-
```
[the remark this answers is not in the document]
dkoo: Removed in 6a5f9ed.

includes/class-newspack-popups-segmentation.php
```diff
@@ -309,7 +309,43 @@ public static function create_database_table() {
 	 * Get all configured segments.
 	 */
 	public static function get_segments() {
-		return get_option( self::SEGMENTS_OPTION_NAME, [] );
+		$segments                  = get_option( self::SEGMENTS_OPTION_NAME, [] );
+		$segments_without_priority = array_filter(
+			$segments,
+			function( $segment ) {
+				return empty( $segment['priority'] );
```
adekbadek, line 316: This filter will always return at least one element, since one segment will have priority of `0` (which will satisfy the `empty()` condition).

includes/class-newspack-popups-segmentation.php
adekbadek: Still, the array passed as the argument to this method will not be updated - this can be verified by logging the value of `$segments` before and after the `array_map`.

adekbadek (changes requested):

includes/class-newspack-popups-segmentation.php
dkoo: Ooh, good catch. I think I actually want `! isset` instead, right? Updated in2958a5e.

includes/class-newspack-popups-segmentation.php
dkoo: Oh, that's weird. How about if I use `array_values` instead—that will reindex an ordered array with the same values, right? Trying that out in `2958a5e`. Kind of hard to test without spoofing an array value in `get_segments`.

adekbadek: This approach will not handle legacy segments, which do not have a `priority` key. 
This can be tested by `unset`ting a segment's `priority` right after option value retrieval in `get_segments` method. 
I'd still recommend the solution from the top comment.

dkoo: Ack, you're right. And I had missed a line from your first suggestion, so I thought it wasn't working. Reimplemented your solution without any missing code in 2b10b26 and it now seems to work for me 😄

adekbadek (approved):

PR #431, opened 2021-02-02
Fix/nomenclature updates

jeffersonrabb:
### All Submissions:

Applies recent naming changes to match the Newspack Plugin Campaigns Wizard. "Campaigns" are renamed to "Prompts." "Campaign Groups" are renamed "Campaigns." All of the changes in this branch apply to UI text and comments only. No function names or variables have been changed.

Go through the UI, see if you can find any incorrect references to "Campigns" or "Campaign Groups."

### Other information:


src/editor/SegmentationSidebar.js
```diff
@@ -42,7 +42,7 @@ const SegmentationSidebar = ( { onMetaFieldChange, selected_segment_id } ) => {
 				href="/wp-admin/admin.php?page=newspack-popups-wizard#/segmentation"
```
dkoo, line 42: Need to update this link to `#/segments` now that we've changed the route in https://github.com/Automattic/newspack-plugin/pull/833.

dkoo (changes requested):
Just found one un-updated link for the renamed "Segments" tab. Other than that, all mentions of "Campaigns" and "Campaign Groups"/"Groups" in the UI seem to be correctly updated to "Prompts" and "Campaigns".

It'll be another and bigger task to update all of our variable and function names!

src/editor/SegmentationSidebar.js
jeffersonrabb: Nice spot! Fixed in https://github.com/Automattic/newspack-popups/pull/431/commits/a48c8d91379e39f05719be3f2ae4f115b6aebe62

dkoo (approved):
Looks good now! Let me know if you need a re-review after fixing the conflicts.

jeffersonrabb:
> Let me know if you need a re-review after fixing the conflicts.

If you wouldn't mind, I think it would be wise for you to give another look. The merge conflicts were fairly substantial and mostly due to `feat/segment-prioritization`, and I want to be sure I didn't break anything.

dkoo (approved):
@jeffersonrabb I found one more instance where I thought we could update the nomenclature, pushed as f8c7a2f. If that looks good to you, this is 👍

jeffersonrabb:
> I found one more instance where I thought we could update the nomenclature

I noted that one and figured we could go either way since it could be interpreted to mean "disable (the) Campaigns (plugin) on this page." But your fix is good, let's keep it.

PR #432, opened 2021-02-04
feat: only show prompts for the best-priority segment

### LINKED ISSUE #445: Non-cascading segment matching
Because of how flexible our segmentation UI is, it's very easy to create segments that will overlap significantly with each other. Readers will very often match multiple segments—for example, given two segments:

1. Donors
2. Active Readers (5+ articles read in session)

It should be possible to create a prompt (e.g. a donation CTA) assigned to Active Readers that will never be shown to Donors, even if many donors will match both segments.

## Cascading segment matching

The current behavior is that if a reader matches more than one segment, we display prompts belonging to every segment the reader matches, plus prompts with no segment (a.k.a. "everyone"), as long as the prompts don't conflict with each other (e.g. multiple overlays or above-header). This makes it easy to show prompts to multiple segments, but it also makes it more difficult to ensure that certain prompts won't be displayed to specific segments. 

In the above example, with cascading segment matching, a prompt assigned to Active Readers will also be shown to any Donors who have read 5+ articles in a session.

## Non-cascading segment matching

The preferred behavior as described by @brianboyer is _non-cascading_ segment matching. This means that we will only show prompts belonging to the **highest-priority** segment that the reader matches, plus prompts with no segment ("everyone"). If this highest-priority segment contains no prompts, then the reader sees only the "everyone" prompts.

In the above example, with non-cascading segment matching, creating a prompt that is shown to Active Readers but not Donors becomes much easier. We just need to ensure that Donors is set to a higher priority than Active Readers, and then readers who match the Donors segment will not see any prompts assigned to Active Readers, even if they match both segments.

Non-cascading segment matching also makes previewing segments much easier to understand. Since we only display prompts in the highest-priority matching segment, previewing as a segment assumes that segment is the highest-priority matching segment, and no other prompts are shown (other than "everyone" prompts). You will never see prompts assigned to a different segment when previewing a specific segment, which often happens with cascading segment matching.

dkoo:
### All Submissions:

An experiment. Changes the concept of segment priority so that instead of showing prompts belonging to all of the reader's matching segments, we _only_ show the prompts that match the reader's highest-priority segment. This applies to prompts of all types (overlay, above-header, and inline), but we can choose to ignore this for inlines if we want.

With this paradigm, prompts with the default/no segment (here renamed as "Everyone" to more accurately convey its behavior along with https://github.com/Automattic/newspack-plugin/pull/853) take on increased importance. While the "Everyone" segment is still always prioritized last when it comes to conflict resolution for overlays and above-headers—so that any overlay or above-header prompt assigned to a reader-matching segment will be displayed instead of those assigned to "Everyone"—any _non-conflicting_ prompt assigned to the "Everyone" segment will be displayed to all readers, regardless of segment.

This also somewhat simplifies the previewing logic. Now, since there's no cascade of matching segments, when previewing a segment, we only need to show the prompts belonging to that segment plus any prompts belonging to "Everyone".

~Adding the "On Hold" status because we're not sure we want to go with this approach, but we want to test it to see how it "feels" from the publisher's perspective. Note that unit tests are failing now; if we decide we want to go with this new approach, I'll update the tests to pass.~ I updated the tests with the new functions, so now they're all passing once again. We could potentially add more tests for the non-cascading functionality, but I'm open to doing ideas.

Closes #445.

1. On this branch, create several overlay, above-header and inline prompts belonging to several different segments with varying overlapping criteria, and at least one prompt belonging to the "Everyone" segment. Examples of segments which can easily overlap:
  - Segment A is New Readers, or max articles read = 10. This should be set to the highest priority.
  - Segment B is Non-subscribers, with Newsletter set to non-subscribers only. This should be set to the middle priority.
  - Segment C is Non-donors, with Donor set to non-donors only. This should be set to the lowest priority.
2. In the Campaigns tab, preview each segment. Confirm that you see only published and unpublished prompts belonging to that segment, plus any prompts belonging to "Everyone", and no others.
  - Also try previewing "Everyone" and confirm that you only see prompts belonging to "Everyone".
  - If there's more than one overlay or above-header with the same segment, you'll still only see the most recently created one (this behavior is unchanged).
3. In a fresh incognito session, view an article on the site. In a fresh session, you should match all three of the example segments above. Confirm that because Segment A is the highest-priority segment, you only see the prompts belonging to that segment, plus any prompts belonging to the "Everyone" segment.
4. In the admin, reorder the segments so that Segment B is now the highest-priority segment. View an article in a new incognito session and this time confirm that you only see prompts belonging to Segment B, plus any prompts belonging to the "Everyone" segment.
5. Reorder segments so that Segment C is now the highest-priority segment. View an article in a new incognito session and this time confirm that you only see prompts belonging to Segment C, plus any prompts belonging to the "Everyone" segment.
6. Remove all prompts from Segment C, keeping it as the highest-priority segment. View the site in a fresh incognito session and confirm that you now see only prompts belonging to the "Everyone" segment.
7. Confirm that the conflict resolution aspect of segment priority still works: publish two overlays and above headers and assign to different segments. When viewing in a fresh incognito session, you should now only see overlays and above-header prompts if they belong to the highest-priority segment that you match as a reader.

### Other information:


dkoo:
Noting that this is on hold until we have a solution for #442.

api/campaigns/class-maybe-show-campaign.php
```diff
@@ -73,27 +73,42 @@ function ( $post_data ) use ( $visit ) {
 			);
 		}
 
-		$page_referer_url = isset( $_REQUEST['ref'] ) ? $_REQUEST['ref'] : ''; // phpcs:ignore WordPress.Security.NonceVerification.Recommended, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
-
+		$referer_url                   = filter_input( INPUT_SERVER, 'HTTP_REFERER', FILTER_SANITIZE_STRING );
+		$page_referer_url              = isset( $_REQUEST['ref'] ) ? $_REQUEST['ref'] : ''; // phpcs:ignore WordPress.Security.NonceVerification.Recommended, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+		$all_segments                  = isset( $settings->all_segments ) ? $settings->all_segments : [];
+		$best_priority_segment         = $this->get_best_priority_segment( $all_segments, $client_id, $referer_url, $page_referer_url, $view_as_spec );
```
adekbadek, line 79: Maybe we could skip calling `get_best_priority_segment` if previewing, and simply assign 
```
$best_priority_segment = view_as_spec['segment'];
```
?

api/campaigns/class-maybe-show-campaign.php
```diff
@@ -118,10 +133,58 @@ function ( $post_data ) use ( $visit ) {
...
+				$client_matches_segment = Campaign_Data_Utils::does_client_match_segment(
+					$segment,
+					$client_data,
+					$referer_url,
+					$page_referer_url
+				);
+
+				// Find the matching segment with the highest priority.
+				if ( $client_matches_segment ) {
+					$segment_priority = isset( $segment->priority ) ? $segment->priority : $best_segment_priority;
```
adekbadek, line 175: When would a segment not have a priority, and why is the `$best_priority_segment` value the default in this case? If it's about a default value, perhaps the `Campaign_Data_Utils::canonize_segment` would be a more suitable place for that.

api/campaigns/class-maybe-show-campaign.php
```diff
@@ -118,10 +133,58 @@ function ( $post_data ) use ( $visit ) {
 
 			$response[ $campaign->id ] = $campaign_should_be_shown;
 		}
+
 		$this->response = $response;
 		$this->respond();
 	}
 
+	/**
+	 * Get the highest-priority segment that matches the client.
```
adekbadek, line 142: Nit: it's a bit fuzzy, cause the "best" priority is the _lowest_ priority integer ;)

api/campaigns/class-maybe-show-campaign.php
```diff
@@ -238,15 +286,31 @@ public function should_campaign_be_shown( $client_id, $campaign, $settings, $ref
...
+		// Using "view as" feature.
+		if ( $view_as_spec ) {
+			$should_display = false;
+			if ( isset( $view_as_spec['segment'] ) && $view_as_spec['segment'] ) {
+				// Show prompts with a matching segment, or "everyone". Don't show any prompts that don't match the previewed segment.
+				if ( $view_as_spec['segment'] === $campaign->s || empty( $campaign->s ) ) {
+					$should_display = true;
+				}
+
+				// Show prompts if the view_as segment doesn't exist.
```
adekbadek, line 310: Is that a possible UI state?

adekbadek (changes requested):

api/campaigns/class-maybe-show-campaign.php
dkoo: Yes, if you preview as "Everyone" (no segment), `view_as_spec` is set but without a `segment` property.

api/campaigns/class-maybe-show-campaign.php
dkoo: Ah yes, I had started out calling it the "highest priority" segment everywhere, but switched to calling it "best priority" instead for this exact reason. Updated the last remaining mentions of "highest priority" in 9609176.

api/campaigns/class-maybe-show-campaign.php
dkoo: To answer your question, I just set it to `$best_segment_priority` so that when the comparison happens in the next line, a segment with no priority is never possibly treated as having a better priority than another segment.

But I think you're right that it's unnecessary, because a segment should never not have a priority (and we ensure this in the segmentation class). Removed the check and default value in 1b88cf7.

api/campaigns/class-maybe-show-campaign.php
dkoo: Good idea—I had some logic to assign the `view_as_spec['segment']` if it exists inside the `get_best_priority_segment` method, but it can be simplified further. How about immediately returning the `view_as_spec['segment']` value from that method if it exists, like in 5b51622?

includes/class-newspack-popups-model.php
```diff
@@ -237,10 +237,6 @@ public static function retrieve_inline_popups( $include_unpublished = false, $ca
 	public static function retrieve_category_overlay_popups( $include_unpublished = false, $campaign_id = false ) {
 		$post_categories = get_the_category();
 
-		if ( empty( $post_categories ) ) {
-			return null;
-		}
```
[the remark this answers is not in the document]
dkoo: Thanks for catching, this was a mistake—re-added in 2a41b76.

api/campaigns/class-maybe-show-campaign.php
```diff
@@ -132,24 +189,22 @@ function ( $post_data ) use ( $visit ) {
 	 * @param string $page_referer_url URL of the referrer of the frontend page that is making the API request.
 	 * @param object $view_as_spec "View As" specification.
 	 * @param string $now Current timestamp.
+	 * @param bool   $check_segment Whether to check if the client matches the prompt's segment. If false, the client is assumed to match the segment.
```
adekbadek, line 192: I may be missing something here, but `should_campaign_be_shown` is called only [here](https://github.com/Automattic/newspack-popups/blob/9609176353d81f454cc4f781548844dd744e5825/api/campaigns/class-maybe-show-campaign.php#L103), with `$check_segment` of value `false`. Therefore, why is this argument needed?

api/campaigns/class-maybe-show-campaign.php
```diff
@@ -73,27 +73,42 @@ function ( $post_data ) use ( $visit ) {
...
+		// Check each matching campaign against other global factors.
+		foreach ( $matching_campaigns as $campaign ) {
 			$campaign_should_be_shown = $this->should_campaign_be_shown(
 				$client_id,
 				$campaign,
 				$settings,
 				filter_input( INPUT_SERVER, 'HTTP_REFERER', FILTER_SANITIZE_STRING ),
-				$page_referer_url,
-				$view_as_spec
+				'',
```
adekbadek, line 100: Why is the `$page_referer_url` argument hard-coded to `''` here?

adekbadek (changes requested):
Added some questions that did not occur to me earlier, sorry!

api/campaigns/class-maybe-show-campaign.php
dkoo: This was maybe a cheating way to get some failing tests to pass. In this branch, we check whether the client matches each segment, then only run `should_campaign_be_shown` on the campaigns with a matching segment. That caused many tests to fail unless we re-checked whether the client matches the segment in `should_campaign_be_shown`.

In `feat/multiple-segments`, I've gone back to running `should_campaign_be_shown` on every campaign, not just ones that match the segment, so I was able to remove the argument here: https://github.com/Automattic/newspack-popups/pull/451/files#diff-fab5ae21610a1add85a1640bebc4871f8fa855dd8f7c88d149e2f1c4c20df540R196

dkoo: Since this is fixed but with a lot of logical reshuffling in #451, I suggest we leave it for this branch since it doesn't cause any problems and will be overwritten immediately by merging #451.

api/campaigns/class-maybe-show-campaign.php
dkoo: This was a mistake. Fixed in 9f52331.

Also fixed in #451 here: https://github.com/Automattic/newspack-popups/pull/451/files#diff-fab5ae21610a1add85a1640bebc4871f8fa855dd8f7c88d149e2f1c4c20df540R103

adekbadek (approved):
Note: the plan is to merge https://github.com/Automattic/newspack-popups/pull/451 into this branch first, then merge both into master.

PR #435, opened 2021-02-09
Campaigns Wizard: mark draft prompts in preview

adekbadek:
### All Submissions:

When previewing, a user might be confused as to what will actually be visible for the reader. In this PR, styling for inactive (draft) prompts is introduced in the preview window.

1. Visit the Campaigns Wizard
2. Preview a segment
3. Observe that inactive (draft) prompts are marked with striped overlay and a label in top-left corner

### Other information:

### DIFF
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -783,6 +783,7 @@ final class Newspack_Popups_Model {
 		$classes                = [];
 		$classes[]              = 'above_header' === $popup['options']['placement'] ? 'newspack-above-header-popup' : null;
 		$classes[]              = 'inline' === $popup['options']['placement'] ? 'newspack-inline-popup' : null;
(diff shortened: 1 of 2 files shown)


thomasguillot:
@adekbadek in https://github.com/Automattic/newspack-popups/pull/435/commits/5410750cb61eec171a202439f576aadda1cdfe83 I made some adjustments

* I removed the stripped background because it felt too distracting and it would be hard for a user to have a real sense of what the prompt would be like on their website.
* The "inactive" label is now in black and white with system font to increase readability.

Should "inactive" be translatable?

dkoo (changes requested):
I agree that the "Inactive" label should be translatable. I also wonder if it should be more descriptive of the actual post status? e.g. "Draft", "Pending Review" or "Scheduled"

adekbadek:
Updated to show the actual (non-published) status.

thomasguillot (approved):

dkoo (approved):
Looks great!

PR #441, opened 2021-02-11
Simplify prompt retrieval

adekbadek:
### All Submissions:

With the changes in #416, separate retrieval of overlay and inline prompts is not necessary anymore, since the multiple overlays are allowed to be rendered on the page (API logic decides which is shown). 
Handling of the category condition was redundant, since categorised prompts are filtered out by `should_display` method.

This is mostly a refactor, with one user-facing change: the preview is now disregarding the category. Before, segment previews would not display a categorised prompt, unless the preview post had matching category (which could happen by chance). With changes here, all prompts tied to a campaign and/or segment will be displayed in the preview.

Mostly a refactor/performance improvement, so verify that everything works as expected

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -59,30 +59,12 @@ final class Newspack_Popups_Inserter {
 			return [];
 		}
 
-		$view_as_spec             = Segmentation::parse_view_as( Newspack_Popups_View_As::viewing_as_spec() );
-		$view_as_spec_campaign    = isset( $view_as_spec['campaign'] ) ? $view_as_spec['campaign'] : false;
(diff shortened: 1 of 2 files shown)


dkoo (approved):
Works really well and makes the logic of displaying prompts much easier to follow! Thanks for the refactor.

> the preview is now disregarding the category. Before, segment previews would not display a categorised prompt, unless the preview post had matching category (which could happen by chance). With changes here, all prompts tied to a campaign and/or segment will be displayed in the preview.

Is this something we need to explain, either in the UI or in documentation?

adekbadek:
> Is this something we need to explain, either in the UI or in documentation?

I don't think so, this change is very subtle.

PR #446, opened 2021-02-11
fix: increase max number of prompts to retrieve

jeffersonrabb:
### All Submissions:

This resolves an issue that can occur when a site has a substantial number of prompts. Most of the prompt queries do not have `posts_per_page` set, which means they will have a limit of 10. If there are many more than 10 prompts, there can be situations where prompts that should be shown are not returned by the query, and therefore not shown.

1. On `master` create at least one post with a category.
2. Create a new inline prompt, set the category to match the post.
3. Create at least 10 more inline prompts. Assign these to a segment that would not include first time visitor (e..g newsletter subscriber)
4. View the categorized post in an incognito window. Observe you do not see the prompt. 
5. Switch to this branch and refresh the page. Observe you now see the prompt.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -201,12 +201,12 @@ final class Newspack_Popups_Model {
 	 */
 	public static function retrieve_inline_popups( $include_unpublished = false, $campaign_id = false ) {
 		$args = [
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Nice catch, works as described 👍

PR #451, opened 2021-02-23
feat: allow multiple segments per prompt

### LINKED ISSUE #442: Multiple Segments Per Prompt
As part of the exploration of Campaigns with @brianboyer it's become clear that we need the ability to assign multiple segments to a prompt. The current one-to-one relationship is limiting and often forces the editor to create duplicate prompts. 

Suggested UI from @thomasguillot:

1) In Prompt editor sidebar:

Screenshot 2021-02-10 at 10 03 55

2) In Campaigns Wizard:

Screenshot 2021-02-10 at 10 10 15

dkoo:
### All Submissions:

Along with https://github.com/Automattic/newspack-plugin/pull/874, allows multiple segments to be assigned to each prompt. When deciding whether to show a particular prompt to a reader, all assigned segments are taken into account and the prompt is shown if one of the assigned segments is the reader's highest-priority matching segment.

Depends on #432.

Closes #442.

Follow testing instructions in https://github.com/Automattic/newspack-plugin/pull/874.

### Other information:

### DIFF
--- a/api/campaigns/class-maybe-show-campaign.php
+++ b/api/campaigns/class-maybe-show-campaign.php
@@ -79,6 +79,15 @@ class Maybe_Show_Campaign extends Lightweight_API {
 		$overlay_to_maybe_display      = null;
 		$above_header_to_maybe_display = null;
 
+		/**
+		 * When previewing, spoof client ID so that user actions performed outside the preview window don't affect prompt display.
+		 * If a `cid` param is passed as part of the view_as spec, use that so it will persist across page loads within the preview session.
+		 * Otherwise, generate a unique ID for the current page view only.
+		 */
+		if ( ! empty( $view_as_spec ) ) {
+			$client_id = isset( $view_as_spec['cid'] ) ? $view_as_spec['cid'] : uniqid( 'view_as_client_' );
+		}
+
 		if ( $settings ) {
 			$settings->best_priority_segment = $this->get_best_priority_segment( $all_segments, $client_id, $referer_url, $page_referer_url, $view_as_spec );
 			$this->debug['matching_segment'] = $settings->best_priority_segment;
--- a/api/campaigns/class-maybe-show-campaign.php
+++ b/api/campaigns/class-maybe-show-campaign.php
@@ -265,6 +265,7 @@ class Maybe_Show_Campaign extends Lightweight_API {
 				}
 			}
 		} elseif ( ! empty( $campaign_segment_ids ) ) {
+			// $settings->best_priority_segment_id should always be present, but in case it's not (e.g. in a unit test), we can fetch it here.
 			$best_priority_segment_id = isset( $settings->best_priority_segment_id ) ?
 				$settings->best_priority_segment_id :
 				$this->get_best_priority_segment_id( $settings->all_segments, $client_id, $referer_url, $page_referer_url, $view_as_spec );
--- a/api/campaigns/class-maybe-show-campaign.php
+++ b/api/campaigns/class-maybe-show-campaign.php
@@ -89,8 +89,8 @@ class Maybe_Show_Campaign extends Lightweight_API {
 		}
 
 		if ( $settings ) {
-			$settings->best_priority_segment = $this->get_best_priority_segment( $all_segments, $client_id, $referer_url, $page_referer_url, $view_as_spec );
-			$this->debug['matching_segment'] = $settings->best_priority_segment;
+			$settings->best_priority_segment_id = $this->get_best_priority_segment_id( $all_segments, $client_id, $referer_url, $page_referer_url, $view_as_spec );
+			$this->debug['matching_segment']    = $settings->best_priority_segment_id;
 		}
 
 		// Check each matching campaign against other global factors.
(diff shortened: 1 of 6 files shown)


adekbadek (changes requested):

api/campaigns/class-maybe-show-campaign.php
```diff
@@ -141,16 +143,17 @@ function( $campaign ) use ( $best_priority_segment ) {
 	/**
 	 * Get the highest-priority segment that matches the client.
 	 *
-	 * @param object      $all_segments All segments.
-	 * @param string      $client_id Client ID.
+	 * @param object      $all_segments Segments to check.
+	 * @param string|bool $client_id Client ID. If false, assume the client matches all given segments.
```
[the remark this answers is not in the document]
dkoo: This is a holdover from something I was trying before implementing the spoofed `client_id` when previewing. Removed in 52c9ff.

api/campaigns/class-maybe-show-campaign.php
```diff
@@ -262,19 +265,28 @@ public function should_campaign_be_shown( $client_id, $campaign, $settings, $ref
 		}
 
 		// Handle segmentation.
-		$campaign_segment = isset( $settings->all_segments->{$campaign->s} ) ? $settings->all_segments->{$campaign->s} : false;
-		if ( ! empty( $campaign_segment ) ) {
+		$campaign_segment_ids = ! empty( $campaign->s ) ? explode( ',', $campaign->s ) : [];
+
+		if ( ! empty( $campaign_segment_ids ) ) {
+			$best_priority_segment = isset( $settings->best_priority_segment ) ?
```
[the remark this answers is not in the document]
dkoo: Good call, updated in 17bcfe9

src/editor/SegmentationSidebar.js
```diff
@@ -5,39 +5,94 @@
...
+			>
+				{ 0 < segmentsList.length && (
+					<FormTokenField
+						value={ [] }
+						onChange={ _segment => {
+							const segmentToAssign = segmentsList.find(
+								segment => segment.name === _segment[ 0 ]
+							);
+
+							if ( ! segmentToAssign.id ) {
```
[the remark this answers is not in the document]
dkoo: Good catch, fixed in 2c0cde6.

api/campaigns/class-maybe-show-campaign.php
```diff
@@ -76,30 +76,32 @@ function ( $post_data ) use ( $visit ) {
...
-				return ( $best_priority_segment === $campaign->s || empty( $campaign->s ) );
-			}
-		);
+		/**
+		 * When previewing, spoof client ID so that user actions performed outside the preview window don't affect prompt display.
+		 * If a `cid` param is passed as part of the view_as spec, use that so it will persist across page loads within the preview session.
+		 * Otherwise, generate a unique ID for the current page view only.
+		 */
+		if ( ! empty( $view_as_spec ) ) {
+			$client_id = isset( $view_as_spec['cid'] ) ? $view_as_spec['cid'] : uniqid( 'view_as_client_' );
```
adekbadek, line 88: It's not working as expected. When I preview, "once" frequency & dismissed prompts are displayed again on navigation. Client ID spoofing should probably implemented at a deeper level, either along with any occurrence of `CLIENT_ID()` AMP "function", or by replacing the cookie upon a preview request.
Either way, this can be tackled in a separate PR, too.

api/campaigns/class-maybe-show-campaign.php
```diff
@@ -256,19 +257,28 @@ public function should_campaign_be_shown( $client_id, $campaign, $settings, $ref
 		}
 
 		// Handle segmentation.
-		$campaign_segment = isset( $settings->all_segments->{$campaign->s} ) ? $settings->all_segments->{$campaign->s} : false;
-		if ( ! empty( $campaign_segment ) ) {
+		$campaign_segment_ids = ! empty( $campaign->s ) ? explode( ',', $campaign->s ) : [];
+
+		if ( ! empty( $campaign_segment_ids ) ) {
+			$best_priority_segment_id = isset( $settings->best_priority_segment_id ) ?
```
adekbadek, line 263: Why is the `best_priority_segment_id` set again here? It's already assigned on this line:

https://github.com/Automattic/newspack-popups/blob/88666acd412fa0e567846308f5bb094d06d1cab7/api/campaigns/class-maybe-show-campaign.php#L92

adekbadek (changes requested):

api/campaigns/class-maybe-show-campaign.php
dkoo: This is a failsafe, mainly for testing purposes. I set it on line 92 to avoid running `get_best_priority_segment_id` for each prompt (instead, we run it once and cache the result in `$settings`). But if the property doesn't exist on `$settings` for whatever reason, we can still get it at this point.

api/campaigns/class-maybe-show-campaign.php
dkoo: > When I preview, "once" frequency & dismissed prompts are displayed again on navigation.

You're right, I see that even though the `cid` is persisting across navigations within the preview window, the client and campaign data aren't being saved. But this is true on `master` too, isn't it?

The main thing I'm trying to fix in this branch is that because we're no longer spoofing the actual segment (instead, we're just showing prompts if their assigned segments match the previewed segment), the actual client info is still being used to determine if prompts have been seen once or suppressed forever. What I see without a spoofed `cid`:

* As a logged in user, I have previously made test donations and subscribed to the newsletter. This is remembered by the preview session and can be confirmed by adding `$this->debug['client_data'] = $this->get_client_data( $client_id )['posts_read'];` to output the client data.
* As the same user, when previewing, this client data caused `suppress_forever` to be set to `true` on newsletter and donation prompts in the preview.
* Or, if I preview a segment with a "once" prompt, the prompt is recorded as having been seen once. If I then close the preview and preview again, the prompt isn't shown because I've already seen it.

For purposes of this PR, what do you think of keeping the spoofed `cid` so at least the expected prompts are seen with each preview, even if user actions and frequency aren't taken into account across navigation within the preview? Then, as you say, there might be a better way we can make previews match up to the front-end experience in another PR.

adekbadek: Took a stab at spoofing the cookie in the plugin in https://github.com/Automattic/newspack-plugin/pull/891 & https://github.com/Automattic/newspack-popups/pull/458 - maybe this would solve the issue?

dkoo: Yeah, let's handle the cookie spoofing in those other two PRs. I've removed all spoofing behavior from this PR and https://github.com/Automattic/newspack-plugin/pull/874 for now.

api/campaigns/class-maybe-show-campaign.php
adekbadek: I think a comment should be added with an explanation, because looking at this class it's hard to imagine circumstances in which this property is not set.

dkoo: Added a comment explaining this in 378e969.

adekbadek (approved):

PR #458, opened 2021-02-26
Handle visit data also in previews

adekbadek:
### All Submissions:

Enable cookie spoofing when previewing.

TODO/Question: this would spam the logs and DB with the preview-generated data – perhaps it should be removed in a CRON job?

Follow instructions in https://github.com/Automattic/newspack-plugin/pull/891

### Other information:

### DIFF
--- a/api/campaigns/class-maybe-show-campaign.php
+++ b/api/campaigns/class-maybe-show-campaign.php
@@ -37,7 +37,7 @@ class Maybe_Show_Campaign extends Lightweight_API {
 			$view_as_spec = Segmentation::parse_view_as( json_decode( $_REQUEST['view_as'] ) ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
 		}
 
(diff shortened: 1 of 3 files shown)


dkoo:
> TODO/Question: this would spam the logs and DB with the preview-generated data – perhaps it should be removed in a CRON job?

I thought about this too. Could the preview UI clean up any preview user data when the preview modal is closed? Either way we should probably not launch this feature without a solution for this, right?

dkoo (approved):
Approving because this works as described, but we'll need to fix conflicts with `master` and probably also answer the unresolved question about DB and log cleanup before we merge this.

adekbadek:
Proposed solution to the preview data logging issue: disregard it when fetching all clients data (used in computing the potential segment reach). This should suffice, since this is the only case when preview data can manifest itself as noise, and we just store last 30 days worth of data anyway.

dkoo (approved):
> Proposed solution to the preview data logging issue: disregard it when fetching all clients data (used in computing the potential segment reach). This should suffice, since this is the only case when preview data can manifest itself as noise, and we just store last 30 days worth of data anyway.

Oh, this is a great solution. Thanks for implementing! Let's get this merged.

PR #464, opened 2021-03-03
Prevent segmentation-derived custom dimensions reporting in non-interaction mode

adekbadek:
### All Submissions:

Ensures non-interaction mode sticks to the promise of disabling API calls.

1. On `master`, verify that when non-interaction mode is on, a call is made to the API on page load, and that segmentation-derived custom dimensions are reported*
2. Switch to this branch, observe the above is not true anymore
3. Turn off the non-interaction mode and verify that segmentation-derived custom dimensions are reported
4. Add a `NEWSPACK_POPUPS_DISABLE_REPORTING_CUSTOM_DIMENSIONS` flag of value `true` to `wp-config.php`
5. Observe that the API calls for prompts are made, but not for custom GA config, and that the segmentation-derived custom dimensions are not reported. This flag is for additional debugging purposes.

\* look for `collect' calls in network tab. The custom dimensions have to be configured in the Analytics wizard

### Other information:

### DIFF
--- a/includes/class-newspack-popups-segmentation.php
+++ b/includes/class-newspack-popups-segmentation.php
@@ -55,9 +55,8 @@ final class Newspack_Popups_Segmentation {
 		add_action( 'init', [ __CLASS__, 'create_database_table' ] );
 		add_action( 'wp_footer', [ __CLASS__, 'insert_amp_analytics' ], 20 );
 
+		add_filter( 'newspack_custom_dimensions', [ __CLASS__, 'register_custom_dimensions' ] );
 		if ( ! Newspack_Popups_Settings::is_non_interactive() && ( ! defined( 'NEWSPACK_POPUPS_DISABLE_REPORTING_CUSTOM_DIMENSIONS' ) || true !== NEWSPACK_POPUPS_DISABLE_REPORTING_CUSTOM_DIMENSIONS ) ) {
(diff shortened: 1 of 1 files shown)


dkoo (changes requested):
This works as described, but I'm a little unsure about the behavior where the segmentation-related custom dimension options aren't even registered in the admin if non-interaction mode is on or the constant is defined.

Wouldn't it be true that the custom dimensions would still be set to report the segmentation data, they just wouldn't be reported? In that case, would it make more sense to show the options in the admin UI, but flag that it won't be reported if non-interaction mode is on, or the constant is `true`? Something like this crude mockup:

Screen Shot 2021-03-03 at 10 07 17 AM

dkoo:
Also noting that testing this requires the fix from https://github.com/Automattic/newspack-plugin/pull/896

adekbadek:
>  the behavior where the segmentation-related custom dimension options aren't even registered in the admin

Good catch, enabled custom dimensions registering in 5482467e5eceaa7ba9a020b339a7f6871d476ed4. I don't think a text in the UI is needed - non-interaction mode is meant to be used very sparingly.

dkoo (approved):
> I don't think a text in the UI is needed - non-interaction mode is meant to be used very sparingly.

I'm okay with this for now—we can always add at a later date in the plugin repo. Works as expected now 👍

PR #467, opened 2021-03-04
feat: custom placements

### LINKED ISSUE #444: Target Zones For Manual Placement Prompts
### The Problem

The current model for placing prompts on the homepage or other one-off pages, is fairly awkward and problematic. The editor creates prompts with placement set to "manual placement," then uses the shortcode block to put them in the desired placement. There are at least three problems with this approach:

1) If there are multiple prompts targeted to different segments for the same spot, you would need to stack a bunch of shortcode blocks, which creates a pretty awkward editor UI.
2) To replace a prompts with a different one, the editor would have to manually find and replace every shortcode used on the site.
3) Manual Placement campaigns are generally (always?) going to be inline, so in the scenario where shortcode blocks are stacked there could be cases where more than one are shown to a user who matches multiple segments. 

### Proposed Solution

- Introduce the idea of prompt "zones" or "slots."
- New Gutenberg block allows a defined zone to be placed anywhere blocks are found. (Maybe zone definition actually occurs in the block?)
- Individual prompts placement UI will include the names of all zones. 
- Prompts whose placement is a zone will be displayed there according to normal rules of segmentation and precedence. 
- Much like overlays and above header placements, there can only be one prompt shown per zone. 

### Open Questions

- Should the block actually show the contents of the selected prompt, or should it just be the name of the zone? And... if the block does show the selected prompt and there are multiple options, which one should be shown?
- Where/how are zones defined?
- What is this thing called? Zone? Slot? Other?

cc: @brianboyer @dkoo @adekbadek @thomasguillot

dkoo:
### All Submissions:

Along with https://github.com/Automattic/newspack-plugin/pull/898, creates a new system for placing prompts manually in post or page content. Prompts assigned to a custom placement will be rendered on the page and displayed according to segmentation rules (one prompt per segment per block).

This MVP provides three custom placements that can be placed in post content via a new Custom Placement block. Could be extended to allow editors to create and update their own custom placements, too, but this isn't included yet.

Closes #444.

Follow testing instructions in https://github.com/Automattic/newspack-plugin/pull/898.

### Other information:

### DIFF
--- a/api/campaigns/class-maybe-show-campaign.php
+++ b/api/campaigns/class-maybe-show-campaign.php
@@ -127,11 +127,17 @@ class Maybe_Show_Campaign extends Lightweight_API {
 
 			// Handle custom placements: Only one prompt should be shown per placement block.
 			// "Everyone" prompts should only be shown if the reader doesn't match any segments.
-			if ( ! empty( $campaign->c ) ) {
-				if ( $campaign_should_be_shown && ! isset( $custom_placements_displayed[ $campaign->c ] ) ) {
-					$custom_placements_displayed[ $campaign->c ] = $campaign->id;
+			if ( $campaign_should_be_shown && ! empty( $campaign->c ) ) {
+				if ( ! isset( $custom_placements_displayed[ $campaign->c ] ) ) {
+					$custom_placements_displayed[ $campaign->c ] = $campaign;
 				} else {
-					$campaign_should_be_shown = false;
+					$previous_item        = $custom_placements_displayed[ $campaign->c ];
+					$higher_priority_item = self::get_higher_priority_item( $previous_item, $campaign, $all_segments );
+
+					// If the previous prompt in this custom placement already has a higher priority, only show that one. Otherwise, show this one instead.
+					$response[ $previous_item->id ]              = $previous_item->id === $higher_priority_item->id;
+					$campaign_should_be_shown                    = $campaign->id === $higher_priority_item->id;
+					$custom_placements_displayed[ $campaign->c ] = $higher_priority_item;
 				}
 			}
 
--- a/api/campaigns/class-maybe-show-campaign.php
+++ b/api/campaigns/class-maybe-show-campaign.php
@@ -78,6 +78,7 @@ class Maybe_Show_Campaign extends Lightweight_API {
 		$all_segments                  = isset( $settings->all_segments ) ? $settings->all_segments : [];
 		$overlay_to_maybe_display      = null;
 		$above_header_to_maybe_display = null;
+		$custom_placements_displayed   = [];
 
 		if ( $settings ) {
 			$settings->best_priority_segment_id = $this->get_best_priority_segment_id( $all_segments, $client_id, $referer_url, $page_referer_url, $view_as_spec );
@@ -121,6 +122,16 @@ class Maybe_Show_Campaign extends Lightweight_API {
 					$response[ $above_header_to_maybe_display->id ] = $above_header_to_maybe_display->id === $higher_priority_item->id;
 					$campaign_should_be_shown                       = $campaign->id === $higher_priority_item->id;
 					$above_header_to_maybe_display                  = $higher_priority_item;
+				}
+			}
+
+			// Handle custom placements: Only one prompt should be shown per placement block.
+			// "Everyone" prompts should only be shown if the reader doesn't match any segments.
+			if ( ! empty( $campaign->c ) ) {
+				if ( $campaign_should_be_shown && ! isset( $custom_placements_displayed[ $campaign->c ] ) ) {
+					$custom_placements_displayed[ $campaign->c ] = $campaign->id;
+				} else {
+					$campaign_should_be_shown = false;
 				}
 			}
 
--- a/includes/class-newspack-popups-custom-placements.php
+++ b/includes/class-newspack-popups-custom-placements.php
@@ -71,5 +71,18 @@ final class Newspack_Popups_Custom_Placements {
 	public static function get_custom_placement_values() {
 		return array_keys( self::get_custom_placements() );
 	}
+
+	/**
+	 * Determine whether the given prompt should be displayed via custom placement.
+	 *
+	 * @param object $prompt The prompt to assess.
+	 * @return boolean Whether or not the prompt has a custom placement.
+	 */
+	public static function is_custom_placement( $prompt ) {
+		return (
+			'manual' === $prompt['options']['frequency'] ||
+			in_array( $prompt['options']['frequency'], self::get_custom_placement_values() )
+		);
+	}
 }
 Newspack_Popups_Custom_Placements::instance();
--- a/includes/class-newspack-popups-custom-placements.php
+++ b/includes/class-newspack-popups-custom-placements.php
@@ -114,24 +114,8 @@ final class Newspack_Popups_Custom_Placements {
 				'methods'             => \WP_REST_Server::READABLE,
 				'callback'            => [ __CLASS__, 'api_get_prompts_for_custom_placement' ],
 				'permission_callback' => '__return_true',
-				'args'                => [
-					'custom_placement' => [
-						'validate_callback' => [ __CLASS__, 'validate_custom_placement' ],
-						'sanitize_callback' => 'esc_attr',
-					],
-				],
 			]
 		);
-	}
-
-	/**
-	 * Validate a custom placement slug.
-	 *
-	 * @param string $slug The slug of the custom placement to check.
-	 * @return boolean Whether the slug is a valid custom placement.
-	 */
-	public static function validate_custom_placement( $slug ) {
-		return in_array( $slug, self::get_custom_placement_values() );
 	}
 
 	/**
--- a/includes/class-newspack-popups-custom-placements.php
+++ b/includes/class-newspack-popups-custom-placements.php
@@ -10,18 +10,18 @@ defined( 'ABSPATH' ) || exit;
 /**
  * Main Newspack Placements Plugin Class.
  */
-final class Newspack_Popups_Placements {
+final class Newspack_Popups_Custom_Placements {
 	/**
 	 * The single instance of the class.
 	 *
-	 * @var Newspack_Popups_Placements
+	 * @var Newspack_Popups_Custom_Placements
 	 */
 	protected static $instance = null;
 
 	/**
 	 * Name of the option to store placements under.
 	 */
-	const PLACEMENTS_OPTION_NAME = 'newspack_popups_placements';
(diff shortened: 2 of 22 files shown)


includes/class-newspack-popups-inserter.php
```diff
@@ -88,7 +88,7 @@ public static function popups_for_post() {
 		$popups_to_maybe_display = array_filter(
```
adekbadek, line 88: There's a comment on the line above that should be updated.

includes/class-newspack-popups-inserter.php
```diff
@@ -88,7 +88,7 @@ public static function popups_for_post() {
...
 			$popups_to_maybe_display,
 			function( $popup ) {
-				return 'manual' !== $popup['options']['frequency'];
+				return ! Newspack_Popups_Custom_Placements::is_custom_placement( $popup );
```
adekbadek, line 91: Since the custom placement is a `placement`, not a `frequency`, wouldn't the custom-placement prompts already be excluded in the return values of `retrieve_*_popups` methods? If so, then this filtering here would not be needed.

includes/class-newspack-popups-custom-placements.php
```diff
@@ -0,0 +1,256 @@
...
+	/**
+	 * Determine whether the given prompt should be displayed via custom placement.
+	 *
+	 * @param object $prompt The prompt to assess.
+	 * @return boolean Whether or not the prompt has a custom placement.
+	 */
+	public static function is_custom_placement( $prompt ) {
+		return (
+			'manual' === $prompt['options']['frequency'] ||
+			in_array( $prompt['options']['frequency'], self::get_custom_placement_values() )
```
adekbadek, line 252: Shouldn't this be looking at `placement`, instead of `frequency` option?

tests/test-insertion.php
```diff
@@ -176,4 +176,24 @@ public function test_non_interactive_inline() {
...
+				'placement' => 'custom1',
+				'frequency' => 'always',
+			]
+		);
+
+		self::render_post();
+		self::assertNotContains(
+			self::$popup_content,
+			self::$post_content,
+			'Does not include the popup content, since it is a custom placement campaign.'
```
adekbadek, line 196: It'd be nice to include a test for rendering of a custom-placement prompt :)

adekbadek (changes requested):

src/blocks/custom-placement/view.php
```diff
@@ -0,0 +1,76 @@
...
+	// Get category-matching prompts for the custom placement.
+	$prompts = \Newspack_Popups_Custom_Placements::get_prompts_for_custom_placement( [ $custom_placement_id ], 'ids', $post_categories );
+
+	// If no category-matching prompts, get uncategorized prompts.
+	if ( empty( $prompts ) ) {
+		$prompts = \Newspack_Popups_Custom_Placements::get_prompts_for_custom_placement( [ $custom_placement_id ], 'ids', [] );
+	}
+
+	if ( ! empty( $prompts ) ) {
+		$content .= '<!-- start custom placement: ' . $custom_placement_id . '-->';
```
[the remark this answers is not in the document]
dkoo: What plugin info do you think we should print? Maybe we could print it if `WP_NEWSPACK_DEBUG` is on?

src/blocks/custom-placement/edit.js
```diff
@@ -0,0 +1,121 @@
...
+						<Notice status="warning" isDismissible={ false }>
+							{ __( 'No prompts found for this custom placement.', 'newspack-popups' ) }
+						</Notice>
+					) }
+					{ 0 < prompts.length && (
+						<>
+							<p>
+								{ sprintf(
+									__(
+										'This custom placement contains the following active prompt%s:',
```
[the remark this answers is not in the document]
dkoo: > placement will display at most all of the following active prompts

This isn't strictly true, either—each instance of the block can only display a single prompt from the reader's best-priority matching segment. But if you place multiple instances of the block in the same post, each one will display a single prompt (and if you place multiple instances of the block set to the same custom placement, each one will display the same single prompt).

In bc9df9b, I've attempted to make the list a bit more informative:

- The prompts are now organized by segment, more like in the Campaigns wizard
- The language now specifies "will display at most one of the following active prompts, depending on the reader's top priority segment"—hopefully this explains the logic a bit better.

includes/class-newspack-popups-inserter.php
dkoo: True, removed in b8344de.

includes/class-newspack-popups-inserter.php
dkoo: No longer relevant since this is removed in b8344de.

src/blocks/custom-placement/view.php
```diff
@@ -0,0 +1,76 @@
...
+ * @package Newspack_Popups
+ */
+
+namespace Newspack_Popups\Custom_Placement_Block;
+
+/**
+ * Dynamic block registration.
+ */
+function register_block() {
+	// Listings block attributes.
```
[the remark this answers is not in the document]
dkoo: Oops! Copy/paste exposed! 😆 Fixed in 58bff8e.

src/blocks/custom-placement/index.js
```diff
@@ -0,0 +1,46 @@
...
+	const isPrompt = Boolean( window.newspack_popups_data?.is_prompt );
+
+	// No prompts inside prompts.
+	if ( isPrompt ) {
+		return null;
+	}
+
+	registerBlockType( name, {
+		title: __( 'Newspack Campaigns: Custom Placement', 'newspack-listing' ),
+		icon: <CallToActionIcon />,
```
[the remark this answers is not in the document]
dkoo: Updated in 39406cc.

includes/class-newspack-popups-custom-placements.php
dkoo: True! 🤦 Fixed in f0b4679

tests/test-insertion.php
dkoo: Great idea! Added in 483cf2b

src/blocks/custom-placement/view.php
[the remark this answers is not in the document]
adekbadek: Just the plugin name, as suggested. Yeah I think we can use the `WP_NEWSPACK_DEBUG`. What I mean is that this HTML comment would not be needed in most situations, only when debugging.

src/blocks/custom-placement/edit.js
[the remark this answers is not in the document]
adekbadek: This looks good, however the block _may_ display multiple prompts. I've a non-subscribers segment and when a prompt assigned to this segment, and another one set to "everyone" segment, are assigned to the same custom placement, the block will display both.

includes/class-newspack-popups-custom-placements.php
```diff
@@ -0,0 +1,286 @@
...
+							'name'     => 'Everyone',
+							'priority' => PHP_INT_MAX,
+						],
+					];
+
+					if ( $segment_ids ) {
+						$segment_ids = explode( ',', $segment_ids );
+						$segments    = array_map(
+							function( $segment_id ) {
+								return Newspack_Popups_Segmentation::get_segment( $segment_id );
```
adekbadek, line 158: This might result in `null` if a segment was deleted - this array should be filtered (or reduce used instead of map)

adekbadek (changes requested):

src/blocks/custom-placement/edit.js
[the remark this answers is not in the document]
dkoo: Ack, you're right. Fixed in 390fc15—now, if a custom placement block has already displayed a prompt, no other prompts can be displayed.

src/blocks/custom-placement/view.php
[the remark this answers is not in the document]
dkoo: Ah, I understand now. Updated in 14f8b42.

includes/class-newspack-popups-custom-placements.php
dkoo: Good catch—updated in cde98b8.

api/campaigns/class-maybe-show-campaign.php
```diff
@@ -124,6 +125,16 @@ function ( $post_data ) use ( $visit ) {
 				}
 			}
 
+			// Handle custom placements: Only one prompt should be shown per placement block.
+			// "Everyone" prompts should only be shown if the reader doesn't match any segments.
```
adekbadek, line 129: I'm not finding that to be true in my testing. With two prompts in the custom placement, one assigned to non-subscribers, other to everyone, the latter is displayed (when viewing site as a non-subscriber). 

Seems that this happens simply because the everyone prompt is being evaluated earlier by this function, and so "takes the spot" of the prompt which should be shown in this custom placement. If the "everyone" prompts were evaluated last, this would work as expected, but I'm not sure if the prompts are ordered by segment priority.

adekbadek (changes requested):

src/blocks/custom-placement/edit.js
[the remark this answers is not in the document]
adekbadek: (continued here: https://github.com/Automattic/newspack-popups/pull/467#pullrequestreview-616667826)

api/campaigns/class-maybe-show-campaign.php
dkoo: Ah, you're right. I think this needed the same kind of priority comparison we're doing for overlays and above headers, but per custom placement block instead of on the whole page. How does it work for you now? 02dff0f should show the single highest-priority prompt for each custom placement regardless of the order the prompts are evaluated in.

adekbadek (approved):
When trying to understand what an issue I encountered was caused by, I've set up tests for the block, and then ended up fixing the issue (54b2593a93a863e27747ccafa76f494633ce4578). 
The issue was that with a conflicting pair of prompts, only one was rendered by the block. However, the API still received all the prompts, and could resolve to displaying a prompt not rendered on the page. 

To reproduce:
1. Create two prompts, assign to the same custom placement & segment (or no segment)
2.  Load the page, dismiss the prompt that is rendered
3. Reload the page, observe that before 54b2593a93a863e27747ccafa76f494633ce4578 no prompt is displayed, and after 54b2593a93a863e27747ccafa76f494633ce4578 the other prompt is

dkoo:
> When trying to understand what an issue I encountered was caused by, I've set up tests for the block, and then ended up fixing the issue ([54b2593](https://github.com/Automattic/newspack-popups/commit/54b2593a93a863e27747ccafa76f494633ce4578)).

Thanks for implementing these tests and fixing this issue!

PR #466, opened 2021-03-03
fix: prompt insertion with slideshows

dkoo:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Fix for a strange issue related to prompt insertion when a post contains Jetpack Slideshow blocks. This fix treats the symptom, but doesn't really solve the root issue, which is elusive.

Addresses #465, but perhaps doesn't close it. This is kind of a workaround which just checks whether the block's rendered content is much longer than expected based on its `innerHTML` attribute, and if it is, updates the inserter logic's internal position marker based on `innerHTML` instead of the usual rendered content. A potentially more ideal but much more involved solution is mentioned in the issue.

### How to test the changes in this Pull Request:

This is nearly impossible to replicate locally, in my experience. For this reason, I created a clone of the site that reported the issue, where the issue can be replicated reliably. The clone of the offending page is [here](https://mississippitoday.newspackstaging.com/2021/02/26/food-thats-going-to-stick-to-your-ribs-the-significance-of-soul-food-in-yalobusha-county/), but it can be replicated on any post on the test site that contains a slideshow block. DM me if you need the password to access the test site.

1. Ensure that the test site has the [latest production release](https://github.com/Automattic/newspack-popups/releases/tag/v1.26.0) of this plugin installed and active.
2. In a new incognito session, visit https://mississippitoday.newspackstaging.com/2021/02/26/food-thats-going-to-stick-to-your-ribs-the-significance-of-soul-food-in-yalobusha-county/ or any post that contains a slideshow block, in AMP mode (slideshow blocks [currently don't work in non-AMP mode with the Campaigns plugin active](https://github.com/Automattic/newspack-popups/issues/170), a known issue).  
3. Observe that the three prompts are displayed as described in #465. Also visit several other posts that don't contain a slideshow, and note the insertion locations for the prompts.
4. From this branch, run `npm run release:archive` to build the plugin code. Install and activate this plugin version on the test site.
[...]

PR #468, opened 2021-03-04
Fix placement computation

### LINKED ISSUE #465: Incorrect prompt insertion when post content contains jetpack/slideshow blocks
First reported by @joeboydston.

This is a hard one to replicate, but sometimes inline prompts will be inserted into the wrong areas of an article if the content contains a `jetpack/slideshow` block.

Example page where it's happening: http://mississippitoday.org/2021/02/26/food-thats-going-to-stick-to-your-ribs-the-significance-of-soul-food-in-yalobusha-county/

In a new session, three inline prompts are supposed to be injected at 25%, 50%, and 100% of the content, but where they're really being injected is all together toward the top of the article, like so:

Screen Shot 2021-03-03 at 3 23 27 PM

This is because the `$block_content` calculated [here](https://github.com/Automattic/newspack-popups/blob/master/includes/class-newspack-popups-inserter.php#L250) is sometimes much longer for Jetpack Slideshow blocks than expected—in the case of this page, it's almost as long as the total `$content` length, somehow. I can't determine why, but this causes the prompt insertion logic to think that the article is much longer than it really is, and then the prompts get inserted at the wrong locations.

**Expected:** The prompts should get inserted at roughly 25%, 50%, and 100% of the article content, with or without the slideshow block.

adekbadek:
### All Submissions:

Closes #465, supersedes #466

1. Create a post with a _massive_ jetpack slideshow block* at the top followed by at least five other blocks
2. Create five inline prompts, each at a position of 25% more than the last one
3. On `master`, observe the prompts are rendered at the top of the post content
4. Switch to this branch, observe the prompts are rendered at intended placements

\* Mine had about 80 images, beat that

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -221,7 +221,13 @@ final class Newspack_Popups_Inserter {
 			$content = scaip_maybe_insert_shortcode( $content );
 		}
 
-		$total_length = strlen( $content );
+		// Dynamic blocks might render an arbitrary amount of content, so the length of the content string is
+		// not an accurate representation of the content length.
+		$total_length = 0;
+		foreach ( parse_blocks( $content ) as $block ) {
+			$block_content = render_block( $block );
+			$total_length += strlen( wp_strip_all_tags( $block_content ) );
+		}
 
 		// 1. Separate prompts into inline and overlay.
 		$inline_popups  = [];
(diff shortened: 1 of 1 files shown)


dkoo (approved):
I couldn't beat 80 images in a slideshow, but 60 or so was enough to replicate and confirm the fix. Thanks for improving on #466!

PR #475, opened 2021-03-08
feat: do not log client data if site has no prompts

### LINKED ISSUE #473: [Question] Logging client data when there are no active prompts?
Currently, if a site has the Campaigns plugin active in a default state (visit logging enabled, non-interactive mode OFF), the site will log client visit data to the database via the segmentation API. Should we consider short-circuiting this functionality if the site lacks any active prompts? This might help avoid unintended database writes for sites that aren't ready to use campaign functionality.

dkoo:
### All Submissions:

I'm not sure if this would cause more harm than help, because of the extra query. But this adds a check of whether there's at least one active prompt before injecting the GA tracking code (which in turn calls the segmentation API and logs client data such as posts read, dismissals, etc.).

The intention here is to avoid writing client data to the DB as well as creating and updating transients for each reader who visits the site if the site isn't serving any prompts. Presumably there's no need to track this client data if the site has no need to segment readers for prompt display logic.

Note: #474 fixes an unrelated bug where shortcoded but unpublished prompts can also trigger the `save_client_data` request and therefore write client data even with the changes in this PR. 

Closes #473.

1. On `master`, with non-interactive mode OFF, unpublish or delete any active prompts.
2. In a new incognito session, visit the site front-end and in Dev Tools > Network, filter XHR requests by `api/segmentation`. Observe that a call is made that looks something like this (this is the API request that triggers writing client data to the DB):

```
https://<SITE URL>.com/wp-content/plugins/newspack-popups/api/segmentation/index.php?client_id=amp-QIR_V0fHXR8I-Km7G-eSIA&post_id=1259795&custom_dimensions={%220%22:{%22role%22:%22category%22,%22option%22:{%22value%22:%22category%22,%22label%22:%22Category%22},%22gaID%22:%22ga:dimension2%22},%221%22:{%22role%22:%22author%22,%22option%22:{%22value%22:%22author%22,%22label%22:%22Author%22},%22gaID%22:%22ga:dimension1%22},%223%22:{%22role%22:%22publish_date%22,%22option%22:{%22value%22:%22publish_date%22,%22label%22:%22Publish%20date%22},%22gaID%22:%22ga:dimension4%22}}&custom_dimensions_existing_values={%22dimension1%22:%22adminnewspack%22,%22dimension4%22:%222020-11-13%2012:03%22}&__amp_source_origin=https%3A%2F%2F<SITE URL>
```

3. To double-verify, you can `error_log` at [this line](https://github.com/Automattic/newspack-popups/blob/master/api/classes/class-lightweight-api.php#L286) and observe that the `save_client_data` function (which actually writes the transient with client data) is executed.
4. Check out this branch. Visit the site front-end and confirm that the `/segmentation` API request is no longer made.
5. (Note: only with #474 as well) Confirm via your `error_log` that the `save_client_data` function is never executed.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-segmentation.php
+++ b/includes/class-newspack-popups-segmentation.php
@@ -56,7 +56,7 @@ final class Newspack_Popups_Segmentation {
 		add_action( 'wp_footer', [ __CLASS__, 'insert_amp_analytics' ], 20 );
 
 		add_filter( 'newspack_custom_dimensions', [ __CLASS__, 'register_custom_dimensions' ] );
-		if ( ! Newspack_Popups_Settings::is_non_interactive() && ( ! defined( 'NEWSPACK_POPUPS_DISABLE_REPORTING_CUSTOM_DIMENSIONS' ) || true !== NEWSPACK_POPUPS_DISABLE_REPORTING_CUSTOM_DIMENSIONS ) ) {
+		if ( ! Newspack_Popups::is_empty() && ! Newspack_Popups_Settings::is_non_interactive() && ( ! defined( 'NEWSPACK_POPUPS_DISABLE_REPORTING_CUSTOM_DIMENSIONS' ) || true !== NEWSPACK_POPUPS_DISABLE_REPORTING_CUSTOM_DIMENSIONS ) ) {
 			// Sending pageviews with segmentation-related custom dimensions.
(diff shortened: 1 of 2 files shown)


adekbadek (commented):
Hm, if a site is not using any prompts, why would it have the plugin active? This seems to me like an over-optimization 🤔

PR #482, opened 2021-03-11
Remove default category on prompt activation

### LINKED ISSUE #447: "Uncategorized" Category Hidden In UI But Breaks Insertion Logic
If a prompt has the "uncategorized" category assigned it [will be suppressed in the UI](https://github.com/Automattic/newspack-popups/blob/master/includes/class-newspack-popups-inserter.php#L591-L596 ). However, the presence of this category will cause the prompt to be inserted like a categorized prompt, only in content bearing the same term. 

Ideally the solution is to ignore the "uncategorized" category in `retrieve_category_overlay_popups()`, or remove this tag every time the prompt is saved.

 If this proves difficult, a short-term half fix would be to remove the logic that hides the category, which at least will make it easier to diagnose this issue.

adekbadek:
### All Submissions:

Closes #447.

1. On `master`, raname your default category to something else than "Uncategorized"
2. Visit the Campaigns wizard, observe that the prompts have the default category assigned on activation
3. Switch to this branch; deactivate & activate a prompt - observe it does not have the default category assigned
4. In Campaign Wizard, assign the default category via the category filtering options (image), observe the category filtering is applied
5. Edit a prompt, assign the default category via the editor sidebar, observe the category is applied

### Other information:

### DIFF
--- a/includes/class-newspack-popups.php
+++ b/includes/class-newspack-popups.php
@@ -649,7 +649,7 @@ final class Newspack_Popups {
 	 * @param bool   $post Post.
 	 */
 	public static function remove_default_category( $new_status, $old_status, $post ) {
-		if ( self::NEWSPACK_POPUPS_CPT === $post->post_type && 'publish' === $new_status ) {
+		if ( self::NEWSPACK_POPUPS_CPT === $post->post_type && 'publish' !== $old_status && 'publish' === $new_status ) {
 			$default_category_id = (int) get_option( 'default_category', false );
 			$popup_has_category  = has_category( $default_category_id, $post->ID );
 			if ( $popup_has_category ) {
--- a/includes/class-newspack-popups.php
+++ b/includes/class-newspack-popups.php
@@ -53,6 +53,7 @@ final class Newspack_Popups {
 		add_action( 'enqueue_block_editor_assets', [ __CLASS__, 'enqueue_block_editor_assets' ] );
 		add_filter( 'display_post_states', [ __CLASS__, 'display_post_states' ], 10, 2 );
 		add_action( 'save_post_' . self::NEWSPACK_POPUPS_CPT, [ __CLASS__, 'popup_default_fields' ], 10, 3 );
+		add_action( 'transition_post_status', [ __CLASS__, 'remove_default_category' ], 10, 3 );
 
 		add_filter( 'show_admin_bar', [ __CLASS__, 'show_admin_bar' ], 10, 2 ); // phpcs:ignore WordPressVIPMinimum.UserExperience.AdminBarRemoval.RemovalDetected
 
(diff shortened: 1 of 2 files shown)


includes/class-newspack-popups.php
```diff
@@ -640,6 +641,23 @@ public static function archive_campaign( $id, $status ) {
...
 
+	/**
+	 * Prevent setting the default category when publishing.
+	 *
+	 * @param string $new_status New status.
+	 * @param string $old_status Old status.
+	 * @param bool   $post Post.
+	 */
+	public static function remove_default_category( $new_status, $old_status, $post ) {
+		if ( self::NEWSPACK_POPUPS_CPT === $post->post_type && 'publish' === $new_status ) {
```
dkoo, line 652: Currently, if you apply the default category after a prompt has been published (the only way to apply it after this PR), it will be removed the next time the post is updated. 

This should prevent the action from running when a prompt is updated after already being published:

```suggestion
		if ( self::NEWSPACK_POPUPS_CPT === $post->post_type && 'publish' !== $old_status && 'publish' === $new_status ) {
```

includes/class-newspack-popups.php
```diff
@@ -640,6 +641,23 @@ public static function archive_campaign( $id, $status ) {
...
+			$default_category_id = (int) get_option( 'default_category', false );
+			$popup_has_category  = has_category( $default_category_id, $post->ID );
```
dkoo, line 654: It's hard to ascertain the editor's intention in the use case where they may want to intentionally apply the default category. But maybe we can be a bit less heavy-handed by checking not just if the post has the default category, but if the default category is the post's _only_ category term?

dkoo (changes requested):
This is a nice and simple solution to a persistent source of frustration. However, this makes it nearly impossible to apply the default category to an active prompt if you for some reason _do_ want to apply it. I can see a use case if a publisher sets the default category to something other than "uncategorized" and actually uses it. Suggestions below.

includes/class-newspack-popups.php
adekbadek: This would prevent default-category-unsetting when the prompt is re-activated, which might happen often.

adekbadek:
> However, this makes it nearly impossible to apply the default category to an active prompt if you for some reason do want to apply it.

While it certainly looks so in code, I've found that it's not the case. I've added two testing steps in the PR description, please see if these work. 

A user can always assign categories via Campaigns Wizard - this does not transition the post status, so the code in this PR is not executed. As for the category-setting in the editor, I guess the transition hook runs after the default category is assigned automatically, but before it's assigned by the user.

includes/class-newspack-popups.php
dkoo: That's not what I see when I test. How can I replicate? Here's what I tested:

1. Create a new prompt and publish without setting any categories: the prompt is activated without the default category.
2. Deactivate the prompt.
3. Assign the default category (only).
4. Reactivate the prompt—the default category is removed.

dkoo:
> A user can always assign categories via Campaigns Wizard - this does not transition the post status, so the code in this PR is not executed. As for the category-setting in the editor, I guess the transition hook runs after the default category is assigned automatically, but before it's assigned by the user.

Ah yes, this is true. It's possible to assign the default category after publishing. But I find that without https://github.com/Automattic/newspack-popups/pull/482#discussion_r592730118, if I enter the post editor again, change the content or attributes and update from there (saving/updating an already active prompt), the category is removed. That's the main thing I'd like to fix before merging this. https://github.com/Automattic/newspack-popups/pull/482#discussion_r592732098 addresses an edge case that we probably don't need to consider, so I'm okay with leaving that.

adekbadek:
Ah, sorry - I did not fully understand the intention before. Added the change in 87dc564

dkoo (approved):
Thanks! Looks good to me now :)

PR #487, opened 2021-03-14
Allow filters to suppress popups better

claudiulodro:
### All Submissions:

This PR fixes a few small issues with the `newspack_newsletters_assess_has_disabled_popups` filter:

1. [The logic in the `get_post_meta( get_the_ID(), 'newspack_popups_has_disabled_popups', true )` check](https://github.com/Automattic/newspack-popups/blob/master/includes/class-newspack-popups-inserter.php#L123-L128) would override anything which hooked into the filter before that check. Best practices around filters are that a filter should only modify the return value if it is different than the default return value; that will correctly ensure that any other plugin is able to modify the return value. See the testing instructions for a more clear example of why this is an issue.
2. [The default return value in `::assess_has_disabled_popups` was `[]`](https://github.com/Automattic/newspack-popups/blob/master/includes/class-newspack-popups-inserter.php#L506) which technically works because it evaluates to `false` but isn't strictly correct. I've modified it to return a boolean.
3. The filter name was `newspack_newsletters_assess_has_disabled_popups`, which would be a great filter name if only it was located in Newspack Newsletters. :) I've modified it to `newspack_popups_assess_has_disabled_popups` and added backwards compatibility for the old hook name.

1. Add and activate the following plugin to your site, and name it `aaaaa.php`:

```
<?php
/**
 * Plugin Name: AAAAAAAAA
 * Description: AAAAAAAAAAAAAAAAAAAAAAAAAAAA!
 * Version: 1.0.0
 * Author: Automattic
 * Author URI: https://newspack.pub/
 * License: GPL2
 */

// Disable those popups!
add_filter( 'newspack_newsletters_assess_has_disabled_popups', function( $disabled ) {
	return true;
} );
```

2. Before applying this PR, create a prompt and view it on a post. With the above plugin active, you'd expect it to be suppressed, but it's not. This is because `aaaaa` is loaded before `newspack-popups` in WordPress (because of the alphabet), so the `get_post_meta` check in Newspack Campaigns runs after the filter in AAAAAAAAA.

Screen Shot 2021-03-14 at 11 28 09 AM

3. Apply this PR. Now the prompt should correctly get suppressed:

Screen Shot 2021-03-14 at 11 29 44 AM

4. Deactivate AAAAAAAAA to verify the default prompt behavior still works:

Screen Shot 2021-03-14 at 11 28 09 AM

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -121,7 +121,7 @@ final class Newspack_Popups_Inserter {
 		add_action( 'delete_widget', [ $this, 'remove_widgets_shortcoded_popup_ids' ] );
 
 		add_filter(
-			'newspack_newsletters_assess_has_disabled_popups',
+			'newspack_popups_assess_has_disabled_popups',
 			function ( $disabled ) {
 				if ( get_post_meta( get_the_ID(), 'newspack_popups_has_disabled_popups', true ) ) {
 					return true;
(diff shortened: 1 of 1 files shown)


adekbadek (approved):
💯

adekbadek:
(merging since Claudiu is AFK)

Issue #478, opened 2021-03-10
Database use optimisations


adekbadek:
Currently, the Lightweight API is not so lightweight on the DB. Some ideas for improvements:

- [x] **Empty client data**. For a first-time visitor to a non-post page, client data entry is created. This is not needed, unless they are a donor or a subscriber (#490). 
- [x] Currently, a row is created in the options table per client-prompt pair (e.g. with the information about prompts suppression). It'd be better to create save that information in the row which is created for the client (e.g. with information about posts read). In essence, it's about creating **not more than one row per client** (#490).
- The rows in the options table are never **purged**. The data in the custom table (for segmentation purposes) is removed after 30 days - but this is just data about read posts, which can be discarded. The subscriber & donor status should be retained. 
    - [ ] Perhaps a good solution would be to save the last client update in the client data, and remove rows (client data) not updated in more than ~two months (enabled by #508, but deploying that change would be possible only after changes from #508 are live for enough time). 
    - [x] The preview session data can safely be purged anytime (#508)
- [x] Client data is saved for **bot traffic** #509

cc @dkoo @jeffersonrabb

dkoo:
Also based on a recent high-traffic event:

[ ] Write client data to a custom table instead of the WP options table

adekbadek:
> We're using wp_cache_set and wp_cache_get and inserting their values directly into the options table. Why aren't we using the actual Transients API?

The direct use of the DB stems from the API being separate from WP for performance reasons.

adekbadek:
Putting this on hold, as we can only proceed with the stale client data removal after the update-tracking changes (#508) are live for about two months.

adekbadek:
> The rows in the options table are never purged. 

Better purging added in #584

PR #488, opened 2021-03-15
refactor: custom db table for campaign transients

jeffersonrabb:
### All Submissions:

The Campaigns plugin sets a transient for each prompt/user combination, which on a high traffic site is a source of significant options table bloat, causing performance issues. https://github.com/Automattic/newspack-popups/issues/478 lists a number of ways this can be optimized. This PR addresses one of these, creating a custom database table for Campaigns transients, taking the burden off the WordPress options table. At the moment when the table is created, all existing Campaigns transients are copied over. I opted not to delete the transients in the options table as this would be a potentially destructive action. We can manually prune the options tables of production sites on a case-by-case basis.

1. With this branch checked on, go to any wp-admin page. Observe a new database table is created named `{prefix}_newspack_campaigns_transients`. Observe all existing campaign transients are copied to the new table. 
2. Verify prompt display logic works normally, respecting frequency settings, permanent dismissal etc.
3. Verify new transients are created in the custom table, not options.
4. Try deleting the table and switching back to `master`. Dismiss a prompt, then switch to this branch. Verify the correct behavior occurs before and after the switch, demonstrating that the transients were seamlessly migrated to the new table.

### Other information:

### DIFF
--- a/api/classes/class-lightweight-api.php
+++ b/api/classes/class-lightweight-api.php
@@ -164,9 +164,8 @@ class Lightweight_API {
 		$table_name       = Segmentation::get_transients_table_name();
 		$name             = '_transient_' . $name;
 		$serialized_value = maybe_serialize( $value );
-		$autoload         = 'no';
 		wp_cache_set( $name, $serialized_value, 'newspack-popups' );
-		$result           = $wpdb->query( $wpdb->prepare( "INSERT INTO `$table_name` (`option_name`, `option_value`, `autoload`) VALUES (%s, %s, %s) ON DUPLICATE KEY UPDATE `option_name` = VALUES(`option_name`), `option_value` = VALUES(`option_value`), `autoload` = VALUES(`autoload`)", $name, $serialized_value, $autoload ) ); // phpcs:ignore
+		$result           = $wpdb->query( $wpdb->prepare( "INSERT INTO `$table_name` (`option_name`, `option_value`) VALUES (%s, %s) ON DUPLICATE KEY UPDATE `option_name` = VALUES(`option_name`), `option_value` = VALUES(`option_value`)", $name, $serialized_value ) ); // phpcs:ignore
 
 		$this->debug['write_query_count'] += 1;
 	}
--- a/api/classes/class-lightweight-api.php
+++ b/api/classes/class-lightweight-api.php
@@ -140,7 +140,7 @@ class Lightweight_API {
 		} elseif ( false === $value ) {
 			$this->debug['read_query_count'] += 1;
 
-			$value = $wpdb->get_var( $wpdb->prepare( "SELECT option_value FROM `$table_name` WHERE option_name = %s LIMIT 1", $name ) ); // phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching;
+			$value = $wpdb->get_var( $wpdb->prepare( "SELECT option_value FROM `$table_name` WHERE option_name = %s LIMIT 1", $name ) ); // phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching, WordPress.DB.PreparedSQL.InterpolatedNotPrepared;
 			if ( $value ) {
 				wp_cache_set( $name, $value, 'newspack-popups' );
 			} else {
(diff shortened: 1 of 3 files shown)


dkoo (approved):
This works as described, so approving.

Now this may or may not be relevant, but two things I keep reading about with respect to performance issues and `wp_options` are:

1. The use of `autoload=true` on most or all rows in the `wp_options` table
2. The lack of indexing in the `wp_options` table beyond the option name

Example article that talks about this a bit: https://kinsta.com/knowledgebase/wp-options-autoloaded-data/

Since this new table is more or less a copy of `wp_options` in terms of columns and indexing, would it perhaps have the same performance issues? For example, I'm not sure whether the `autoload` value is significant at all to WP since this is now a custom table. But I want to make sure that if we're moving the transients to a custom table, we're not just going to experience the same performance issues that `wp_options` would present. And furthermore, since this PR leaves the transients data intact in the `wp_options` table, it might actually mean the performance gets worse if that's the case, right?.

jeffersonrabb:
> 1. The use of `autoload=true` on most or all rows in the `wp_options` table
True, the autoload column has no meaning in the custom table. I'll remove.

> The lack of indexing in the wp_options table beyond the option name
If I'm not mistaken, every read/write to this custom table will have a `WHERE` clause on `option_name`, so that should be the only index that's meaningful.

> Since this new table is more or less a copy of wp_options in terms of columns and indexing, would it perhaps have the same performance issues?
The main goal is to get the campaigns transients out of the "normal" options table, so that the many normal options queries performed by WordPress aren't bogged down by the campaigns transients. Even if the custom table still performs poorly, having a lean options table should pay big dividends unrelated to Campaigns functionality.

> And furthermore, since this PR leaves the transients data intact in the wp_options table, it might actually mean the performance gets worse if that's the case, right?.
I'm only leaving the data intact out of initial caution. Once this change is deployed and verified to be working well, we can manually remove the legacy transients from the options table. As a future enhancement we could remove these programmatically, but since this is a one-time operation we might as well just manually apply it to the production sites and be done with it.

dkoo (approved):
Thanks for removing the `autoload` column and for responding to my concerns. I think this is a good change overall, and we can handle further optimizations in other PRs. :shipit:

api/classes/class-lightweight-api.php
```diff
@@ -130,16 +130,17 @@ public function error( $code ) {
...
+		$value      = wp_cache_get( $name, 'newspack-popups' );
 		if ( -1 === $value ) {
 			$this->debug['read_empty_transients'] += 1;
 			$this->debug['cache_count']           += 1;
 			return null;
 		} elseif ( false === $value ) {
 			$this->debug['read_query_count'] += 1;
-			$value                            = $this->get_option( $name );
+
+			$value = $wpdb->get_var( $wpdb->prepare( 'SELECT option_value FROM %s WHERE option_name = %s LIMIT 1', $table_name, $name ) ); // phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching;
```
adekbadek, line 143: Not sure if that's maybe related to my local setup, but I was getting a ` WordPress database error You have an error in your SQL syntax` error because of bad escaping of table name. Here's what fixed it:

```suggestion
			$value = $wpdb->get_var( $wpdb->prepare( "SELECT option_value FROM `$table_name` WHERE option_name = %s LIMIT 1", $name ) ); // phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching;
```

Didn't you encounter the same issue?

adekbadek (changes requested):

api/classes/class-lightweight-api.php
dkoo: Strange, I didn't encounter this error when testing this PR.

adekbadek: FWIW I'm running PHP `7.4.1` & MySQL `8.0.16`

jeffersonrabb: Me neither, but the fix looks good: https://github.com/Automattic/newspack-popups/pull/488/commits/2b7ede234bdf9068aec71e4c42230122c911a0a5

dkoo: Also running PHP 7.4.1 and MySQL 8.0.16

adekbadek: > Me neither, but the fix looks good: 2b7ede2

Almost! PHP needs double quotes to interpolate a string.

adekbadek: Fixed in 5f52d6a0225c106bdc12b0b60563afdb09138902

adekbadek (approved):

PR #489, opened 2021-03-15
fix: introduce blacklisted blocks to ignore when inserting prompts

### LINKED ISSUE #465: Incorrect prompt insertion when post content contains jetpack/slideshow blocks

dkoo:
### All Submissions:

#466 and #468 tweaked the inline prompt insertion logic to account for slideshows, but it wasn't enough to truly fix the insertion logic in posts with slideshows. This PR introduces the concept of ignoring certain types of blocks for purposes of inline prompt insertion; these blocks may contain a massive amount of block content but will only take up a small amount of visual space in the post content, so their content length shouldn't be used for purposes of inline prompt insertion.

The new feature lets us designate any block to be ignored by adding it to the `$blacklisted_blocks` array in `class-newspack-popups-inserter.php`. At time of PR, only `jetpack/slideshow` and `newspack-blocks/carousel` are designated as such, but more can be added if they're found to cause similar issues with inline prompt insertion.

Closes #465 (again).

1. Create a post with 10 very short paragraphs, perhaps a sentence or two each.
2. At the top of the post, insert a Jetpack slideshow block with a truly massive number of slides (at least 20, preferably with captions, the more the better to inflate the size of its `innerHTML`).
3. Create and publish several inline prompts at various insertion points try (0%, 25%, 50%, 75%, and 100%).
4. On `master`, view the post while not logged in and observe that several of the inline prompts appear bunched together toward the top of the post.
5. Check out this branch and refresh the post. Now confirm that the prompts appear more evenly spaced throughout the short paragraphs.
6. Edit the post and replace the slideshow with a Newspack posts carousel block. Make sure the block is set to a large number of posts (20+).
7. Repeat steps 4-5 and confirm the same issue on `master`, and a similar result on this branch.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -221,12 +221,18 @@ final class Newspack_Popups_Inserter {
 			$content = scaip_maybe_insert_shortcode( $content );
 		}
 
-		// Dynamic blocks might render an arbitrary amount of content, so the length of the content string is
-		// not an accurate representation of the content length.
-		$total_length = 0;
-		foreach ( parse_blocks( $content ) as $block ) {
-			$block_content = $block['innerHTML'];
-			$total_length += strlen( wp_strip_all_tags( $block_content ) );
+		// For certain types of blocks, their innerHTML is not a good representation of the length of their content.
+		// For example, slideshows may have an arbitrary amount of slide content, but only show one slide at a time.
+		// Don't insert prompts adjacent to these.
+		$override_blocks = [ 'jetpack/slideshow', 'newspack-blocks/carousel' ];
+		$parsed_blocks   = parse_blocks( $content );
+		$total_length    = 0;
+
+		foreach ( $parsed_blocks as $block ) {
+			if ( ! in_array( $block['blockName'], $override_blocks ) ) {
+				$block_content = $block['innerHTML'];
+				$total_length += strlen( wp_strip_all_tags( $block_content ) );
+			}
 		}
 
 		// 1. Separate prompts into inline and overlay.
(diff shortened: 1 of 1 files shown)


adekbadek (commented):
This causes the blacklisted blocks to be simply disregarded - this way if a post starts with one of these, and there's an inline prompt at 0%, it will be placed after the blacklisted block. 
Perhaps instead of ignoring the blacklisted blocks in computation, they can be given a fake length? This way a prompt at 0% would render always before any content.

dkoo:
> This causes the blacklisted blocks to be simply disregarded - this way if a post starts with one of these, and there's an inline prompt at 0%, it will be placed after the blacklisted block. Perhaps instead of ignoring the blacklisted blocks in computation, they can be given a fake length? This way a prompt at 0% would render always before any content.

This is true (and I like your term "blacklisted" better than override...was having trouble thinking of the word I wanted yesterday!). 2f66203 gives blacklisted blocks a length of `1` so that prompts at 0% will still appear before them, even when they're at the top of the post. The very small length means that they still have little to no effect on insertion logic after that.

To test:

1. Ensure that the first block in the post is a blacklisted block (slideshow or carousel) with lots of slides
2. Confirm that an inline prompt with 0% placement appears before the block
3. Also confirm that a prompt at 100% will appear after the post content, even if the last block is a blacklisted block.

includes/class-newspack-popups-inserter.php
```diff
@@ -221,12 +221,21 @@ function ( $popup ) use ( $shortcoded_popups_ids ) {
...
 
-		// Dynamic blocks might render an arbitrary amount of content, so the length of the content string is
-		// not an accurate representation of the content length.
-		$total_length = 0;
-		foreach ( parse_blocks( $content ) as $block ) {
-			$block_content = $block['innerHTML'];
-			$total_length += strlen( wp_strip_all_tags( $block_content ) );
+		// For certain types of blocks, their innerHTML is not a good representation of the length of their content.
+		// For example, slideshows may have an arbitrary amount of slide content, but only show one slide at a time.
+		// Don't insert prompts adjacent to these.
```
adekbadek, line 226: Nit: this line in the comment is not accurate after the recent changes ;)

adekbadek (approved):

includes/class-newspack-popups-inserter.php
dkoo: Updated in f479c4d 😄

PR #490, opened 2021-03-16
refactor: minimize transients written to one per reader

dkoo:
### All Submissions:

Taking inspiration from #478, makes two changes to the way the plugin handles writing transients:

a. Transients are not written for unrecognized readers who do not have any ~read-posts~ recorded history (this should avoid writing new empty client data rows for bounced readers), and
b. Writes a maximum of one row per client, with all prompt-related data collected in that row.

Does not do anything to add expiration logic to transients (let's save that for another PR to keep this testable).

Since the transient name and data format differ slightly, this PR also handles the legacy format. If we can't find client or campaign data using the new transient name, we look for data using the legacy name. Then, if legacy data is found, we save that data with the new name and format and then delete the legacy rows. This is not a replacement for a full migration strategy (for that we'll want to programmatically migrate all legacy data to the new format at once), but it will help the plugin fall back gracefully to legacy data while also cleaning up after itself.

May need some light refactoring after #488 lands.

1. On `master`, publish several prompts to varying segments.
2. In a new incognito session, visit the homepage without visiting any articles.
3. Inspect the `wp_options` table and observe that a new row is created for the new reader with a name like `_transient_amp-wRGBrPpeCBfl_JSik4xLQw-popups` (note the `-popups` suffix), despite having no reader activity yet.
4. View an article and dismiss a prompt or two.
5. Inspect the `wp_options` table again and observe that new rows are created for each dismissed prompt to log the dismissals. Note the option names with the `-id_####-popup` suffix.
6. Check out this branch. In the same incognito session, refresh the article or view another article. Confirm that the prompts you previously dismissed remain dismissed (so the changes in this PR won't have any noticeable effect to front-end readers).
7. Inspect the `wp_options` table. Confirm the following:
  a. There is a new row with a name like `_transient_amp-wRGBrPpeCBfl_JSik4xLQw` (note no suffix). If you inspect the row value, it should have a `prompts` key containing data for each viewed and suppressed prompt.
  b. The previous rows for the reader (with `-popups` suffix) and each prompt (with `-id_####-popup` suffix) are deleted.
8. Close out the session and open a new incognito session. View the homepage without visiting any articles.
9. Inspect the `wp_options` table and confirm that no new row is created, since the new reader has no activity yet.
10. Read an article and inspect the `wp_options` table again—a new row should be created for the reader with posts read activity.

### Other information:

### DIFF
--- a/api/classes/class-lightweight-api.php
+++ b/api/classes/class-lightweight-api.php
@@ -318,7 +318,7 @@ class Lightweight_API {
 	 * @param string $client_data_update Client data.
 	 */
 	public function save_client_data( $client_id, $client_data_update ) {
-		$existing_client_data                       = $this->get_client_data( $client_id, true );
+		$existing_client_data = $this->get_client_data( $client_id, true );
 
 		// Update prompts data: new data should replace existing data.
 		if ( isset( $client_data_update['prompts'] ) && ! empty( $existing_client_data['prompts'] ) ) {
--- a/api/classes/class-lightweight-api.php
+++ b/api/classes/class-lightweight-api.php
@@ -198,14 +198,7 @@ class Lightweight_API {
 	 * @return object Prompt data.
 	 */
 	public function get_campaign_data_legacy( $client_id, $campaign_id ) {
-		$data = $this->get_transient( $this->get_transient_name_legacy( $client_id, $campaign_id ) );
-		return [
-			'count'            => ! empty( $data['count'] ) ? (int) $data['count'] : 0,
-			'last_viewed'      => ! empty( $data['last_viewed'] ) ? (int) $data['last_viewed'] : 0,
-			// Primarily caused by permanent dismissal, but also by email signup
-			// (on a newsletter prompt) or a UTM param suppression.
-			'suppress_forever' => ! empty( $data['suppress_forever'] ) ? (int) $data['suppress_forever'] : false,
-		];
+		return $this->get_transient( $this->get_transient_name_legacy( $client_id, $campaign_id ) );
 	}
 
 	/**
@@ -292,21 +285,18 @@ class Lightweight_API {
 
 		$this->save_client_data(
 			$client_id,
-			array_merge(
-				$this->client_data_blueprint,
-				[
-					'posts_read' => array_map(
-						function ( $item ) {
-							return [
-								'post_id'      => $item->post_id,
-								'category_ids' => $item->category_ids,
-								'created_at'   => $item->created_at,
-							];
-						},
-						$client_post_read_events
-					),
-				]
-			)
+			[
+				'posts_read' => array_map(
+					function ( $item ) {
+						return [
+							'post_id'      => $item->post_id,
+							'category_ids' => $item->category_ids,
+							'created_at'   => $item->created_at,
+						];
+					},
+					$client_post_read_events
+				),
+			]
 		);
 
 		return $this->get_transient( $client_id );
--- a/api/classes/class-lightweight-api.php
+++ b/api/classes/class-lightweight-api.php
@@ -36,6 +36,7 @@ class Lightweight_API {
 		'donations'                      => [],
 		'email_subscriptions'            => [],
 		'user_id'                        => false,
+		'prompts'                        => [],
 	];
 
 	/**
@@ -50,6 +51,7 @@ class Lightweight_API {
 		$this->debug = [
 			'read_query_count'       => 0,
 			'write_query_count'      => 0,
+			'delete_query_count'     => 0,
 			'cache_count'            => 0,
 			'read_empty_transients'  => 0,
 			'write_empty_transients' => 0,
@@ -77,12 +79,13 @@ class Lightweight_API {
 	}
 
 	/**
+	 * LEGACY FORMAT - one transient per reader per prompt.
 	 * Get transient name.
 	 *
 	 * @param string $client_id Client ID.
 	 * @param string $popup_id Popup ID.
 	 */
-	public function get_transient_name( $client_id, $popup_id = null ) {
+	public function get_transient_name_legacy( $client_id, $popup_id = null ) {
 		if ( null === $popup_id ) {
 			// For data about popups in general.
 			return sprintf( '%s-popups', $client_id );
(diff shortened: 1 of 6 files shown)


api/classes/class-lightweight-api.php
```diff
@@ -224,23 +285,31 @@ public function get_client_data( $client_id, $do_not_rebuild = false ) {
...
-						return [
-							'post_id'      => $item->post_id,
-							'category_ids' => $item->category_ids,
-							'created_at'   => $item->created_at,
-						];
-					},
-					$client_post_read_events
-				),
-			]
+			array_merge(
```
adekbadek, line 295: The merge is not necessary here, the blueprint will be merged in during the save here:

https://github.com/Automattic/newspack-popups/blob/a18ae51fc405f3f6a1b42207510330f6afde6a51/api/classes/class-lightweight-api.php#L345

& 

https://github.com/Automattic/newspack-popups/blob/a18ae51fc405f3f6a1b42207510330f6afde6a51/api/classes/class-lightweight-api.php#L290

I've added a test to demonstrate an issue with merging it twice:

https://github.com/Automattic/newspack-popups/blob/a18ae51fc405f3f6a1b42207510330f6afde6a51/tests/test-api.php#L595-L598
 
After the merging of blueprint is removed, the test will pass.

api/classes/class-lightweight-api.php
```diff
@@ -224,23 +285,31 @@ public function get_client_data( $client_id, $do_not_rebuild = false ) {
 			$wpdb->prepare( "SELECT * FROM $events_table_name WHERE client_id = %s AND type = 'post_read'", $client_id ) // phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared
 		);
 
+		// If client hasn't read any posts yet, don't write any data.
```
adekbadek, line 288: This threw me off a bit, since the posts read are not the only data stored - it's possible to have a client with no posts read, but with a donation made. _But_ donations are not stored in the custom events table, and so do not matter here. I suggest a change in the comment:

```suggestion
		// If there is no relevant client data in the events table, do not save any data.
```

api/classes/class-lightweight-api.php
```diff
@@ -170,14 +172,33 @@ public function set_transient( $name, $value ) {
...
 	 *
 	 * @param string $client_id Client ID.
 	 * @param string $campaign_id Prompt ID.
 	 * @return object Prompt data.
 	 */
-	public function get_campaign_data( $client_id, $campaign_id ) {
-		$data = $this->get_transient( $this->get_transient_name( $client_id, $campaign_id ) );
+	public function get_campaign_data_legacy( $client_id, $campaign_id ) {
+		$data = $this->get_transient( $this->get_transient_name_legacy( $client_id, $campaign_id ) );
 		return [
```
adekbadek, line 202: No need to process the return value here - it's processed again by `get_campaign_data`. Just returning `$data` would suffice.

adekbadek (changes requested):
Looks good! Added & edited some tests, mostly as a side effect of trying to understand it all :D

tests/test-api.php
```diff
@@ -584,11 +586,155 @@ public function test_client_data() {
...
+			'Returns data with prompt data after a prompt is reported.'
+		);
+		self::assertEquals(
+			$api->get_campaign_data( $client_id, $popup_id ),
+			$expected_popup_data,
+			'Returns prompt data only.'
+		);
+
+		// Report another view of the same prompt.
+		self::$report_campaign_data->report_campaign(
```
adekbadek, line 659: Note: changed use of `save_client_data` (more of an internal function) to `report_campaign`/`report_client_data`, so the tests are more aligned with how the code operates.

dkoo: Makes sense, thanks for clarifying!

api/classes/class-lightweight-api.php
dkoo: You're right, thanks for the new test and for explaining why the merge isn't needed! Updated in dfc3b19.

api/classes/class-lightweight-api.php
dkoo: Good call, removed in dfc3b19.

api/classes/class-lightweight-api.php
```diff
@@ -195,7 +245,10 @@ public function get_campaign_data( $client_id, $campaign_id ) {
 	 * @param string $campaign_data Prompt data.
 	 */
 	public function save_campaign_data( $client_id, $campaign_id, $campaign_data ) {
-		return $this->set_transient( $this->get_transient_name( $client_id, $campaign_id ), $campaign_data );
+		$client_data                            = $this->get_client_data( $client_id, true );
+		$client_data['prompts'][ $campaign_id ] = $campaign_data;
+
+		return $this->set_transient( $client_id, $client_data );
```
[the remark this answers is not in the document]
dkoo: Tried this in d9bf5fd. Note that in the `save_client_data` function, I had to replace `array_merge_recursive` with `array_replace_recursive` because the former was giving me a result that merged single array values into arrays, e.g. 

```
'count' => [ 0, 1, 2 ]
```

[`array_replace_recursive`](https://www.php.net/manual/en/function.array-replace-recursive.php) seems closer to the behavior of `array_merge` anyway, right?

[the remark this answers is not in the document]
dkoo: 0b4fc2b fixes the failing tests—`array_merge_recursive` is still needed for the non-prompt client data, since we don't actually want to replace `posts_read` but append to it, for example.

dkoo:
@jeffersonrabb this could use your review, since Adam contributed code too.

adekbadek (approved):
Awaiting @jeffersonrabb's review, since I have contributed.

api/classes/class-lightweight-api.php
```diff
@@ -171,32 +172,62 @@ public function set_transient( $name, $value ) {
...
+	 *
+	 * @param string $name The transient's name.
+	 */
+	public function delete_transient( $name ) {
+		global $wpdb;
+		$name = '_transient_' . $name;
+
+		wp_cache_delete( $name );
+		$wpdb->delete( // phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery
+			$wpdb->options,
```
jeffersonrabb, line 185: This deletes the transient from the options table, but as of https://github.com/Automattic/newspack-popups/pull/488 these are in the custom table.

jeffersonrabb (changes requested):

jeffersonrabb (approved):
Great improvements!

PR #509, opened 2021-04-02
API: ignore crawler traffic

adekbadek:
### All Submissions:

Ignores requests made by commonly known web crawlers in the API.

1. Verify the API works for regular requests made from a web browser
2. Set up Google Search Console w/ the site (e.g. by setting up Site Kit plugin)
2. Take note of the row count in the transients and events table*
3. Submit a URL of a post (on which prompts are displayed) to the URL inspection in Search Console UI
4. Wait for it to be crawled by Google** 
5. Verify that the row counts in the tables did not increase

\* `$ wp db query "select count(*) from wp_newspack_campaigns_transients"`, `$ wp db query "select count(*) from wp_newspack_campaigns_events"`
\** In my testing I've seen that Google hit my site after a few minutes (by means of an `error_log`ing the user agent in API), but it did not show up in the Search Console UI 🤷‍♀️

### Other information:

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -24,6 +24,7 @@ require_once $_tests_dir . '/includes/functions.php';
  */
 function _manually_load_plugin() {
 	$_SERVER['HTTP_REFERER'] = 'https://' . $_SERVER['HTTP_HOST']; // phpcs:ignore
+	$_SERVER['HTTP_USER_AGENT'] = 'Mozilla\/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit\/537.36 (KHTML, like Gecko) Chrome\/89.0.4389.90 Safari\/537.36'; // phpcs:ignore
 
 	require dirname( dirname( __FILE__ ) ) . '/newspack-popups.php';
 	require dirname( dirname( __FILE__ ) ) . '/src/blocks/custom-placement/view.php';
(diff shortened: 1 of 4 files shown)


tests/bootstrap.php
```diff
@@ -24,6 +24,7 @@
  */
 function _manually_load_plugin() {
 	$_SERVER['HTTP_REFERER'] = 'https://' . $_SERVER['HTTP_HOST']; // phpcs:ignore
+	$_SERVER['HTTP_USER_AGENT'] = 'Mozilla\/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit\/537.36 (KHTML, like Gecko) Chrome\/89.0.4389.90 Safari\/537.36'; // phpcs:ignore
```
dkoo, line 27: Do we want to consider a unit test for this? Maybe a control test with this spoofed user agent, and then another test with one of the crawler user agents to ensure that the data isn't written? Not sure if necessary.

dkoo (approved):
This works as described. 👍 Added a question below but feel free to ignore if you think it's not necessary.

One tip to take some of the guesswork out of testing: you can click the "Test Live URL" button after inspecting a URL to trigger the crawler immediately (it will show a loading state until the URL has been crawled).

tests/bootstrap.php
adekbadek: That would require a more end-to-end test setup, since we're just testing the API handlers in isolation now. But I've added a basic test for `is_a_web_crawler` method in fd2c1ef2b712d99fd744a1546bb8b1a80712939b

PR #510, opened 2021-04-05
Retrieve widget popups ids from options

### LINKED ISSUE #470: Error when editing Text Widget in the Customizer
**Steps to reproduce**

1. Start on a site with the Campaigns plugin activated
2. Navigate to Customize > Widgets and add a text widget, and write some text in it
3. Note the error it throws:

image

4. Try to save your changes, but it won't actually save.
5. Deactivate the Campaigns plugin and repeat step 2 -4 -- it should now work. 

I'm not sure exactly what's happening here, but commenting out the following code for me made the error go away: 

https://github.com/Automattic/newspack-popups/blob/32836f12b5e498898125c3d320c6cfcf183165c9/includes/class-newspack-popups-inserter.php#L115-L120

(What it broke in the plugin is another story).

Other widgets seem to save fine in the Customizer, and if you navigate to WP Admin > Appearance > Widgets, you can edit Text Widgets there as a work-around.

adekbadek:
### All Submissions:

Since Text Widget's options are store in `text_widget` option, retrieving prompts displayed in the widgets can be simplified.
Closes #470.

1. Add prompts in widgets
2. Visit a page on which these prompts would be displayed only in widget areas (e.g. category archive)
3. Verify the prompts are displayed

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -87,14 +87,6 @@ final class Newspack_Popups_Inserter {
 		add_action( 'wp_enqueue_scripts', [ $this, 'enqueue_scripts' ] );
 
 		add_filter(
-			'widget_update_callback',
-			[ $this, 'save_widgets_shortcoded_popup_ids' ],
-			1,
-			4
-		);
-		add_action( 'delete_widget', [ $this, 'remove_widgets_shortcoded_popup_ids' ] );
-
-		add_filter(
 			'newspack_popups_assess_has_disabled_popups',
 			function ( $disabled ) {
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Works great and much simpler than the previous code. Thanks @philipjohn for the idea!

Testing this made me question the behavior of in-widget prompts when rendered on pages where they might also appear inline, though—if you place an inline prompt in a widget, then view an article that also shows the widget, the prompt will appear both in the article content and the widget. Is this the desired behavior, or should the presence of the shortcoded prompt in the widget "override" the inline instance, to avoid the duplication?

adekbadek:
Good question. Maybe we should revive the "manual only" placement option? But I think this is an edge case, though.

philipjohn:
Nice work, thanks both!

PR #505, opened 2021-03-31
Small shortcode, amp-access fixes

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Fixes two small bugs, refactors and adds tests.

### How to test the changes in this Pull Request:

1. On `master`,
2. Create two inline prompts at 0% scroll, "everyone" segment, then insert one of them using a shortcode in a post, right at the top.
3. Open Campaigns Wizard, preview the non-shortcoded prompt 
4. Issue 1. - Observe that the previewed prompt is rendered twice - in the shortcode spot and according to the inline placement rules
5. Load a page as non-admin, where some inline prompts, but none custom-placement prompts are rendered 
6. Issue 2. - Observe the payload in `GET /newspack-popups/api/campaigns/index.php` contains duplicated items in `popups` query var
7. Switch to this branch, observe both issues gone

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #513, opened 2021-04-07
feat: new option to show border for inline prompts

### LINKED ISSUE #483: Make prompts respond to flush left or right alignment. Maybe full width.
Enhancement (I can't label issues the normal way)

Goal: Make prompts respond to alignment control when contents are grouped. 

A groups can usually be aligned left or right, making it half the column width, with type wrapping. They can also be full or wide width. 

Screen Shot 2021-03-12 at 1 24 47 PM

I tried this with prompts, and the left aligned looked promising, with some stray lines at the top. Right didn't do anything. Setting it at wide width made it appear on the right, still with the stray lines. So it's rather unpredictable now.

dkoo:
### All Submissions:

Adds a new option for inline and custom placement prompts to display the border wrapping the prompt content on the front-end. This is ON by default, so it won't affect existing prompts.

Closes #483 in a roundabout way. This way, if the prompt contains only a Group block with left or right alignment, turning off the border will allow it to float with the article content in a less obtrusive way:

Screen Shot 2021-04-06 at 6 02 54 PM

Full-width and wide alignment will still have no visible effect (more or less the same as no alignment), to avoid collision with sidebar elements and other injected elements.

1. On `master`, create some inline and custom placement prompts.
2. Check out this branch and run `npm run build`.
3. Edit the existing prompts and confirm that a new option appears under Prompt Settings called "Display Prompt Border", which should be enabled by default for existing prompts:

Screen Shot 2021-04-06 at 5 58 42 PM

4. Confirm that the option only appears for which are inline or custom placement (not overlay or above-header). Don't update yet because we want to test the default.
5. View the inline and custom placement prompts in an incognito session. Confirm that they appear with a border, as usual.
6. Turn OFF the "Display Prompt Border" option and update. Confirm in the incognito session that these prompts now render without a border or container padding.
7. Create a new prompt and confirm that the "Display Prompt Border" option is ON by default.
8. Publish the prompt and test it on the front-end with the option both on and off.

### Other information:


includes/class-newspack-popups-model.php
```diff
@@ -344,10 +346,11 @@ public static function create_popup_object( $campaign_post, $include_taxonomies
 			'title'   => $campaign_post->post_title,
 			'content' => $campaign_post->post_content,
 			'options' => wp_parse_args(
-				array_filter( $post_options ),
```
adekbadek, line 347: Why is the filtering removed? I think this served the purpose of allowing `wp_parse_args` to overwrite empty options.

adekbadek (changes requested):
The param is not handled in the preview triggered from the editor.

image

includes/class-newspack-popups-model.php
dkoo: Because the `display_border` option is `true` by default, if you then toggle off the option the filtering overrides it, making it impossible to actually turn it off.

dkoo: At first I created the option as "Hide prompt border" with a default value of `false`, but it seemed weird to have all the other options phrased in the positive ("Display prompt title") with that option in the negative.

dkoo:
> The param is not handled in the preview triggered from the editor.

It looks like several params weren't handled in the preview triggered from the editor (such as `display_title`). c8189dc should fix this.

includes/class-newspack-popups-model.php
adekbadek: The only change I've spotted is that without filtering `$post_options`, `dismiss_text_alignment`'s default value is not applied by `wp_parse_args`. This is not a problem though, because it defaults to 'center' anyway here:

https://github.com/Automattic/newspack-popups/blob/c8189dc9c6376fa89509bdf98b75e12f60c6f27f/includes/class-newspack-popups-model.php#L417-L419

However, this makes the `wp_parse_args` call misleading (not all defaults will be applied). This is not a big issue, but it'd be worth it to centralise the default values setting to `create_popup_object` method (without the need for stuff like `get_dismiss_text_alignment` or `get_dismiss_text`).

includes/class-newspack-popups-model.php
```diff
@@ -853,6 +859,11 @@ class="<?php echo esc_attr( implode( ' ', $classes ) ); ?>"
 	 * @return string The generated markup.
 	 */
 	public static function generate_popup( $popup ) {
+		// If previewing a single prompt, override saved settings with preview settings.
+		if ( Newspack_Popups::previewed_popup_id() ) {
+			$popup = self::retrieve_preview_popup( Newspack_Popups::previewed_popup_id() );
+		}
```
adekbadek, line 865: Why is this needed? In a preview situation, the argument to this method is already the preview popup:

https://github.com/Automattic/newspack-popups/blob/c8189dc9c6376fa89509bdf98b75e12f60c6f27f/includes/class-newspack-popups-inserter.php#L52-L55

dkoo: This is needed because for regular inline prompts, they're inserted into post content with [this `the_content` hook callback](https://github.com/Automattic/newspack-popups/blob/c8189dc9c6376fa89509bdf98b75e12f60c6f27f/includes/class-newspack-popups-inserter.php#L147). In this callback, they're created as shortcodes, which means they're generated by [this method](https://github.com/Automattic/newspack-popups/blob/c8189dc9c6376fa89509bdf98b75e12f60c6f27f/includes/class-newspack-popups-inserter.php#L345), which doesn't use the `popups_for_post` method where the preview popup is generated.

includes/class-newspack-popups-model.php
dkoo: I see what you mean how it's confusing that default values are set in multiple places. What about standardizing defining the defaults when the meta fields are registered? Tried it in 5f17039 and I think this lets us remove the `wp_parse_args` entirely as well as the getter methods such as `get_dismiss_text` and `get_dismiss_text_alignment` which previously defined defaults.

dkoo: Oh, I see what you mean about the defaults not being set, because the `retrieve_preview_popup` sets all of the options even if corresponding params aren't set on the URL. Let me revisit this.

dkoo: After struggling with this for a bit, I think it's not worth the extensive refactoring we'd need to ensure that a non-default `false` value is preserved over a default `true`. In ce04bce I restored the `array_filter` and flipped the `display_border` option back to `hide_border` with a default of `false`. But I think we can still remove the `get_dismiss_text` and `get_dismiss_text_alignment` methods since the defaults are already set with `wp_parse_args`.

adekbadek: Yeah, that makes sense 👌

includes/class-newspack-popups-model.php
```diff
@@ -745,13 +728,10 @@ public static function get_access_attrs( $popup ) {
 	 * @param object $popup Popup.
 	 */
 	private static function render_permanent_dismissal_form( $element_id, $popup ) {
-		$dismiss_text = self::get_dismiss_text( $popup );
-		if ( ! $dismiss_text ) {
-			return '';
-		}
```
adekbadek, line 751: Still, this should return an empty string if the dismiss text is empty - otherwise a form element will be rendered with no content.

includes/class-newspack-popups-model.php
adekbadek: Ok, now I get it - this was not needed before (before #505) because in a single prompt preview situation _every_ shortcode was displaying the previewed prompt. This was [changed in #505](https://github.com/Automattic/newspack-popups/pull/505/files#diff-84a23fd0b9ee98b91687d95435cc335c32a0f5c6cb217816816fe330ae72c4f3L354-R355), but as we see here it did introduce a regression, which the code here in the PR fixes.

adekbadek (changes requested):

includes/class-newspack-popups-model.php
dkoo: Ah, good point. Re-added in 08947e8.

adekbadek (approved):

PR #506, opened 2021-03-31
Fix tag filtering

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Fixes a bug

### How to test the changes in this Pull Request:

1. On `master`, create a prompt and set a tag on it
2. Visit a post with no tags, observe that the prompt is displayed anyway
3. Switch to this branch, observe the prompt is not displayed

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #515, opened 2021-04-08
Fix tag criterion

adekbadek:
### All Submissions:

Fixes a regression introduced in #506 

1. On `master`,
2. Create a post with a tag and an inline prompt with no tags assigned
3. Observe the prompt is not displayed on the post
4. Switch to this branch, observe the prompt is displayed
5. Assign a different tag to the prompt, observe it's now not displayed (since tags don't match)
6. Assign the post's tag to the prompt, observe the prompt is displayed

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -660,21 +660,15 @@ final class Newspack_Popups_Inserter {
 	 * @return bool Should popup be shown based on tags it has.
 	 */
 	public static function assess_tags_filter( $popup ) {
-		$post_tags = get_the_tags();
-		if ( false === $post_tags ) {
-			$post_tags = [];
-		}
 		$popup_tags = get_the_tags( $popup['id'] );
 		if ( false === $popup_tags ) {
-			$popup_tags = [];
+			return true; // No tags on the popup, no need to compare.
 		}
-		if ( count( $post_tags ) || count( $popup_tags ) ) {
-			return array_intersect(
-				array_column( $post_tags, 'term_id' ),
-				array_column( $popup_tags, 'term_id' )
-			);
-		}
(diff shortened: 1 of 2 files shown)


includes/class-newspack-popups-inserter.php
```diff
@@ -660,21 +660,15 @@ public static function assess_categories_filter( $popup ) {
...
 		$popup_tags = get_the_tags( $popup['id'] );
 		if ( false === $popup_tags ) {
-			$popup_tags = [];
-		}
-		if ( count( $post_tags ) || count( $popup_tags ) ) {
-			return array_intersect(
-				array_column( $post_tags, 'term_id' ),
-				array_column( $popup_tags, 'term_id' )
-			);
+			return true; // No tags on the popup, no need to compare.
```
dkoo, line 665: This works well, but a nit: isn't this more or less the same logic as `assess_categories_filter` right above? It'd be nice to standardize both functions so that it's clearer they do the same thing but with different taxonomies. Perhaps we could even refactor both functions into a single function that takes a taxonomy argument to make it extra clear.

dkoo (changes requested):
Tested and works as described. Just one nit detailed below, let me know what you think.

dkoo (approved):

dkoo:
I'm approving this because it's clear this affects many sites, so I'd rather take care of the tiny refactor in another PR. Merging for a hotfix release.

includes/class-newspack-popups-inserter.php
dkoo: Decided to move this to a [new PR](https://github.com/Automattic/newspack-popups/pull/518) since the fix is somewhat urgent but the refactor certainly isn't.

PR #518, opened 2021-04-09
refactor: combine tags/category filtering methods into one

dkoo:
### All Submissions:

Inspired by https://github.com/Automattic/newspack-popups/pull/515#discussion_r609913318. I felt that because the category and tag filtering should behave the same way, they should share the same logic and method. This PR combines the methods that compare post and prompt categories and tags into a single method that takes a second `$taxonomy` arg (defaulting to category). Just seemed like one tiny thing we could do to reduce the mental load when troubleshooting prompt display logic.

This is a pure refactor, so the behavior shouldn't differ at all from `master`. Test that prompts with and without categories and tags display as expected on posts with and without matching categories and tags.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -653,8 +653,6 @@ final class Newspack_Popups_Inserter {
 			array_column( $post_terms ? $post_terms : [], 'term_id' ),
 			array_column( $popup_terms, 'term_id' )
 		);
-
-		return true;
 	}
 
 	/**
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -644,7 +644,7 @@ final class Newspack_Popups_Inserter {
 	 * @return bool Whether the prompt should be shown based on matching terms.
 	 */
 	public static function assess_taxonomy_filter( $popup, $taxonomy = 'category' ) {
-		$post_terms  = get_the_terms( $post_id, $taxonomy );
+		$post_terms  = get_the_terms( get_the_ID(), $taxonomy );
 		$popup_terms = get_the_terms( $popup['id'], $taxonomy );
 
 		if ( $post_terms && count( $post_terms ) && $popup_terms && count( $popup_terms ) ) {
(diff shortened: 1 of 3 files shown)


includes/class-newspack-popups-inserter.php
```diff
@@ -635,40 +635,26 @@ public static function assess_is_post( $popup ) {
...
 	 */
-	public static function assess_categories_filter( $popup ) {
-		$post_categories  = get_the_category();
-		$popup_categories = get_the_category( $popup['id'] );
+	public static function assess_taxonomy_filter( $popup, $taxonomy = 'category' ) {
+		$post_terms  = get_the_terms( get_the_ID(), $taxonomy );
+		$popup_terms = get_the_terms( $popup['id'], $taxonomy );
 
-		if ( $post_categories && count( $post_categories ) && $popup_categories && count( $popup_categories ) ) {
+		if ( $post_terms && count( $post_terms ) && $popup_terms && count( $popup_terms ) ) {
```
adekbadek, line 650: This makes the assessment true of only one of the compared items has terms. While this is irrelevant for categories (a post will always have a category), this fails for tags - a post might have no tags, but a prompt might do, and then the latter should not be displayed.

This was caught by a PHP test, but because of changes in #509 the tests were [exiting prematurely](https://app.circleci.com/pipelines/github/Automattic/newspack-popups/1263/workflows/4615f2f7-cabe-4169-8a28-28df4248b8b2/jobs/4621/parallel-runs/0/steps/0-104). That's fixed by 643a5e92aeb248c45b83e2f244d653870c589c99 (also some more tests around category filtering are added there). Now the test is properly [reported as failing on CI](https://app.circleci.com/pipelines/github/Automattic/newspack-popups/1268/workflows/bd67ccb1-ac6c-40f3-9149-f81aa8dce366/jobs/4642).

adekbadek (changes requested):

includes/class-newspack-popups-inserter.php
dkoo: Thanks for fixing the test issue and exposing the flaw in the shared logic. 226136b should fix this by applying the tags-based logic instead—even though it's not really possible that a post won't have a category, the logic will still work for both tags and categories now.

adekbadek (approved):
👏

PR #521, opened 2021-04-12
feat: handle classic content for inline prompt insertion

### LINKED ISSUE #520: Position of inline prompt not respected if post is a single classic block
To reproduce:

1. Create a post with a single, "Classic" block
2. Create an inline prompt to be displayed halfway in the content
3. Observe the prompt is inserted at the top of the content

dkoo:
### All Submissions:

Adds special handling for "Classic Content" blocks when it comes to inline prompt insertion. Instead of treating the entire block as a single block, like other blocks, we now break up the Classic Content string by block-level HTML elements (`</p>`, `</ul>`, `</ol>`, etc.—not [every single one](https://developer.mozilla.org/en-US/docs/Web/HTML/Block-level_elements), but the main ones I think make sense to insert prompts before or after) and insert prompts between those elements. Should work whether the Classic Content block is the only block in a post, or if it's mixed with other blocks.

The insertion approach is inspired by the logic in the Super Cool Ad Inserter plugin, but simplified since we already have handling for short content, lots of prompts, etc. It also looks for various block-level elements, not just paragraphs.

This change should ensure that prompts get inserted in a more logical fashion on imported or legacy content that hasn't been (or can't be) converted to Gutenberg blocks.

Closes #520.

1. Create at least two posts containing Classic Content blocks, each containing a mix of different HTML content. In one post, ensure that the Classic Content block is the only block in the post; in the other, make sure it's interspersed with blocks of other types containing content as well.
2. Also publish several inline prompts at different insertion points—try 25%, 50%, and 75%.
3. On `master`, in an incognito session, observe that the prompts are inserted around the Classic Content blocks as if they're single blocks, regardless of the classic content length—so in posts that only contain Classic Content, all prompts will appear at the top of the post.
4. Check out this branch, refresh the posts. Confirm that the prompts are now inserted within the Classic Content blocks' content at more or less the expected positions, and that the Classic Content rendered on the front-end matches what's in the editor.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -213,8 +213,9 @@ final class Newspack_Popups_Inserter {
 
 		foreach ( $parsed_blocks as $block ) {
 			if ( ! in_array( $block['blockName'], $blacklisted_blocks ) ) {
-				$block_content = $block['innerHTML'];
-				$total_length += strlen( wp_strip_all_tags( $block_content ) );
+				$is_classic_block = null === $block['blockName']; // Classic block doesn't have a block name.
+				$block_content    = $is_classic_block ? force_balance_tags( wpautop( $block['innerHTML'] ) ) : $block['innerHTML'];
+				$total_length    += strlen( wp_strip_all_tags( $block_content ) );
 			} else {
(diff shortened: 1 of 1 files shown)


adekbadek (approved):
Works great 👏

PR #524, opened 2021-04-20
Fix: empty blocks handling

adekbadek:
### All Submissions:

Fixes a small issue.

It looks like the Group or Columns blocks are inserting an empty block somewhere, which gets picked up as the classic block in the insertion logic.

0. On `master`
1. Create a post with the following content: a Group, with Columns, a single Column inside, and a Heading in the Column
2. Ensure some inline prompts are inserted in the post
3. Observe and error (when in WP debug mode) shown on the front-end about `Offset not contained in string`
4. Switch to this branch, observe no error shown

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -213,7 +213,7 @@ final class Newspack_Popups_Inserter {
 
 		foreach ( $parsed_blocks as $block ) {
 			if ( ! in_array( $block['blockName'], $blacklisted_blocks ) ) {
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Maybe my test content isn't right, but I didn't see any warnings or errors on `master` with empty Group and Columns in my post. But this is still a good change that we should implement.

While working on an unrelated feature, I found that Classic blocks _can_ sometimes have a name, and they're called `core/freeform`. So in 30ae2e2 I added a check for that name, too. Since it's such a small change, if it looks good to you I don't think we need to bring in a third code reviewer, but feel free to if you want.

PR #529, opened 2021-04-23
Filter out non-existing categories in segments

### LINKED ISSUE #491: Segmentation - handle removed categories
## To reproduce:

1. Create a segment, set "Category Affinity" to a category
2. Remove that category from WordPress instance
3. Edit the segment, observe that there's nothing in the "Category Affinity" input, but "Has favorite categories" still appears in segment description in segments list
4. Note that any prompts assigned to this category will never be shown

## Proposed solution:

The non-existent categories should be filtered out from segment.

adekbadek:
### All Submissions:

Closes #491.

Observe that #491 is not reproducible 

### Other information:

### DIFF
--- a/includes/class-newspack-popups-segmentation.php
+++ b/includes/class-newspack-popups-segmentation.php
@@ -369,6 +369,27 @@ final class Newspack_Popups_Segmentation {
 			$segments = self::reindex_segments( $segments );
 		}
 
+		// Filter out non-existing categories.
+		$existing_categories_ids = get_categories(
+			[
+				'hide_empty' => false, 
+				'fields'     => 'ids',
+			] 
+		);
+		foreach ( $segments as &$segment ) {
+			if ( ! isset( $segment['configuration']['favorite_categories'] ) ) {
+				continue;
+			}
+			$fav_categories = $segment['configuration']['favorite_categories'];
+			if ( ! empty( $fav_categories ) ) {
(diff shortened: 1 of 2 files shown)


dkoo (changes requested):
In my experience, if the category you delete is the only category applied to a prompt, that prompt will inherit the default ("Uncategorized" on my site) category. Can we check for the default category and remove it in this case?

adekbadek:
That's a different scenario - this PR deals with assigning categories to a segment, not to a prompt.
Removing the default category application was handled in #482 - please open an issue if that does not work as expected.

dkoo:
Got it, let me re-test with that in mind.

dkoo (approved):
This works as expected for segments (created a new issue #530 for what I mentioned above). 

Non-blocking question, though: if a segment's only criteria is the deleted category, after deleting the category the segment effectively becomes "empty", which means it will behave the same as "Everyone" (prompts in the segment will be shown to all readers). 

The segmentation UI already tries to prevent admins from creating empty segments, so this becomes a backdoor way of doing just that. Is this something we should be concerned about or at least notify users about?

adekbadek:
> The segmentation UI already tries to prevent admins from creating empty segments, so this becomes a backdoor way of doing just that. Is this something we should be concerned about or at least notify users about?

I think this is very much an edge case. If a segment targets a specific category, it's probably an important one (and so not prone to accidental removal).

PR #536, opened 2021-04-29
fix: bump query limit up to 100 for eligible prompts

dkoo:
### All Submissions:

Fixes a bug discovered while testing https://github.com/Automattic/newspack-plugin/pull/944, which limits the number of non-shortcoded and non-custom-placement posts checked against `amp-access` to 10. This fix bumps that limit up to 100, which should give sites a lot more headroom.

1. On `master`, publish a lot of prompts to "Everyone" or an easily matched segment—at least ~15 of all types (inline, overlay, above-header) which will be programmatically inserted and should be displayed together. Additionally, publish a couple of shortcoded and custom placement prompts and insert into a post.
2. View the post in a new incognito session. Observe that only 10 of the programmatically inserted prompts are displayed, plus the shortcoded and custom placement prompts.
  - If you inspect the DOM and search for `<amp-layout`, you'll see that all prompts are inserted as expected.
  - If you inspect the Network Requests tab and filter by `api/campaigns`, you'll see that the AMP Access request only contains 10 programmatically inserted prompts + shortcoded prompts + custom placement prompts, meaning only those will be potentially displayed.
3. Check out this branch, view the page again in a new session.
4. This time, observe that all prompts are displayed as expected.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -152,11 +152,11 @@ final class Newspack_Popups_Model {
 	}
 
 	/**
-	 * Retrieve all overlay popups.
+	 * Retrieve all popups eligible to be programmatically inserted (not shortcoded or custom placement).
 	 *
 	 * @param  boolean     $include_unpublished Whether to include unpublished prompts.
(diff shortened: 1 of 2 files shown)


adekbadek (approved):
Good catch!

PR #539, opened 2021-05-03
API debugging

adekbadek:
### All Submissions:

On top of a config variable, exposes the API debugging info via a request param. I don't this this is a security concern, as the debugging information does not contain anything sensitive.
Enables debugging the prompt suppression reasons.
Also, makes a small tweak to `amp-access` payload creation.

1. Visit a page with some prompts, observe that (only) after appending a `newspack-campaigns-debug` parameter to the URL, `debug` key is returned along with the response from `GET /api/campaigns/index.php`
2. Observe that the response contains a `suppression` key, with data about why a particular prompt was suppressed
3. Try out different suppression scenarios, observe them reflected in the response.
4. Insert an overlay prompt via a shortcode _EDIT: and assign it to a category, visit a post of different category to ensure it will not be inserted programatically_ - observe that on `master` its ID is in the payload sent to the aforementioned endpoint, while on this branch it is not. Only inline prompts will be rendered in shortcodes - this just cleans _EDIT: cleans the output_

### Other information:

### DIFF
--- a/api/campaigns/class-maybe-show-campaign.php
+++ b/api/campaigns/class-maybe-show-campaign.php
@@ -107,9 +107,14 @@ class Maybe_Show_Campaign extends Lightweight_API {
 					// If the previous overlay already has a higher priority, only show that one. Otherwise, show this one instead.
 					$response[ $overlay_to_maybe_display->id ] = $overlay_to_maybe_display->id === $higher_priority_item->id;
 					$campaign_should_be_shown                  = $campaign->id === $higher_priority_item->id;
-					$overlay_to_maybe_display                  = $higher_priority_item;
+					if ( false === $campaign_should_be_shown ) {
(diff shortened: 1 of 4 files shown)


dkoo (approved):
This works as described, and I like the additional logging output with suppression info. That all seems to work great.

However, with this step: 

> 4. Insert an overlay prompt via a shortcode - observe that on master its ID is in the payload sent to the aforementioned endpoint, while on this branch it is not. Only inline prompts will be rendered in shortcodes - this just cleans

I am still seeing the overlay markup both in the content and in the API request payload. I suspect it's because the prompt is active and thus a candidate for programmatic display if it matches the reader's segment. Without a "manual placement" option that shows the prompt in a shortcode only, I can't think of a situation where it's actually possible to have an active overlay prompt but not have it appear in the content and the API request, unless it has a non-matching category assigned? Am I missing something?

adekbadek:
> I am still seeing the overlay markup both in the content and in the API request payload.

Forgot to mention one step - I've update the testing steps.

PR #558, opened 2021-06-10
feat: duplicate a prompt

### LINKED ISSUE #554: Feature:  copy prompt
When a user needs to create a similar prompt, they currently have to rely on block patterns or copy-pasting. There should be a way to copy a prompt.

dkoo:
### All Submissions:

Adds functionality to duplicate a prompt. Copies are created as new draft prompts with the same title/content, taxonomy terms, and prompt-related meta values as the source prompt.

Prompts can be duplicated from either the Campaigns dashboard or from a single prompt editor screen.

Closes #554.

1. Ensure Campaigns is on this branch and Newspack Plugin is on https://github.com/Automattic/newspack-plugin/pull/1001.
2. In WP admin, visit **Newspack > Campaigns**. Confirm that in the primary popover menu for any prompt, there's a new menu item to "Duplicate" the prompt:

Screen Shot 2021-06-11 at 11 47 40 AM

3. Click "Duplicate". Confirm that a new draft prompt is created with the title "Copy of <original prompt title". It should have the same segment assignments, categories, tags, and meta values (so if copying an overlay, it should be created as an overlay, etc.). 
  - **Note** that if you're duplicating from the "Active Prompts" filter, the copy won't be visible until you filter by its campaigns (or "Unassigned" if it has none), because it's not active. Should we therefore not show the "Duplicate" option when viewing "Active Prompts"? What do you think, @thomasguillot?
4. View both the original prompt and its copy in the editor and confirm that all content, category and tag terms, selected campaign(s), and prompt-related settings match across both prompts.
5. In the editor for a prompt, confirm there's a new "Duplicate" button next to the "Preview" button:

Screen Shot 2021-06-11 at 11 40 26 AM

6. Click the button and observe that after the request completes, you're automatically redirected to the editor screen for a new draft prompt with all the same content, terms, prompt options, etc. as the original.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-api.php
+++ b/includes/class-newspack-popups-api.php
@@ -192,7 +192,7 @@ final class Newspack_Popups_API {
 
 
 	/**
-	 * Duplicate a Pop-up.
+	 * Duplicate a prompt.
 	 *
 	 * @param WP_REST_Request $request Full details about the request.
 	 * @return WP_REST_Response with complete info to render the Engagement Wizard.
--- a/includes/class-newspack-popups-api.php
+++ b/includes/class-newspack-popups-api.php
@@ -67,6 +67,21 @@ final class Newspack_Popups_API {
 				],
 			]
 		);
+
+		\register_rest_route(
+			'newspack-popups/v1',
+			'/(?P<id>\d+)/duplicate',
+			[
+				'methods'             => \WP_REST_Server::EDITABLE,
+				'callback'            => [ $this, 'api_duplicate_popup' ],
+				'permission_callback' => [ $this, 'permission_callback' ],
+				'args'                => [
+					'id' => [
+						'sanitize_callback' => 'absint',
+					],
+				],
+			]
+		);
 	}
 
 	/**
@@ -173,6 +188,18 @@ final class Newspack_Popups_API {
 		}
 
 		return new \WP_REST_Response( [] );
+	}
+
+
+	/**
+	 * Duplicate a Pop-up.
+	 *
+	 * @param WP_REST_Request $request Full details about the request.
+	 * @return WP_REST_Response with complete info to render the Engagement Wizard.
+	 */
+	public function api_duplicate_popup( $request ) {
+		$response = Newspack_Popups::duplicate_popup( $request['id'] );
+		return rest_ensure_response( $response );
 	}
 }
 $newspack_popups_api = new Newspack_Popups_API();
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -315,16 +315,13 @@ final class Newspack_Popups_Model {
 	}
 
 	/**
-	 * Create the popup object.
+	 * Get options for the given prompt, with defaults.
 	 *
-	 * @param WP_Post $campaign_post The prompt post object.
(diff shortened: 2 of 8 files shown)


thomasguillot:
> Note that if you're duplicating from the "Active Prompts" filter, the copy won't be visible until you filter by its campaigns (or "Unassigned" if it has none), because it's not active. Should we therefore not show the "Duplicate" option when viewing "Active Prompts"? What do you think, @thomasguillot?

I didn't read this when I first tested and I was super confused as to where my duplicated prompt was... so it's certainly confusing.

Here are a few things to consider:

* When we click "Duplicate", should it open a modal so we can have the option to rename the prompt and display the category filtering to enable the user to change these
* Add a notice if the prompt will be "unassigned"
* Once the user clicks on the "Duplicate" button (Primary -- modal) they get a confirmation message with a link to edit it in Gutenberg?
* If they click "Cancel" (Secondary -- modal)  then nothing happens

thomasguillot (approved):
Overall it works as expected.

I just have a few thoughts https://github.com/Automattic/newspack-popups/pull/558#issuecomment-860592769 but feel free to tackle these in a separate PR (that's why I'm approving 😊)

_Note: I've fixed a small colour issue with `is-link` in https://github.com/Automattic/newspack-plugin/pull/1001/commits/330ff85bc284339474a9eca2baafcc2cc81f4618`

dkoo:
@thomasguillot I added your UI improvements to #560 so we can handle it in another PR, like you suggest.

dkoo:
@thomasguillot Just had another thought...I like your modal idea for the UI, but it does involve duplicating some UI from the single prompt editing screen. What if after creating the duplicate post we automatically redirect to the single editor page for the duplicated prompt? That screen would let users edit everything in the familiar WP editor and there'd be no confusion about where to find it.

adekbadek (approved):
Nice! I like that the subsequent duplicate titles are handled 👌

thomasguillot:
> @thomasguillot Just had another thought...I like your modal idea for the UI, but it does involve duplicating some UI from the single prompt editing screen. What if after creating the duplicate post we automatically redirect to the single editor page for the duplicated prompt? That screen would let users edit everything in the familiar WP editor and there'd be no confusion about where to find it. We already do this when creating a duplicate from the single prompt editor screen, so I'm talking about doing this from the Campaigns dashboard, too.

I worry that with an automatic redirection some users might not expect this.
If the user wants to duplicate a bunch of prompts before editing them they'd have to go all the way back to Newspack > Campaigns each time.
So... what if:

Modal:
- Title
- Notice: the prompt will be "unassigned"
- Button: "Cancel" (tertiary)
- Button: "Edit" (secondary) duplicates the prompt and takes you to the editor
- Button: "Duplicate" (primary) you stay in the campaigns dashboard

Or we could simplify my original suggestion:

Modal:
- Title
- Button: "Cancel" (secondary)
- Button: "Duplicate" (primary):
    - Success message
    - Notice: the prompt is currently "unassigned"
    - Button: "Close" (secondary) closes the modal
    - Button: "Edit" (primary) takes you to the editor

dkoo:
Oh, I like the second suggestion (showing an optional link to edit the duplicated prompt in the success message). Sounds like the best of both worlds to me. I'll move forward with that idea in a separate PR. Thanks for brainstorming!

Issue #556, opened 2021-06-09
Editor crash on single prompt preview


dkoo:
This may be an error with the Newspack Components NPM package, but I couldn't reproduce in the components demo page nor from the Campaigns dashboard (also in Newspack Plugin), only here. 

1. Edit a prompt
2. Preview the prompt from the block editor
3. Observe a hard editor crash and JS console error:

<img width="946" alt="Screen Shot 2021-06-09 at 11 21 12 AM" src="https://user-images.githubusercontent.com/2230142/121400449-d4b39580-c914-11eb-8319-92beeda4310a.png">

```
_icons is undefined
    WebPreview webpack:///./node_modules/newspack-components/dist/cjs/web-preview/index.js?:99
    render webpack:///./node_modules/newspack-components/dist/cjs/web-preview/index.js?:229
    Ie https://siteurl/wp-includes/js/dist/vendor/react-dom.min.js?ver=16.13.1:104
    [...]
```

Seems to stem from not being able to import the `@wordpress/icons` package, but the latest version of this package is properly defined as a dependency in Newspack Components, so I'm not sure what the problem is.

dkoo:
I spent a lot of time trying to debug this, without getting anywhere so far. Seems to be a pretty hairy dependency bug related to the replacement of Material icons with WP icons in #982. The only thing I could find is that when following the links from the JS error into the dist file for `web-preview.js`, webpack seems to be calling `@wordpress/icons` using the wrong path. Note the paths for `i18n` and `element` vs. `icons`:

```
var _element = __webpack_require__(/*! @wordpress/element */ "@wordpress/element");

var _i18n = __webpack_require__(/*! @wordpress/i18n */ "@wordpress/i18n");

var _icons = __webpack_require__(/*! @wordpress/icons */ "./node_modules/@wordpress/icons/build-module/index.js");
```

However, I'm not sure what this means and I couldn't figure out how to update it. Things I tried with no luck:

* Installing `@wordpress/icons` as a top-level dependency in Campaigns' package.json
* Running `npm install` using various versions of NPM
* Downgrading webpack to the same version used in `newspack-plugin`

I wonder if this is a bug in `@wordpress/icons` itself that we won't be able to resolve in our code?

adekbadek:
[...]

PR #564, opened 2021-06-18
feat: improved duplication UI via modal

### LINKED ISSUE #560: "Duplicate prompt" UI improvements
#558 adds the ability to duplicate prompts. @thomasguillot has some good ideas on how to improve the UI to do so:

* When we click "Duplicate", should it open a modal so we can have the option to rename the prompt and display the category filtering to enable the user to change these
* Add a notice if the prompt will be "unassigned"
* Once the user clicks on the "Duplicate" button (Primary -- modal) they get a confirmation message with a link to edit it in Gutenberg?
* If they click "Cancel" (Secondary -- modal)  then nothing happens

dkoo:
### All Submissions:

Improves the UI for duplicating prompts. When duplicating a prompt from the single-prompt edit screen, you are now presented with a modal with the opportunity to rename the duplicate before creating it, an option to close the modal or edit the duplicate after it's created, and a notice if the duplicate is created without a campaign (which would cause it to be visible in the Campaigns dashboard only under the "unassigned" filter).

Closes #560.

#### Campaigns dashboard UI

1. Go to **Newspack > Campaigns**. Open the primary popover menu and select "Duplicate". Confirm that you now see a modal with a text input to rename the duplicate, which loads the default name ("Prompt title copy", "Prompt title copy 2", etc.) at first.

Screen Shot 2021-06-17 at 4 18 30 PM

2. Edit the name, but click "Cancel", click "Duplicate" again on the same prompt, confirm that the changes you had entered are discarded.
3. Click "Cancel" again and click "Duplicate" for another prompt. Confirm that the modal now shows info for whatever prompt you clicked "Duplicate" for.
4. Duplicate the prompt. Confirm that after successful duplication, you are presented with a success message and the option to close the modal (remaining on the Campaigns dashboard page) or edit the duplicate (which redirects you to the edit page for the newly created prompt). You can also test the failure state by blocking the request in the Dev Tools > Network tab. Note that when blocking a request you must attempt to duplicate the same prompt and with the same title as when blocked. A failed duplication request should result in the standard "fatal error" modal for all Newspack dashboard pages.

Screen Shot 2021-06-17 at 4 24 07 PM

5. Edit the duplicate and confirm that the title of the duplicate matches whatever you entered in the input field before, and that the content, taxonomy terms, and prompt settings all match those of the prompts you duplicated.
6. Create another duplicate, but make sure you duplicate a prompt that's not assigned. Confirm that both before and after you create the duplicate, the modal shows a warning that the prompt is unassigned.

Screen Shot 2021-06-18 at 2 37 55 PM

7. Also test that if you create multiple duplicates of the same prompt (or duplicates of the duplicates of a prompt), the default title increments as expected.

Screen Shot 2021-06-17 at 4 24 18 PM

#### Single prompt edit UI

1. Edit any prompt. Click the "Duplicate" button next to the "Preview" button.
2. The modal UI should be pretty much exactly the same as from the Campaigns dashboard. However, we're using core WP components here instead of Newspack components because when I tried using Newspack components, I kept running into [this same issue](https://github.com/Automattic/newspack-popups/issues/556) that happens when trying to preview. @thomasguillot or @adekbadek, do you have any ideas about why the `@wordpress/icons` package seems to not be importable when Newspack Components are imported as an NPM package? If we can solve this bug, we can switch the components here to Newspack Components instead of core.

Screen Shot 2021-06-18 at 2 42 21 PM

### Other information:

### DIFF
--- a/src/editor/Duplicate.js
+++ b/src/editor/Duplicate.js
@@ -2,16 +2,57 @@
  * WordPress dependencies
  */
 import apiFetch from '@wordpress/api-fetch';
-import { Button } from '@wordpress/components';
+import { Button, Modal, Notice, TextControl } from '@wordpress/components';
 import { compose } from '@wordpress/compose';
 import { withSelect, withDispatch } from '@wordpress/data';
-import { __ } from '@wordpress/i18n';
+import { useEffect, useState } from '@wordpress/element';
+import { __, sprintf } from '@wordpress/i18n';
+import { addQueryArgs } from '@wordpress/url';
 
-const DuplicateButton = ( { autosave, createNotice, isSavingPost, postId } ) => {
-	const duplicatePrompt = async () => {
+const DuplicateButton = ( {
+	autosave,
+	campaignGroups,
+	duplicateOf,
+	isSavingPost,
+	postId,
+	title,
+} ) => {
+	const [ error, setError ] = useState( null );
+	const [ modalVisible, setModalVisible ] = useState( false );
+	const [ duplicateTitle, setDuplicateTitle ] = useState( null );
+	const [ duplicated, setDuplicated ] = useState( null );
+
+	useEffect(() => {
+		setError( null );
+		if ( modalVisible && ! duplicateTitle ) {
+			getDefaultDupicateTitle();
+		}
+		if ( ! modalVisible ) {
+			setDuplicated( null );
+			setDuplicateTitle( null );
+		}
+	}, [ modalVisible ]);
+
+	const getDefaultDupicateTitle = async () => {
+		const promptToDuplicate = parseInt( duplicateOf || postId );
+		try {
+			const defaultTitle = await apiFetch( {
+				path: `/newspack-popups/v1/${ promptToDuplicate }/duplicate`,
+			} );
+
+			setDuplicateTitle( defaultTitle );
+		} catch ( e ) {
+			setDuplicateTitle( title + __( ' copy', 'newspack-popups' ) );
+		}
+	};
+
+	const duplicatePrompt = async ( popupId, titleForDuplicate ) => {
+		setError( null );
 		try {
 			const newId = await apiFetch( {
-				path: `/newspack-popups/v1/${ postId }/duplicate`,
+				path: addQueryArgs( `/newspack-popups/v1/${ popupId }/duplicate`, {
+					title: titleForDuplicate,
+				} ),
 				method: 'POST',
 			} );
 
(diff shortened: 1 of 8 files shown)


thomasguillot (approved):
@dkoo, I pushed a few changes: https://github.com/Automattic/newspack-popups/pull/564/commits/e8a5cf81b446b685768f4325b52bec0bc35a7e88

thomasguillot:
> do you have any ideas about why the @wordpress/icons package seems to not be importable when Newspack Components are imported as an NPM package? If we can solve this bug, we can switch the components here to Newspack Components instead of core.

I don't know why it crashes but we should be using Core components in the editor anyway.

dkoo:
@thomasguillot Changes look good, thanks!

src/editor/Duplicate.js
```diff
@@ -2,58 +2,168 @@
...
+								<Button isPrimary href={ `/wp-admin/post.php?post=${ duplicated }&action=edit` }>
+									{ __( 'Edit', 'newspack-popups' ) }
+								</Button>
+							</Flex>
+						</>
+					) : (
+						<>
+							{ ( ! campaignGroups || 0 === campaignGroups.length ) && (
+								<Notice status="warning" isDismissible={ false }>
+									{ __( 'This prompt will be unassigned.', 'newspack-popups' ) }
```
adekbadek, line 117: Just to make it more understandable:

```suggestion
									{ __( 'The new prompt will no be assigned to any campaign.', 'newspack-popups' ) }
```

src/editor/Duplicate.js
```diff
@@ -2,58 +2,168 @@
...
+						<>
+							<Notice status="success" isDismissible={ false }>
+								{ sprintf(
+									__( 'Duplicate of “%s” created as a draft.', 'newspack-popups' ),
+									title
+								) }
+							</Notice>
+							{ ( ! campaignGroups || 0 === campaignGroups.length ) && (
+								<Notice status="warning" isDismissible={ false }>
+									{ __( 'The prompt is currently unassigned.', 'newspack-popups' ) }
```
adekbadek, line 101: WDYT think about demystifying "unassigned" a bit:

```suggestion
									{ __( 'This prompt is currently not assigned to any campaign.', 'newspack-popups' ) }
```

adekbadek (changes requested):

src/editor/Duplicate.js
dkoo: Good suggestion! Updated in d40359b.

adekbadek (approved):

PR #572, opened 2021-07-12
fix(wp-5.8): avoid an editor crash due to missing data in widgets block editor

### LINKED ISSUE #570: [WP 5.8] Campaigns plugin causes editor crash in widget block editor
In the new blocks-based widget editor, the Campaigns plugin may cause an editor crash with the following JS error message:

```
TypeError: _wordpress_edit_post__WEBPACK_IMPORTED_MODULE_4__ is undefined
    PopupsSettingsPanel webpack:///./src/document-settings/index.js?:31
```

This seems to be caused by the use of `PluginDocumentSettingPanel` in a non-post editor context [here](https://github.com/Automattic/newspack-popups/blob/master/src/document-settings/index.js).

Since the option to "Disable prompts on this post or page" is irrelevant for the widgets editor, we should probably not enqueue that JS file on the widgets editor page [here](https://github.com/Automattic/newspack-popups/blob/master/includes/class-newspack-popups.php#L388-L401).

dkoo:
### All Submissions:

Avoids a hard editor crash in the new WP 5.8 widgets block editor by not enqueuing editor assets that assume we're editing a single post.

Closes #570.

This one is a bit hard to test because the crash doesn't always occur on every site. To truly confirm the fix, you'll need to have access to a site such as [this one](https://wp58test-localprofile.newspackstaging.com/wp-admin/widgets.php) where the bug is confirmed to be happening with the latest plugin release (ping me for credentials to this site).

1. On `master` or the latest release package, visit **Appearance > Widgets** and observe a hard editor crash with a JS console error such as:

```
TypeError: _wordpress_edit_post__WEBPACK_IMPORTED_MODULE_4__ is undefined
    PopupsSettingsPanel webpack:///./src/document-settings/index.js?:31
    ...
```

2. Check out and build this branch as a package, e.g. `npm run build && npm run release:archive`
3. Install/activate the built package and refresh the Widgets page. Confirm that the crash no longer occurs, that you're able to place the Campaigns blocks (Custom Placement, Single Campaign) in a widget area, and that the blocks render their prompts according to segmentation logic as expected on the front-end.
4. If you're not able to replicate the crash in the first place, you can confirm in the JS console on the Widgets page by looking for `window.newspack_popups_data`. This object should be `undefined` on this branch.

### Other information:

### DIFF
--- a/includes/class-newspack-popups.php
+++ b/includes/class-newspack-popups.php
@@ -385,6 +385,11 @@ final class Newspack_Popups {
 		wp_style_add_data( 'newspack-popups-blocks', 'rtl', 'replace' );
 		wp_enqueue_style( 'newspack-popups-blocks' );
 
+		// Don't enqueue Prompt editor files if we don't have a valid post type or ID (e.g. on the Widget Blocks screen).
+		if ( empty( $screen->post_type ) || empty( get_post_ID() ) ) {
+			return;
+		}
+
 		if ( self::NEWSPACK_POPUPS_CPT !== $screen->post_type ) {
 			if ( 'page' !== $screen->post_type || 'post' !== $screen->post_type ) {
 				// Script for global settings.
--- a/includes/class-newspack-popups.php
+++ b/includes/class-newspack-popups.php
@@ -386,7 +386,7 @@ final class Newspack_Popups {
 		wp_enqueue_style( 'newspack-popups-blocks' );
 
 		// Don't enqueue Prompt editor files if we don't have a valid post type or ID (e.g. on the Widget Blocks screen).
-		if ( empty( $screen->post_type ) || empty( get_post_ID() ) ) {
+		if ( empty( $screen->post_type ) || empty( get_the_ID() ) ) {
 			return;
 		}
 
--- a/includes/class-newspack-popups.php
+++ b/includes/class-newspack-popups.php
@@ -391,8 +391,9 @@ final class Newspack_Popups {
 		}
 
 		if ( self::NEWSPACK_POPUPS_CPT !== $screen->post_type ) {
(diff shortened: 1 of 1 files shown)


adekbadek:
I missed the link in #570 and duplicated the effort in #575 😬 I'd be inclined towards my solution, though, because it fixes this terrible line:

https://github.com/Automattic/newspack-popups/blob/8e12b2efcc98d5cce5332cc2770791b7ba0a1af1/includes/class-newspack-popups.php#L389

dkoo:
> I missed the link in #570 and duplicated the effort in #575 😬 I'd be inclined towards my solution, though, because it fixes this terrible line:

Oh, whoops! I agree that I had a hard time figuring out what that line is supposed to do. 😆 But one thing this branch does that yours doesn't is also bail before the `editor.js` files are loaded. These files load the sidebar options that control post meta, so they aren't relevant or needed in the widgets block editor. Not loading them also avoids this PHP warning which gets thrown in the latest RC nightly build:

```
[13-Jul-2021 18:29:55 UTC] PHP Notice:  wp_enqueue_script() was called <strong>incorrectly</strong>. "wp-editor" script should not be enqueued together with the new widgets editor (wp-edit-widgets or wp-customize-widgets). Please see <a href="https://wordpress.org/support/article/debugging-in-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 5.8.0.) in /Users/dkoo/Local Sites/newspack/app/public/wp-includes/functions.php on line 5535
```

I've cherry-picked your commit from the other PR. What do you think about combining our solutions?

adekbadek:
> What do you think about combining our solutions?

🙌👌

adekbadek (approved):

PR #579, opened 2021-07-16
fix: assume DB prefix; ensure https

adekbadek:
### All Submissions:

Small tweaks:
1. Assume the DB prefix is `wp_` if it's not defined. 
2. Ensure API URL has uses HTTPS - it happened on my local instance that the `plugins_url` returned HTTP URL despite the site URL set to HTTPS. Don't know why, but this won't hurt - if site is not HTTPS, `amp-access` won't work anyway.

1. Remove `DB_PREFIX` definition from your Campaigns config file
2. Ensure the Campaigns API works correctly

### Other information:

### DIFF
--- a/api/setup.php
+++ b/api/setup.php
@@ -60,11 +60,12 @@ if ( file_exists( WP_CONTENT_DIR . '/object-cache.php' ) ) {
 }
 
 global $wpdb;
+$db_prefix = defined('DB_PREFIX') ? DB_PREFIX : 'wp_';
 $wpdb = new wpdb( DB_USER, DB_PASSWORD, DB_NAME, DB_HOST );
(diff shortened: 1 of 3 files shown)


miguelpeixe (approved):
Can confirm the campaigns API did not work without `DB_PREFIX` defined and this PR fixes it.

I was not able to reproduce the HTTP protocol issue, but it looks like good practice.

Perhaps we should add a notice to Newspack users without SSL/TLS about the importance of it and incompatibility issues?

adekbadek:
> Perhaps we should add a notice to Newspack users without SSL/TLS about the importance of it and incompatibility issues?

Good idea, this could be an admin notice. But it's not that important ATM.

PR #584, opened 2021-07-21
feat(api): remove reader data if no updates in 30 days

### LINKED ISSUE #583: Data pruning: prune transients rows > 30 days?
In order to avoid server performance issues, we prune reader activity data daily to ensure that the read events (data on posts read) and transients (data on reader activity, e.g. subscription, donation, or prompt dismissal) tables don't get too big. However, for some very large sites the transients table still gets extremely big (~10 million rows). We prune read events older than 30 days—should we do the same for transients data?

The [current behavior](https://github.com/Automattic/newspack-popups/blob/master/includes/class-newspack-popups-segmentation.php#L630-L642) prunes preview data, but that doesn't seem to be enough for some sites. Another option: only prune data older than 30 days if the number of transients rows approaches 1 million or some other reasonable limit.

adekbadek:
### All Submissions:

Remove reader data not updated in 30 days and not containing any donations or subscriptions.

Closes #583.

There is a test added, but double-checking won't hurt.

1. Manipulate the reader data* so that you can test the following premises:
    1. A reader who did subscribe or donate, and was not updated since more than 30 days should **not** be removed
    2. A reader who did not subscribe nor donate, and was **not updated** since more than 30 days should be removed
    3. A reader who did not subscribe nor donate, and **was updated** less than 30 days ago should **not** be removed
2. Force pruning by running `wp cron event delete newspack_popups_segmentation_data_prune` and loading any admin page
3. Observe pruning results in the DB 

\* In the `wp_newspack_campaigns_transients` table. The `date` (the last update date) can be changed directly, for the encoded data (donations, subscriptions) you can change the value of `$value` saved to the table [here](https://github.com/Automattic/newspack-popups/blob/b4e4e6d1ef3ba18fb2a45e8df38fa06ca32e34fc/api/classes/class-lightweight-api.php#L193-L208).

### Other information:

### DIFF
--- a/includes/class-newspack-popups-segmentation.php
+++ b/includes/class-newspack-popups-segmentation.php
@@ -633,15 +633,20 @@ final class Newspack_Popups_Segmentation {
 
 		// Remove reader data if not containing donations nor subscriptions, and not updated in 30 days.
 		$removed_rows_count_transients = $wpdb->query( "DELETE FROM $transients_table_name WHERE `option_value` LIKE '%\"donations\";a:0%' AND `option_value` LIKE '%\"email_subscriptions\";a:0%' AND date < now() - interval 30 DAY" ); // phpcs:ignore WordPress.DB.PreparedSQL.InterpolatedNotPrepared, WordPress.DB.DirectDatabaseQuery.NoCaching, WordPress.DB.DirectDatabaseQuery.DirectQuery
-		error_log( 'Newspack Campaigns: Data pruning – removed ' . $removed_rows_count_transients . ' rows from ' . $transients_table_name . ' table.' ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
 
 		// Remove all preview sessions data.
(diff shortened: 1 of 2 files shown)


claudiulodro (approved):
Works great! Thanks for implementing this so quickly! 

One thing I noticed, since the `error_log` is always called, is that the output shows up in PHPunit:

```
..............................................................Newspack Campaigns: Data pruning – removed 1 rows from wptests_newspack_campaigns_transients table.
Newspack Campaigns: Data pruning – removed 0 preview session rows from wptests_newspack_campaigns_transients table.
Newspack Campaigns: Data pruning – removed 0 rows from wptests_newspack_campaigns_events table.
.   63 / 63 (100%)
```

Not a big deal or blocking though (and I'm not sure the best solution), and it's been going on for a while anyways. :)

adekbadek:
@claudiulodro does bac91d11df574274458708624ece63449ebe89c7 help?

PR #585, opened 2021-07-21
feat: sanitize utm suppression setting

### LINKED ISSUE #428: Question Mark In UTM Suppression Field Breaks API
If there is a question mark in the UTM Suppression field of a campaign it breaks the parameters sent to the API. To reproduce:

1) Create a campaign, set UTM Suppression to `?utm_medium=newsletter`, publish
2) View the site in an incognito window.
3) In Developer Console->Network->XHR, look for a request to `/wp-content/plugins/newspack-popups/api/campaigns/index.php`
4) Observe there is no response.
5) View Headers->Query String Parameters, observe `cid` and `popups` are present but `settings` is not.
6) Change the UTM Suppression to `newsletter`, observe the campaign performs correctly.

The API bails [here](https://github.com/Automattic/newspack-popups/blob/master/api/campaigns/class-maybe-show-campaign.php#L26-L28) because settings is missing from the request.

adekbadek:
### All Submissions:

Closes #428.

Observe #428 irreproducible.

### Other information:

### DIFF
--- a/includes/class-newspack-popups.php
+++ b/includes/class-newspack-popups.php
@@ -173,7 +173,7 @@ final class Newspack_Popups {
 				'single'            => true,
 				'auth_callback'     => '__return_true',
 				'sanitize_callback' => function( $input ) {
-					return preg_replace( '/\?|"/', '', $input );
+					return preg_replace( '~[^-a-z0-9_]+~', '', $input );
 				},
 			]
 		);
--- a/includes/class-newspack-popups.php
+++ b/includes/class-newspack-popups.php
@@ -173,7 +173,7 @@ final class Newspack_Popups {
 				'single'            => true,
 				'auth_callback'     => '__return_true',
 				'sanitize_callback' => function( $input ) {
-					return preg_replace( '~[^-a-z0-9_]+~', '', $input );
+					return preg_replace( '~[^-\w0-9_\s]+~', '', $input );
 				},
 			]
 		);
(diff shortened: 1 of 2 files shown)


dkoo (changes requested):
It seems like any [reserved URL character](https://developers.google.com/maps/url-encoding) triggers this bug. What about using `urlencode` to encode the value on save, and passing the value through `urldecode` before displaying in the editor or using for comparison in the API?

adekbadek:
> It seems like any reserved URL character triggers this bug.

Good catch!

> What about using urlencode

Still, the encoded/decoded characters will break the API. Anyway, reserved URL characters should never be part of the UTM parameter - I've updated the regex and added a placeholder to the input in fa7d7ab5af1abc0dcc38c82b9a9320118f77f255

dkoo (approved):
So the regex only allows alphanumeric characters, whitespace, `-` or `_`? Sounds good to me 👍

My last non-blocking suggestion would be to maybe add a `help` prop to the editor input field explaining that "The UTM campaign name must be alphanumeric only (A–Z, 0–9, -, _, or spaces)."

adekbadek:
> My last non-blocking suggestion would be to maybe add a help prop to the editor input field explaining that "The UTM campaign name must be alphanumeric only (A–Z, 0–9, -, _, or spaces)."

I hoped the placeholder would do the trick. This is a semi-advanced feature, so I think this should suffice 🤞

PR #586, opened 2021-07-22
feat: show prompts inside widget blocks in customizer preview

### LINKED ISSUE #577: [WP 5.8] Show prompts in Customizer preview
WP 5.8 introduces blocks-based widgets, which can be edited in the Customizer. This includes prompts via the Custom Placement and Single Prompt blocks. We made a decision early on to not show prompts to logged-in users, so prompts placed in widget areas via the Customizer's widget editor are not shown in the preview.

### The fix

The simple solution would be to show the first prompt that matches the block attributes—published or unpublished, regardless of segment—in the Customizer preview. Or if not a prompt, then a placeholder element that suggests a prompt would be rendered in this widget area. I'd suggest we do this as an MVP for the 5.8 release on July 20.

An enhancement to this would be to use the existing segmentation preview functionality from the Campaigns dashboard in the Customizer as well, and let admins choose which segment they're previewing as within the Customizer. This is a much more complex solution that would also require new UI inside the Custom Placement and Single Prompt blocks, when edited in the Customizer only. I'm not sure this would be worth the extra effort over the simpler solution.

dkoo:
### All Submissions:

In WP 5.8, it's possible to add blocks to widget areas in the Customizer. These blocks currently don't show anything when logged in, so this PR ensures that if a widget block can show a prompt, it displays at least something in the widget area in the Customizer preview.

It does this by using the new [`widget_block_content`](https://github.com/WordPress/WordPress/blob/master/wp-includes/widgets/class-wp-widget-block.php#L68-L82) filter to add a shortcode attribute when a prompt is displayed inside a widget block, then if the block is being rendered in the Customizer preview, it bypasses the AMP access check to ensure that the block displays the first prompt that can be shown by the block.

Question: is it now weird that only prompts inside widgets are shown in the preview? At first I tried making all prompts be displayed in the Customizer, but it got annoying because every prompt is displayed every time you change a Customizer setting (so every overlay pops up again even if dismissed, for example). I think showing only prompts that are inside widget blocks is a good compromise even if it's not logically consistent.

Closes #577.

1. Publish three inline prompts. Make sure two are assigned to a custom placement.
2. Go to **Appearance > Customize > Widgets**. In any widget area, add a Single Prompt block and a Custom Placement block and select a prompt and custom placement containing prompts, respectively. In the Customizer preview, navigate to a page that displays the widget area with your blocks and observe that nothing is shown in the preview widget area.
3. Check out this branch, refresh the Customizer, and repeat step 2. This time confirm that one prompt is shown for each block, ignoring segmentation rules for previewing purposes.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -86,6 +86,9 @@ final class Newspack_Popups_Inserter {
 		// Always enqueue scripts, since this plugin's scripts are handling pageview sending via GTAG.
 		add_action( 'wp_enqueue_scripts', [ $this, 'enqueue_scripts' ] );
 
+		// Filter prompt shortcode if ti's being rendered as a widget block.
+		add_filter( 'widget_block_content', [ $this, 'render_prompt_inside_widget_block' ] );
+
 		add_filter(
 			'newspack_popups_assess_has_disabled_popups',
 			function ( $disabled ) {
@@ -412,6 +415,9 @@ final class Newspack_Popups_Inserter {
 	 * @return HTML
 	 */
 	public static function popup_shortcode( $atts = array() ) {
+		$is_widget_block = isset( $atts['is-widget-block'] ) && 'true' === $atts['is-widget-block'];
+		$always_show     = $is_widget_block && is_customize_preview(); // Show prompt content if being rendered as a widget block in the Customizer preview.
+
 		if ( isset( $atts['id'] ) ) {
 			$include_unpublished = Newspack_Popups::is_preview_request();
 			$found_popup         = Newspack_Popups_Model::retrieve_popup_by_id( $atts['id'], $include_unpublished );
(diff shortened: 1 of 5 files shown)


adekbadek (changes requested):
This seems to me to be adding unnecessary complexity - there's already a mechanism for preview sessions (used by segment previews in `newspack-plugin`), and ideally it should be reused. 
I have an idea about how to do that, I should have something to propose as an alternative soon.

adekbadek:
Proposing an alternative in https://github.com/Automattic/newspack-popups/pull/587. The main difference is that the Customizer interaction is a separate "preview session", just like segment previews in the Campaigns Wizard.

dkoo:
@adekbadek I haven't had a chance to test #587 yet, but I think you're correct—it's a better approach than what's laid out here. Closing this PR in favor of the new one.

PR #587, opened 2021-07-26
Display popups in the Customizer

### LINKED ISSUE #577: [WP 5.8] Show prompts in Customizer preview

adekbadek:
### All Submissions:

Closes #577, alternative approach than #586.

As pointed out [in #586](https://github.com/Automattic/newspack-popups/pull/586#pullrequestreview-714657397), there's already a mechanism in place for preview session - and opening the Customizer is just that. This PR uses that approach, which is the same as in the Campaigns Wizard.

The biggest change is that _all_ active popups' data will be included in the request to the API (the `amp-access` endpoint). This is because with all the possible ways of inserting prompts, retrieving just the rendered ones gets unmanageable. The cost of a bigger payload is worth the gain in reduced complexity and reduced chance of missing a prompt in `amp-access` IMO.

One small, unrelated addition is the improvement of the debug message in 2f2abda98a71bacc78933c9975528b77038f4c20, which should be quite self-explanatory.

Similar to steps as in #586:

1. Publish three inline prompts. Make sure two are assigned to a custom placement.
2. Go to **Appearance > Customize > Widgets**. In any widget area, add a Single Prompt block and a Custom Placement block and select a prompt and custom placement containing prompts, respectively. In the Customizer preview, navigate to a page that displays the widget area with your blocks and observe that nothing is shown in the preview widget area.
3. Check out this branch, refresh the Customizer, and repeat step 2. This time confirm that one prompt is shown for each block, ignoring segmentation rules for previewing purposes.
4. Observe that the Customizer interaction is a separate "preview session" - if an overlay prompt is dismissed, that action is remembered, but the overlay will show up again on a subsequent Customizer session.

### Other information:


miguelpeixe (changes requested):
I'm having issues when I switch the custom placement. When I first select the widget, the popup is displayed correctly, but if I switch to a different placement it doesn't display the popup for this placement, nor the initial placement if I switch back:


https://user-images.githubusercontent.com/820752/127178407-f85e3251-6057-482d-9b37-dad98508acfb.mov

adekbadek:
> When I first select the widget, the popup is displayed correctly, but if I switch to a different placement it doesn't display the popup for this placement, nor the initial placement if I switch back:

Good catch, thanks! Should be fixed inc15723af1eb43b35dfb3cd1829435fe7b8d1e80a

miguelpeixe (approved):
Working as expected! Thank you for the fix!

adekbadek:
Noticed one issue - inline prompts were displayed on pages when previewing. Fixed in 77b4de73bf844652db204b5c4c3f5e5b48a47ad3

miguelpeixe (changes requested):
This change makes the pages that use a template with sidebar not able to preview the inline popups in the customizer.

adekbadek:
Thanks for catching that! 172fc8ad6d627da2dcc063a6bbd6fc5cb27bb93b should fix it.

miguelpeixe (approved):
Working nicely! Thank you for the updates! 💯

PR #621, opened 2021-09-09
feat(inline prompt): add a new placement option for the archive pages

### LINKED ISSUE #450: New inline placement option: Archives
For inline prompts, we have two placement options:

1. In article content
2. Above header

"In article content" obviously only lets editors place inline prompts in actual post content. "Above header" can appear on posts, pages, or archive pages, but only at the very top of the page.

What about adding a third placement option—"In archive pages"—that has a slider similar to the "in article content" slider to choose position? For archive pages, this slider would represent the number of archive items to render before the prompt, with a minimum of 1 (so the prompt can never appear first) and a reasonable maximum of say, 20. If there aren't enough items on an archive page to meet this setting, then the prompt gets rendered at the end of the list. If the archive list is empty, then the prompt doesn't get rendered.

These prompts will follow category filtering rules: if the prompt has categories assigned, it will only appear on archive pages for those categories. If no categories are assigned, then they can appear on any archive page.

Another idea, maybe not necessary: there could also be an option to repeat the prompt—so if the "number of items before prompt" setting is 3, and the "repeat" option is enabled, then the prompt appears after every three items until the end of the archive page list.

Mockup:

Screen Shot 2021-02-17 at 9 55 39 AM

kariae:
### All Submissions:

enable inserting prompts in after x articles or every x article in the archive pages

fix #450

1. If it's not released, please checkout https://github.com/Automattic/newspack-theme/pull/1507 or add that small change manually inside the `newspack-theme` that you use.
2. Pull this PR, and add a new `inline` prompt from the campaign menu.
3. In the `placement` option select `In archive pages`.
4. Set `2` for example as the number of items to show before the prompt.
5. Load an archive page (as a non-admin) and observe that the prompt is displayed after the second article in the list.
6. Edit that prompt, and switch on the `repeat prompt`.
7. Reload the archive page and observe that the prompt now is displayed after every two articles.

### Other information:


includes/class-newspack-popups-model.php
```diff
@@ -231,6 +231,8 @@ public static function retrieve_preview_popup( $post_id ) {
 				'trigger_type'            => filter_input( INPUT_GET, 'trigger_type', FILTER_SANITIZE_STRING ),
 				'trigger_delay'           => filter_input( INPUT_GET, 'trigger_delay', FILTER_SANITIZE_STRING ),
 				'trigger_scroll_progress' => filter_input( INPUT_GET, 'trigger_scroll_progress', FILTER_SANITIZE_STRING ),
+				'trigger_posts_count'     => filter_input( INPUT_GET, 'trigger_posts_count', FILTER_SANITIZE_STRING ),
```
adekbadek, line 234: The other `trigger_*` meta values denote what triggers the prompt display, not insertion logic. How about `archive_insertion_posts_count`?

includes/class-newspack-popups-model.php
```diff
@@ -231,6 +231,8 @@ public static function retrieve_preview_popup( $post_id ) {
...
+				'repeat_prompt'           => filter_input( INPUT_GET, 'repeat_prompt', FILTER_VALIDATE_BOOLEAN ),
```
adekbadek, line 235: This meta name feels too generic for me, how about `archive_insertion_is_repeating`?

includes/class-newspack-popups.php
```diff
@@ -376,6 +402,17 @@ public static function preview_post_permalink() {
...
 
+	/**
+	 * Get preview archive permalink. Used to preview prompts in archive pages in the popup wizard.
+	 *
+	 * @return string
+	 */
+	public static function preview_archive_permalink() {
+		$categories = get_categories();
+
+		return count( $categories ) > 0 ? get_category_link( $categories[0] ) : '';
```
adekbadek, line 413: On my local instance, `get_categories` returns an associative array, which results in a wrong offset error. After changing the `$categories` assignment to `array_values( get_categories() );` it's fine. Can you confirm that `get_categories` returns a regular array?

adekbadek (changes requested):

includes/class-newspack-popups-model.php
kariae: Yes, make sense 👍

includes/class-newspack-popups-model.php
kariae: Yes, make sense 👍

includes/class-newspack-popups.php
kariae: Hum, I can confirm that `get_categories` returns a regular array
```
array (size=13)
  0 => 
    object(WP_Term)[3826]
      public 'term_id' => int 35
      public 'name' => string 'Arts' (length=4)
      public 'slug' => string '_newspack_arts' (length=14)
      public 'term_group' => int 0
      public 'term_taxonomy_id' => int 35
      public 'taxonomy' => string 'category' (length=8)
      public 'description' => string '' (length=0)
      public 'parent' => int 0
      public 'count' => int 3
      public 'filter' => string 'raw' (length=3)
      public 'cat_ID' => int 35
      public 'category_count' => int 3
      public 'category_description' => string '' (length=0)
      public 'cat_name' => string 'Arts' (length=4)
      public 'category_nicename' => string '_newspack_arts' (length=14)
      public 'category_parent' => int 0
  1 => 
    object(WP_Term)[3827]
      public 'term_id' => int 30
      public 'name' => string 'Entertainment' (length=13)
      public 'slug' => string '_newspack_entertainment' (length=23)
      public 'term_group' => int 0
      public 'term_taxonomy_id' => int 30
      public 'taxonomy' => string 'category' (length=8)
      public 'description' => string '' (length=0)
      public 'parent' => int 0
      public 'count' => int 4
      public 'filter' => string 'raw' (length=3)
      public 'cat_ID' => int 30
      public 'category_count' => int 4
      public 'category_description' => string '' (length=0)
      public 'cat_name' => string 'Entertainment' (length=13)
      public 'category_nicename' => string '_newspack_entertainment' (length=23)
      public 'category_parent' => int 0
...
```

Can you share a dump of `get_categories`, please? Adding a `array_values ` will fix it for your case and won't change mine, I just want to see how it looks.

kariae: So wired, using `error_log` I can still see a regular array:
```
[
    {
        "term_id": 35,
        "name": "Arts",
        "slug": "_newspack_arts",
        "term_group": 0,
        "term_taxonomy_id": 35,
        "taxonomy": "category",
        "description": "",
        "parent": 0,
        "count": 3,
        "filter": "raw",
        "cat_ID": 35,
        "category_count": 3,
        "category_description": "",
        "cat_name": "Arts",
        "category_nicename": "_newspack_arts",
        "category_parent": 0
    },
    {
        "term_id": 30,
        "name": "Entertainment",
        "slug": "_newspack_entertainment",
        "term_group": 0,
        "term_taxonomy_id": 30,
        "taxonomy": "category",
        "description": "",
        "parent": 0,
        "count": 4,
        "filter": "raw",
        "cat_ID": 30,
        "category_count": 4,
        "category_description": "",
        "cat_name": "Entertainment",
        "category_nicename": "_newspack_entertainment",
        "category_parent": 0
    },
```
Anyway, I guess that an `array_value` won't hurt here!

adekbadek (changes requested):
Sorry for the late re-review 🙈

kariae:
Thank you @adekbadek for your feedback, changes implemented :)

adekbadek (approved):
🙌

PR #624, opened 2021-09-23
Add/overlay more placements and size option

### LINKED ISSUE #114: Overlays: Add position and size option
This would allow users to choose where they want to display the overlay on the screen and how big or small it should be.

options

_Position is inspired by CoBlocks' Hero block._

kariae:
### All Submissions:

Closes #114.

1. Checkout the PR branch.
2. From the campaign menu, add a new popup prompt.
3. Observe the new placements selector and size options.
4. select any placement and size for the new popup and save it.
5. From the campaigns list, preview the newly created popup.
6. Observe the popup showing in the selected placement with the selected size.

### Some notes:

- The size option works for large screens only, for the small screens, it will always be at 100%.
- The popup size is set by default to full-width when the popup is centered on top or on the bottom.
- The sizes are: Small: 25%, Medium: 50%, Large: 75%, Full-width: 100%.

### Other information:


src/editor/PositionPlacementControl.js
```diff
@@ -0,0 +1,93 @@
...
+			/* translators: block layout */
+			label: __( 'Bottom right' ),
+		},
+	];
+
+	return (
+		<Fragment>
+			<div className="newspack-popups-css-grid-selector">
+				<p className="components-base-control__label">{ label }</p>
+				<ButtonGroup aria-label={ __( 'Select layout', 'coblocks' ) } { ...props }>
```
adekbadek, line 65: ;)

```suggestion
				<ButtonGroup aria-label={ __( 'Select layout', 'newspack-popups' ) } { ...props }>
```

src/editor/style.scss
```diff
@@ -62,3 +63,69 @@
...
+.newspack-popups-css-grid-selector {
+	margin-bottom: 1em;
+
+	.components-button-group {
+		display: grid;
+		grid-template-columns: repeat( 3, 1fr );
+		grid-template-rows: repeat( 3, 1fr );
+
+		div {
+			// background: $white;
```
adekbadek, line 90: Commented out code should be removed.

src/editor/style.scss
```diff
@@ -1,3 +1,4 @@
+// css from CoBlocks Hero placement selector
```
adekbadek, line 1: This comment should rather be placed above the relevant block, if at all.

src/editor/style.scss
```diff
@@ -62,3 +63,69 @@
...
+		display: flex;
+		flex-wrap: wrap;
+
+		.components-button {
+			display: table-cell;
+			padding: 18px;
+			width: 50%;
+		}
+	}
+}
```
adekbadek, line 79: Why is this necessary? Does not seem to have any effect in the editor.

src/editor/style.scss
```diff
@@ -62,3 +63,69 @@
...
+			padding: 20px;
+			width: 100%;
+
+			svg {
+				display: none;
+			}
+		}
+
+		.is-selected {
+			z-index: 99;
```
adekbadek, line 106: Why is this necessary? Does not seem to have any effect in the editor.

src/editor/style.scss
```diff
@@ -62,3 +63,69 @@
...
+		}
+
+		.components-button {
+			border-radius: 0 !important;
+			height: 100%;
+			padding: 20px;
+			width: 100%;
+
+			svg {
+				display: none;
```
adekbadek, line 101: There are no SVG elements rendered in this component.

src/view/style.scss
```diff
@@ -139,13 +143,61 @@ $font__system: -apple-system, blinkmacsystemfont, 'Segoe UI', roboto, oxygen, ub
...
 			overflow-y: auto;
-			width: 100%;
+			// width: 100%;
+		}
+
+		.newspack-popup {
+			// max-width: 90%;
+			padding-left: 0;
+			padding-right: 0;
+			// width: 1200px;
```
adekbadek, line 153: Commented out code should be removed - here & in few other instances in this file.

src/editor/style.scss
adekbadek, line 90: Commented out code should be removed.

src/editor/Sidebar.js
```diff
@@ -44,34 +47,45 @@ const Sidebar = ( {
...
-									label: customPlacements[ key ],
-								} ) )
-						  )
-				}
 			/>
+			{ isOverlay ? (
+				<PositionPlacementControl
+					layout={ placement }
+					label={ __( 'Position' ) }
+					help={ getPlacementHelpMessage( placement, trigger_scroll_progress ) }
```
adekbadek, line 56: The new cases should be handled in the help message.

adekbadek (changes requested):
@thomasguillot could you please take a look at the new sizes? For example, there's a min-width of about 700px on desktop for overlay prompts - maybe with small size, there should be no minimum width at all?

kariae:
Thank you @adekbadek for the review 🙏

kariae:
@thomasguillot Thank you for your help!

thomasguillot:
> @thomasguillot nice one, I just add this 👌

Looks great. I've just updated the max-width a little bit (https://github.com/Automattic/newspack-popups/pull/624/commits/57a390efb212152947a6b83e1c438ce071ba86f0)

src/editor/PositionPlacementControl.js
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * External dependencies
+ */
+import classnames from 'classnames';
+
+const PositionPlacementControl = ( { layout, label, help, onChange, size, ...props } ) => {
+	/**
+	 * Set layout options
+	 */
+	const layoutOptions =
```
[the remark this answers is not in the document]
kariae: IMHO the right way to do this would be to reset the placement whenever the site is set to full-width based on the previous placement value (e.g. if it was `center-left`, the new value would be `center`) the drawback to this would be the loss of the previous value, continuing on the same example if we select back a non-full-width size the position will still `center` and won't go back to `center-left`.

A hacky way to do this is to _fake_ the position selected by not relying only on `value === layout` but adding conditions depending on the size. Thing that should be considered in the backend too.

I think the first approach is safe and clean here, but I would like to have your inputs on this :)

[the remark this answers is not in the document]
adekbadek: > IMHO the right way to do this would be to reset the placement whenever the site is set to full-width based on the previous placement value

Yeah, makes sense 👌 

> the drawback to this would be the loss of the previous value

I don't think this is much of a concern, this is a pretty simple setting.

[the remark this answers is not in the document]
kariae: Thank you @adekbadek

src/editor/Sidebar.js
```diff
@@ -59,53 +84,59 @@ const Sidebar = ( {
...
+			) : (
+				<SelectControl
+					label={ __( 'Placement' ) }
+					help={ getPlacementHelpMessage( placement, trigger_scroll_progress ) }
+					value={ placement }
+					onChange={ updatePlacement }
+					options={ [
+						{ value: 'inline', label: __( 'In article content', 'newspack-popups' ) },
+						{ value: 'above_header', label: __( 'Above site header', 'newspack-popups' ) },
+						{ value: 'manual', label: __( 'Manual only', 'newspack-popups' ) },
```
adekbadek, line 115: `'archives'` placement is missing here.

adekbadek (changes requested):

src/editor/Sidebar.js
kariae: Nice catch 👍

src/editor/EditorAdditions.js
```diff
@@ -54,6 +54,19 @@ const EditorAdditions = () => {
...
+
+	// Setting editor size as per the popup size.
+	useEffect(() => {
+		const blockEditor = document.querySelector( '.block-editor-block-list__layout' );
+		if ( blockEditor ) {
+			if ( isOverlay( placement ) ) {
+				blockEditor.className = `block-editor-block-list__layout is-size-${ overlay_size }`;
+			} else {
+				// Reset the class in case we switch from overlay placement.
+				blockEditor.className = `block-editor-block-list__layout`;
```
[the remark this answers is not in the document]
kariae: Hum, agree with you 👍

adekbadek (approved):

PR #630, opened 2021-09-30
enable homepage-posts posts duplication  between a page and an overlay prompt

### LINKED ISSUE #126: Homepage Posts Block Rendering And Robbery
The Homepage Posts block creates a few problems when used in a Campaign. 

1) No posts are shown in the block.
2) If there are other instances of the block on the page, the deduplication logic "robs" them of their posts. 

#### Case 1: Inline placement

Screen Shot 2020-05-14 at 7 17 46 AM

#### Case 2: Homepage Overlay With No Filtering

Note that all blocks render without posts:

Screen Shot 2020-05-14 at 7 07 42 AM

#### Case 3: Homepage Overlay With Filtering

In this example the Homepage Posts in the campaign is showing posts tagged `events`, and as a result only those posts are robbed from the page's blocks.

Screen Shot 2020-05-14 at 7 31 37 AM

kariae:
### All Submissions:

enable showing the same post in popups and page if they both contain the homepage-posts block with
the same category filter.

fix #126 

1. On a new page add the homepage-posts block and filter the posts to be displayed by a category.
2. Add a new overlay prompt with a homepage-posts block as its content.
3. Filter the posts by the same category as 1. on the overlay homepage-posts block.
4. Load the page and observe that the posts on the page don't contain the posts on the popup.
5. Reload the page a second time and observe that the page still doesn't show the posts from the popup, but the popup doesn't show (popups are shown once or once a day).
6. If this https://github.com/Automattic/newspack-blocks/pull/886 is not merged, pull it, or apply the small changes from it.
7. Checkout this branch.
8. Reload the page.
9. Observe now the page loading all the category posts.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -900,6 +900,10 @@ final class Newspack_Popups_Model {
 			$popup = self::retrieve_preview_popup( Newspack_Popups::previewed_popup_id() );
 		}
 
+		if ( has_block( 'newspack-blocks/homepage-articles', $popup['content'] ) ) {
+			add_filter( 'newspack_blocks_homepage_enable_duplication', '__return_true' );
+		}
+
 		if ( ! self::is_overlay( $popup ) ) {
 			return self::generate_inline_popup( $popup );
 		}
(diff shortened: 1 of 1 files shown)


kariae:
> In the case of the popup being shown the posts are being duplicated as well. Is this the expected behavior?

Yes.

> The filters are currently being applied for every type of popup. What do you think about enabling duplication only for overlay popups?

Actually, it's a bug, thank you for raising this, I was intending to set this only to popups, I made the correction. Thank you @miguelpeixe 🙏

miguelpeixe:
Thank you for revising!

Inline and custom placement popups are now going through the deduplication process. The current logic does not ensure that overlay popups will contain every post queried. If a previous query populates the global `$newspack_blocks_post_id`, the overlay popup will not contain the populated posts.

You can test this behavior by having an "above header" popup with posts from the same category as an overlay popup and see that the overlay popup does not contain all of the queried posts. I don't know which other cases could occur, though.

miguelpeixe (approved):
This is a good solution! Thank you for the revisions 🎉

kariae:
Thank you @miguelpeixe ;)

PR #647, opened 2021-10-29
fix: prevent insertions which break layout; improve classic-content insertion

### LINKED ISSUE #555: Inline placement breaking inline images & sub-headings
Wisconsin Watch reported that their Recent Stories campaign often breaks the story text in odd places:

Screen Shot 2021-06-02 at 3 32 27 PM

[Post link](https://wisconsinwatch.org/2021/06/la-crosse-teacher-julie-welch-reflects-on-freedoms-brought-by-vaccination/)

Screen Shot 2021-06-02 at 3 34 21 PM

[Post link](https://wisconsinwatch.org/2021/05/a-milwaukee-man-explains-how-his-college-debt-spiraled-out-of-control-and-how-to-avoid-a-similar-fate/)

Screen Shot 2021-06-02 at 3 36 58 PM

[Post link](https://wisconsinwatch.org/2021/04/wisconsin-multigenerational-homes-high-covid-19-risk/)

-----

The [inline prompt](https://wisconsinwatch.org/wp-admin/post.php?post=1262604&action=edit#) is a homepage posts block that is set to display just the titles at 35%. It appears that we're not gracefully accounting for inline images or sub-headings. 

* Could we amend the content detection to better account for the right- and left- adjusted images? Can we set the logic to avoid adding a prompt after a header block? 
* Alternatively, can we add a means to manually override an inline prompt per-post?  

---

_Moved from https://github.com/Automattic/newspack-plugin/issues/989_

adekbadek:
### All Submissions:

Closes #555.

1. On `master`,
1. Create a post which has mid-content headings and left- or right-aligned (floated) images
2. Concoct a set of inline prompts (by assigning percentage-placement values) that break the layout: 
    1. prompts inserted right after a heading

image
 
    2. prompts inserted right after a left- or right-aligned image
 
image
  
3. Observe, as described in #555, that prompts can break the layout
4. Switch to this branch, observe issues solved:

image

image

5. Now do the same, but with a legacy content post. This can be mimicked by creating a post with a single "Classic" block.
6. Observe that prompts are inserted correctly into the classic content.  

### Other information:


includes/class-newspack-popups-inserter.php
```diff
@@ -309,15 +309,18 @@ function ( $popup ) use ( $shortcoded_popups_ids ) {
 					$inline_popup['is_inserted'] = true;
 				}
 			}
-			$block_content = render_block( $block );
-			$output       .= $block_content;
+
+			// Render blocks from the block group.
+			foreach ( $block_group as $block ) {
+				$block_content = render_block( $block );
```
miguelpeixe, line 315:
```suggestion
				$block_content = serialize_block( $block );
```

We should return the block serialized instead of rendered to not interfere with other filters.

includes/class-newspack-popups-inserter.php
```diff
@@ -326,9 +329,63 @@ function ( $popup ) use ( $shortcoded_popups_ids ) {
...
+			}
+		);
+
+		if ( empty( $popups ) ) {
+			return $content;
+		}
+
+		if ( function_exists( 'scaip_maybe_insert_shortcode' ) ) {
+			// Prevent default SCAIP insertion.
+			remove_filter( 'the_content', 'scaip_maybe_insert_shortcode', 10 );
```
miguelpeixe, line 374: With the new approach for detecting `core/heading` and aligned `core/image`, maybe we can do the same with SCAIP inserted content and preserve how the SCAIP filter is applied?

It would be better to maintain since we would be parsing another plugin inserted content instead of manually applying its logic.

miguelpeixe (changes requested):
It works as described without SCAIP. I couldn't find the root cause of the issue, but when SCAIP is inserting ads, the prompts are inserted after images just like on `master` branch.

Note: The `can_block_be_followed_by_prompt` logic will be very useful on SCAIP as well.

### Without SCAIP:

image

### With SCAIP

image

includes/class-newspack-popups-inserter.php
adekbadek: Done in 8a543475c5fd1004387ba9f582e0a48ba3d558bd

adekbadek:
@miguelpeixe 

> It works as described without SCAIP. I couldn't find the root cause of the issue, but when SCAIP is inserting ads, the prompts are inserted after images just like on master branch.

I've found that it's fixed by https://github.com/Automattic/super-cool-ad-inserter-plugin/pull/78

includes/class-newspack-popups-inserter.php
adekbadek: > With the new approach for detecting core/heading and aligned core/image, maybe we can do the same with SCAIP inserted content 

You mean add the same changes as this PR introduces to SCAIP?

> It would be better to maintain since we would be parsing another plugin inserted content instead of manually applying its logic.

Actually with the block-insertion change to SCAIP this manual triggering of SCAIP is no longer necessary. This block was here to prevent SCAIP from inserting ads mid-prompt, but it won't do that with the block-based insertion. I've removed the block in 7952d33e282a8b12f11f3d1a4e4b9066fff0adaf

miguelpeixe: > You mean add the same changes as this PR introduces to SCAIP?

No, I meant to use the `can_block_be_followed_by_prompt` to also check for SCAIP instead of modifying the filter. My guess is the `render_block()` returned here from a filter with priority 1 was also messing with the `parse_bock()` in SCAIP, which is priority 5. That would call for a manual triggering as well.

miguelpeixe (approved):

PR #653, opened 2021-11-04
fix: post type support conditions

adekbadek:
### All Submissions:

Cleans up the insertion logic a bit. Before, the post type conditions used a bit convoluted logic of checking the equality of globally-supported post types and prompt's post types. Also, the default value for the supported post types meta was set to the default post types in [meta registration](https://github.com/Automattic/newspack-popups/blob/89ba2147be4cdb4839eab5a4a59d02887a0a4fcf/includes/class-newspack-popups.php#L333), but to globally-supported post types as fallback in `get_popup_options` method. I've only run into this when running tests in isolation, and was not able to come up with a reproducible scenario.

Logic is refactored so that it's clear and readable:

1. If prompt's supported post types are same as the defaults, use the globally-supported post types
2. Otherwise, use prompt's post types – since this means the defaults were overridden

### How to test the changes in this Pull Request (if you need to!):

1. On `master`, run the `test_insertion_in_cpt_global` test in isolation*
2. Observe it failing
3. Switch to this branch, observe it passing

\* `phpunit --filter=test_insertion_in_cpt_global`

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -755,37 +755,35 @@ final class Newspack_Popups_Inserter {
 			}
 		}
 
-		// Context in which the popup appears
-		// - the taxonomy of the post.
-		$is_taxonomy_matching = self::assess_taxonomy_filter( $popup, 'category' ) && self::assess_taxonomy_filter( $popup, 'post_tag' );
-		// - the types of the posts supported for all popups.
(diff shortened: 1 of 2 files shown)


dkoo (approved):
This fixes the failing test and makes the logic of checking the post's supported post types vs. the globally supported post types much easier to follow! 🙌

PR #658, opened 2021-11-10
Don't use "here" as link text

### LINKED ISSUE #454: Don't use "here" as a link word
https://github.com/Automattic/newspack-popups/blob/master/includes/class-newspack-popups.php#L545

```
Newspack Campaigns requires a custom configuration file, which is missing.
Please create this file following instructions found here.
```

This is a bad UX and AX practice. See:
* https://www.smashingmagazine.com/2012/06/links-should-never-say-click-here/
* https://webaccess.berkeley.edu/ask-pecan/click-here

aschweigert:
Adjust the wording in the config instructions to avoid use of "here" as link text.

### All Submissions:

* [x ] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x ] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x ] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

Closes #454

### DIFF
--- a/includes/class-newspack-popups.php
+++ b/includes/class-newspack-popups.php
@@ -743,9 +743,9 @@ final class Newspack_Popups {
 		?>
 			<div class="notice notice-error">
 				<p>
-					<?php _e( 'Newspack Campaigns requires a custom configuration file, which is missing. Please create this file following instructions found ', 'newspack-popups' ); ?>
+					<?php _e( 'Newspack Campaigns requires a custom configuration file, which is missing. Please create this file by following ', 'newspack-popups' ); ?>
 					<a href="https://github.com/Automattic/newspack-popups/blob/master/README.md#config-file">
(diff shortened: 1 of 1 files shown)


adekbadek (approved):

paulschreiber:
To a screen reader user, "these instructions," "this link" and "here" are all unhelpful. Imagine a list of link text, out of context.


Something like this would be better
```
Newspack Campaigns requires a configuration file, which is missing. Please <a href="...">use these instructions to create a custom configuration file</a>.
```

PR #666, opened 2021-11-16
Alpha release Nov 16

adekbadek:
A new alpha release.


adekbadek:
Manually unwinding the unintentional major release (#638 was _not_ breaking after all)

PR #706, opened 2021-11-30
fix: naming duplicated prompts

### LINKED ISSUE #628: Duplicate prompt naming issue
## As is 

When duplicating a _duplicated_ prompt, the [title will be inherited](https://github.com/Automattic/newspack-popups/blob/f76db06ecdf5293ea0641bbeb9d862a99b36e5ee/includes/class-newspack-popups.php#L831) from the originally duplicated prompt, not the first duplicate.

## To be 

The title should be inherited from the duplicated prompt.

kariae:
### All Submissions:

Closes #628.

1. Create a prompt from the campaigns screen with `My Prompt` as a title.
2. Duplicate the prompt (from the prompt edit screen) and notice that the suggested name is correct and should be `My Prompt copy`.
3. Duplicate again the first prompt and change its title this time to `Second Prompt`.
4. Duplicate the `Second Prompt`, and notice that the title generated `My Prompt copy 2` is based on the original prompt and not the parent prompt.
5. Check out this PR.
6. Duplicate the first prompt and change its name to `Third Prompt`.
7. Duplicate `Third Prompt` and notice that this time, the title generated `Third Prompt copy` is based on the parent prompt.
8. Duplicate `Third Prompt` or `Third Prompt copy` and notice that the suggested title `Third Prompt copy 2` is based on the parent title and the total number of occurrences.

### Other information:

All the duplicate actions should be done using the `duplicate` button from the prompt edit screen. [A separate PR](https://github.com/Automattic/newspack-plugin/pull/1326) will add the functionality to the duplicate menu in the prompts list screen.


includes/class-newspack-popups.php
```diff
@@ -988,19 +1017,22 @@ public static function duplicate_popup( $id, $title = '' ) {
...
+				'post_title'    => ! empty( $title ) ? $title : self::get_duplicate_title( $duplicate_of, $id ),
 				'post_author'   => $old_popup->post_author,
 				'post_content'  => $old_popup->post_content,
 				'post_excerpt'  => $old_popup->post_excerpt,
 				'post_category' => wp_get_post_categories( $id, [ 'fields' => 'ids' ] ),
 				'tags_input'    => wp_get_post_tags( $id, [ 'fields' => 'ids' ] ),
 				'meta_input'    => [
-					'duplicate_of' => $original_id,
+					// A campaign is set as the origin of another one, if the later have the same title with the count of occurences suffixed (e.g. my prompt 3).
+					'duplicate_of' => empty( $title ) || preg_match( "/^$original_title_base\s*\d*$/", $title )
```
dkoo, line 1034: If we're solely basing the copy titles off of the name of the prompt being copied (whether or not it is itself a duplicate) plus the existence of other prompts with the same title, I believe we might be able to deprecate the `duplicate_of` meta field, right?

dkoo (changes requested):
I don't think the iterator is working as expected. Here's what I saw:

- [x] **Duplicating a prompt named `My Prompt`:** The suggested title is `My Prompt copy`, as expected.
- [ ] **Duplicating `My Prompt copy`:** The suggested title is `My Prompt copy copy`
  - **Expected:** I would expect the suggested title to be `My Prompt copy 2`. If there's already a prompt named `My Prompt copy 2`, I would expect `My Prompt copy 3` and so on.
- [ ] **Duplicating `My Prompt` again:** The suggested title is once again `My Prompt copy` despite there already being a prompt with that title.
  - **Expected:** I'd expect the suggested title to be `My Prompt copy 2` (or `3`, etc.) if there are already other prompts with the same title.

dkoo (approved):
My mistake! This works from both the editor screen and from the Campaigns dashboard in combo with https://github.com/Automattic/newspack-plugin/pull/1326. Suggested titles increment as expected if there were other prompts with the same `copy` name.

I see how you're using `duplicate_of` to query other copies of the same original prompt before incrementing the suggested title. This works well, but it does present an unlikely edge case where if you have two prompts with the same title that aren't duplicates of each other, duplicating each will result in the same suggested titles. But this is such an edge case that I'm okay with it.

kariae:
Thank you @dkoo for the review and feedback.

Yes, I'm using `duplicate_of` exactly as you described, and I had in mind that edge case, that technically is possible, but in the day to day usage we 99% won't have 😅

PR #717, opened 2021-12-02
fix: stricter empty block check

miguelpeixe:
### All Submissions:

#647 introduced a new bug in which some blocks are being stripped out of the page while being parsed by its filters. This is due the filtering out of blocks without `innerHTML`.

Not all blocks require `innerHTML` to display content. Blocks such as the Donation block and the Homepage Posts block uses other attributes to render their content.

This PR implements a stricter check to only filter out empty innerHTML of the following blocks:

- `null`
- core/paragraph
- core/heading
- core/list
- core/quote
- core/html
- core/freeform

It also implements a new assertion for more readable and less sensitive testing of block content insertion.

1. On the master branch, make sure you have inline campaigns visible
2. Create a page with donation and homepage posts block
3. Visit the page and see the blocks are not visible
4. Check out this branch and confirm they are displayed as expected

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -219,25 +219,6 @@ final class Newspack_Popups_Inserter {
 
 		$parsed_blocks = self::convert_classic_blocks( parse_blocks( $content ) );
 
-		// List of blocks that require innerHTML to render content.
-		$innerhtml_blocks_names = [
-			'core/paragraph',
-			'core/heading',
-			'core/list',
-			'core/quote',
-			'core/html',
-			'core/freeform',
-		];
-		$parsed_blocks          = array_values( // array_values will reindex the array.
-			// Filter out empty blocks.
-			array_filter(
-				$parsed_blocks,
-				function( $block ) use ( $innerhtml_blocks_names ) {
-					return ! ( in_array( $block['blockName'], $innerhtml_blocks_names ) && empty( trim( $block['innerHTML'] ) ) );
-				}
-			)
-		);
-
 		$block_index          = 0;
 		$parsed_blocks_groups = array_reduce(
 			$parsed_blocks,
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -228,7 +228,7 @@ final class Newspack_Popups_Inserter {
 			'core/html',
 			'core/freeform',
 		];
-		$parsed_blocks = array_values( // array_values will reindex the array.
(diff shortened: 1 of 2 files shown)


adekbadek (commented):
How about getting rid of that check instead of making it smarter? There might be other blocks (by other plugins) this can seem empty, we’ll never know. 
If a block is empty, it's 0-length content won't throw off the insertion computation anyway.

miguelpeixe:
a5e11fa71f303167273b3f44316dd6c38b70dd04 restores the check including `null` and 8827fbfd6939a7c54960fe453c5f6db2d1ea630d implements a new assertion for more readable and less sensitive testing of block content insertion.

includes/class-newspack-popups-inserter.php
```diff
@@ -219,12 +219,24 @@ public static function insert_popups_in_post_content( $content, $popups ) {
 
 		$parsed_blocks = self::convert_classic_blocks( parse_blocks( $content ) );
 
-		$parsed_blocks = array_values( // array_values will reindex the array.
+		// List of blocks that require innerHTML to render content.
+		$innerhtml_blocks_names = [
```
adekbadek, line 223: The comment and variable name are a bit unclear to me. This is effectively a list of blocks which render nothing if their `innerHTML` is empty. While I don't have a great idea for the variable name, I think the comment does not reflect the intention here.

tests/test-content-intertion.php
```diff
@@ -35,6 +35,18 @@ public static function rendered_popup( $id ) {
 		return '<!-- wp:shortcode -->[newspack-popup id="' . $id . '"]<!-- /wp:shortcode -->';
 	}
 
+	/**
+	 * Assert that that serialized blocks match the block names.
```
adekbadek, line 39:
```suggestion
	 * Assert that serialized blocks match the block names.
```

adekbadek (approved):
Just some naming/comment nitpicks 👌

includes/class-newspack-popups-inserter.php
miguelpeixe: I agree! What do you think of 93785c75ff2a9f3f9ea6be68dabb823df51d1016 ?

miguelpeixe:
For documentation purposes, we've restored the empty block check due to its importance on the _do-not-insert-after-block_ logic. Empty blocks passing through would visually disrupt the purpose of the logic, making it seem like a block is inserted after the restricted block when it's not.

adekbadek (approved):

PR #641, opened 2021-10-21
feat: enable excluding categories per prompt

kariae:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Closes #68.

### How to test the changes in this Pull Request:

1. Create a prompt (inline or overlay), open a post from category A.
2. Observe the prompt rendering on the post.
3. Checkout this.
4. Edit the prompt, from the advanced filter on the right, exclude category A.
5. Reload the post (you need to clear cache or switch to incognito in case of overlays), and observe the prompt is no longer rendered.
6. Edit the prompt, and include category A.
7. Reload the post and observe that the prompt is not loaded (exclusion rule has priority).
8. Bonus for the two tests:
```
phpunit --filter test_categories_exclusion
phpunit --filter test_categories_exclusion_priority_over_inclusion
```

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #726, opened 2021-12-07
feat: add a filter for supported post types

### LINKED ISSUE #720: Inline prompts do not appear on pages
image

`page` is an available post type for filtering and inserting inline prompts, but they do not appear.

It should either not allow this type of prompt configuration or display the configured prompts.

dkoo:
### All Submissions:

Instead of supporting just `[ 'post', 'page' ]` as the default post types for Campaigns, uses `get_post_types` to support posts, pages, and all public CPTs that have archives by default. Also applies a filter to the array of supported post types to allow other plugins to add or remove Campaigns support for specific post types.

Fixes some unexpected behavior that @kariae and I debugged together when trying to display an archive prompt on a category archive page that contained both posts and listings.

As a bonus, also fixes #720, as before we implemented post type support, we had hard-coded an exception to prevent inline prompts from being rendered on pages.

1. On your test site, make sure you have Newspack Listings and Newsletter plugins active, and some listings/newsletter posts published and assigned to categories or tags. These should have more recent publish dates so that they appear at the top of their respective category or tag archives, before any regular posts with the same terms.
2. In Newspack Campaigns, publish an archive prompt that should display after 1 post in category archives. Don't edit the Post Types options (but do observe that it shows listing post types, but they're not selected by default).
3. On `master`, observe that the prompt is not displayed on the category archive (because the first post shown is a listing, not a post).
4. Check out this branch, confirm that the archive prompt and any new prompt now have listing post types selected by default.
5. Confirm that the category archive now shows the prompt after the first post.

#### Bonus fix

1. Publish an inline prompt and make sure "Page" is selected as one of the supported post types for the prompt (it should be on by default).
2. Confirm that the prompt gets displayed on pages.
3. De-select "Page" from the prompt's supported post types.
4. Confirm that the prompt no longer gets displayed on pages, but continues to get displayed on other supported post types.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -811,14 +811,6 @@ final class Newspack_Popups_Inserter {
 	 */
 	public static function should_display( $popup, $check_if_is_post = false ) {
 		$post_type = get_post_type();
-		// Inline prompts may not be rendered on pages, unless they are above-header.
-		if (
-			$check_if_is_post &&
-			'page' === $post_type &&
-			( Newspack_Popups_Model::is_inline( $popup ) && ! Newspack_Popups_Model::is_above_header( $popup ) )
-		) {
-			return false;
-		}
 
 		// Unless it's a preview request, perform some additional checks.
 		if ( ! Newspack_Popups::is_preview_request() ) {
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -507,9 +507,22 @@ final class Newspack_Popups_Model {
 	 * Get the default supported post types.
 	 */
 	public static function get_default_popup_post_types() {
+		// Any custom post type that is both public and has a post type archive.
+		$public_post_types = array_values(
+			get_post_types(
+				[
+					'has_archive' => true,
+					'public'      => true,
+				]
+			)
+		);
+
+		// Default 'post' and 'page' post types actually have 'is_archive' => false, but we still want them.
+		$public_post_types = array_merge( [ 'post', 'page' ], $public_post_types );
+
 		return apply_filters(
 			'newspack_campaigns_default_supported_post_types',
-			[ 'post', 'page' ]
+			$public_post_types
 		);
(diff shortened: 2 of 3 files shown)


miguelpeixe (changes requested):
If we expect Campaigns to work on every public post type with archive support, perhaps the [`get_post_types()` function](https://developer.wordpress.org/reference/functions/get_post_types/) is a more generic solution. What do you think?

I still think the filter is useful for odd cases, but the function arguments would be something like:

```php
get_post_types( array(
	'public'      => true,
	'has_archive' => true,
) );
```

dkoo:
> If we expect Campaigns to work on every public post type with archive support, perhaps the [`get_post_types()` function](https://developer.wordpress.org/reference/functions/get_post_types/) is a more generic solution. What do you think?

That's actually open for discussion. I implemented the filter to handle public post types that can appear alongside posts on _post_ archives (not just their own archives), which for our purposes are newsletters and listings. Defaulting to showing prompts on any post type archive would be one step further but would also mean we don't need the extra changes in the Newsletters and Listings plugins. @adekbadek @kariae what do you guys think?

yogeshbeniwal:
@dkoo If we can keep it generic for showing prompts on any post type archive, that will be better. Like some publishers has series archives etc. This will save individual PR to add more supported post types.

kariae:
I think having Campaigns working on all public post types with archive support is the way to go, so a 👍 for using `get_post_types`. In case we need an exclude filter we can add it the same as [we did for tags and categories](https://github.com/Automattic/newspack-popups/pull/641).

dkoo:
@Automattic/newspack-product cc @yogeshbeniwal Thanks to all for your input! I've implemented the filter with `get_post_types` as suggested, so now by default Campaigns will support posts, pages, and all public CPTs with post type archives. I kept the filter in place so it can still be used to remove or otherwise edit the list of post types, but this should eliminate the need to add support for CPTs in other plugins.

miguelpeixe (approved):

PR #732, opened 2021-12-11
perf: shorten query params for preview iframe URLs

dkoo:
### All Submissions:

Shortens the preview URLs used to render prompts in a preview iframe. This should help avoid a bug in Chrome where iframes with very long URLs sometimes refuse to resolve.

1. On a server-based test site (not a local environment), using Chrome, preview a single prompt from the editor screen. You may have to repeat several times, but eventually the preview modal will hang at the "Loading..." view.
2. Check out this branch, rebuild.
3. In a new Chrome session, attempt to replicate the bug in step 1 again. You shouldn't be able to, theoretically.

### Other information:


kariae (approved):
Looks good to me.

To help with the test of this, we can make our local site public (using ngrok or other), create a new prompt, and change its default meta parameters (e.g. background color, placement, included/excluded categories, ...) this will lengthen the preview URL of the prompt.

The fixes worked like a charm, we might need to find another solution in case we're planning to add more parameters to prompts in the future, but this is doing the trick for now.

One last thing @dkoo I think we'll need to add the fix in the core plugin https://github.com/Automattic/newspack-plugin/blob/master/assets/wizards/popups/index.js#L229 which will fix the issue when we preview the prompt from the prompts list screen. We need to merge both changes now since the preview on the prompts list screen won't be triggered.

dkoo:
> The fixes worked like a charm, we might need to find another solution in case we're planning to add more parameters to prompts in the future, but this is doing the trick for now.

This is true—if customizing the prompt's options extensively, it will add on more parameters and the URL might get too long even with the abbreviated keys.

> One last thing @dkoo I think we'll need to add the fix in the core plugin https://github.com/Automattic/newspack-plugin/blob/master/assets/wizards/popups/index.js#L229 which will fix the issue when we preview the prompt from the prompts list screen. We need to merge both changes now since the preview on the prompts list screen won't be triggered.

Yes, you're right. https://github.com/Automattic/newspack-plugin/pull/1366 should apply the shortened query params to the Campaigns wizard as well. Note that 72dee99 does a slight refactor which lets us share the same mapping for shortened query keys in both plugins.

One other thing I noticed while testing previewing from the Campaigns wizard: I have a lot of published and draft prompts on my test site (~30 or so). When previewing from the wizard, config for all published and draft prompts is included in the preview request, which makes for a really long preview URL. I was getting [`414 Request-URI Too Large`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/414) errors on the AMP Access request, so I added some filtering in the `insert_popups_amp_access` method to only check AMP Access for prompts that are actually being previewed.

kariae:
> One other thing I noticed while testing previewing from the Campaigns wizard: I have a lot of published and draft prompts on my test site (~30 or so). When previewing from the wizard, config for all published and draft prompts is included in the preview request, which makes for a really long preview URL. I was getting 414 Request-URI Too Large errors on the AMP Access request, so I added some filtering in the insert_popups_amp_access method to only check AMP Access for prompts that are actually being previewed.

Yes, I can confirm this now that you're raising it. https://github.com/Automattic/newspack-popups/commit/72dee99b7ded1445a0bc09c78aa9e09ffe7cbe72 seems to fix it, thanks.

kariae (approved):

PR #755, opened 2021-12-20
Insert prompts block count basis

### LINKED ISSUE #623: Block-count-based insertion
## As is

An inline prompt can be inserted based on percentage

## To be 

Inline prompts should be possible to insert at block-count positions, e.g. after 2nd block

kariae:
### All Submissions:

Closes #623.

1. Load this PR.
2. Create a new inline prompt.
3. Set the trigger to `Blocks count` from the right sidebar, and set the number of blocks to be rendered before inserting the prompt.
4. Load a post and observe the prompt is inserted after the number of blocks set in 2.
5. Smock test some edge cases, like having the prompt after a heading (it shouldn't be inserted), the prompt as the first/last element on the content.

### Other information:


src/editor/Sidebar.js
```diff
@@ -156,13 +161,35 @@ const Sidebar = ( {
...
-					label={ __( 'Approximate Position (in percent)', 'newspack-popups' ) }
-					value={ trigger_scroll_progress }
-					onChange={ value => onMetaFieldChange( 'trigger_scroll_progress', value ) }
-					min={ 0 }
-					max={ 100 }
-				/>
+				<>
+					<SelectControl
+						label={ __( 'Trigger', 'newspack-popups' ) }
+						help={ __( 'The event to trigger the prompt.', 'newspack-popups' ) }
```
adekbadek, line 167: This is not about an event, just an insertion point. 

```suggestion
						help={ __( 'The position at which to insert the prompt.', 'newspack-popups' ) }
```

src/editor/Sidebar.js
```diff
@@ -156,13 +161,35 @@ const Sidebar = ( {
...
-				<RangeControl
-					label={ __( 'Approximate Position (in percent)', 'newspack-popups' ) }
-					value={ trigger_scroll_progress }
-					onChange={ value => onMetaFieldChange( 'trigger_scroll_progress', value ) }
-					min={ 0 }
-					max={ 100 }
-				/>
+				<>
+					<SelectControl
+						label={ __( 'Trigger', 'newspack-popups' ) }
```
adekbadek, line 166: The overlay prompts are "triggered", since they are initially hidden. Inline prompts are just inserted into content.

```suggestion
						label={ __( 'Insertion position', 'newspack-popups' ) }
```

src/editor/Sidebar.js
```diff
@@ -156,13 +161,35 @@ const Sidebar = ( {
...
+						help={ __( 'The event to trigger the prompt.', 'newspack-popups' ) }
+						value={ trigger_type }
+						options={ [
+							{ label: __( 'Scroll Progress', 'newspack-popups' ), value: 'scroll' },
```
adekbadek, line 170: "Scroll progress" is more suitable in the case of overlay prompts, where the action of scrolling is triggering the prompt to be displayed. Here it's just about the insertion position.

```suggestion
							{ label: __( 'Percentage', 'newspack-popups' ), value: 'scroll' },
```

src/editor/utils.js
```diff
@@ -235,14 +237,25 @@ export const getPlacementHelpMessage = (
...
 		case 'inline':
-			return sprintf(
-				// Translators: Trigger percentage.
-				__(
-					'The prompt will be automatically inserted about %s into article content.',
-					'newspack-popups'
-				),
-				triggerPercentage + '%'
-			);
+			return triggerCount
```
adekbadek, line 240: This is not an accurate condition. When switching between percentage-insertion and block-count-insertion, `triggerCount` will stay the same and so the help text will not be updated:

image

The condition should be based on the `trigger_type` meta.

includes/class-newspack-popups.php
```diff
@@ -152,6 +153,18 @@ public static function register_meta() {
 			]
 		);
 
+		\register_meta(
+			'post',
+			'trigger_blocks_count',
```
adekbadek, line 158:
```suggestion
			// Not really a "trigger", since this meta applies only to inline prompts. Keeping the "trigger"-based naming for consistency.
			'trigger_blocks_count',
```

adekbadek (changes requested):
The "trigger"-based naming in the code is misleading, but this is caused by `trigger_scroll_progress` meta being reused by overlay & inline prompts. Ideally, the name of the new meta should be something more like `insertion_blocks_count`, but on the other hand the `trigger_*` meta names do form a group here. Just pointing this out for a possible future refactoring.

adekbadek (approved):

PR #771, opened 2022-01-12
fix: view.js conflict with mediaelement.js

### LINKED ISSUE #729: Campaigns removes Podcast Block play bar
When the Newspack Campaign plugin is active, it removes the play bar in the Podcast Block from front-end of the site and also seems to remove the Cover Art (this can be seen as a logged out viewer or in incognito). Disabling prompts for a page or post doesn't resolve the issue, the play bar will only reappear when the Campaigns plugin is deactivated.

How it appears when Campaigns plugin is active:
Markup 2021-12-09 at 12 45 54

How it appears when Campaigns plugin is deactivated:
Markup 2021-12-09 at 12 42 23

Reported by Atavist.

dkoo:
### All Submissions:

Does two things:

1. Fixes a conflict between this plugin's non-AMP `view.js` and the core `mediaelement-core` JS, when the latter is also enqueued on the page. Note that the podcast player block still doesn't seem to work when AMP is enabled, but this doesn't seem to be related to this plugin.
2. Avoids enqueuing front-end JS and CSS if the "Disable prompts on this post or page" option is enabled, since it won't be needed in that case.

Closes #729.

#### `mediaelement` conflict

1. Add a Jetpack Podcast Player to a post and configure it with a valid URL (try: https://feeds.megaphone.fm/no-place-like-home)
2. Disable AMP on the post.
3. View the post in an incognito window. On `master`, observe that the podcast player block is missing the media player component.
4. Check out this branch and refresh. Confirm that the media player loads correctly now, and that prompts still load as expected.

#### Avoid enqueuing unnecessary assets

1. On the post, check the "Disable prompts on this post or page" option.
2. View the post in an incognito window. On `master`, view source and observe that the `newspack-popups-view` JS and CSS files are still enqueued.
3. Check out this branch, refresh, confirm that those assets are no longer enqueued.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -482,6 +482,13 @@ final class Newspack_Popups_Inserter {
 		if ( defined( 'IS_TEST_ENV' ) && IS_TEST_ENV ) {
 			return;
 		}
+
+		// Don't enqueue assets if prompts are disabled on this post.
+		$has_disabled_prompts = is_singular() && ! empty( get_post_meta( get_the_ID(), 'newspack_popups_has_disabled_popups', true ) );
+		if ( $has_disabled_prompts ) {
+			return;
+		}
+
 		$is_amp = function_exists( 'is_amp_endpoint' ) && is_amp_endpoint();
 		if ( ! $is_amp ) {
 			wp_register_script(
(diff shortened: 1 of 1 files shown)


miguelpeixe (approved):
Fixes the issue and can confirm the assets are only loaded when necessary.

PR #777, opened 2022-01-18
feat: allow admins to define merge fields for donors

### LINKED ISSUE #684: Make the merge field we check for donation status in Mailchimp configurable per-newsroom
Currently we check for merge fields starting with "DONA" and this appears to be mostly relevant for sites using News Revenue Hub for donations.

Ideally this is configurable so that we're able to support other membership/donation/CRM vendors.

Asana: https://app.asana.com/0/1200942827698993/1201001433518020
Details: https://docs.google.com/document/d/1s83l3e_kwYnMziNjN4Fhju3lwlrKEMbX6eUuPpZrb7U/edit#heading=h.6k09b8jn8rie

dkoo:
*|### All Submissions:

Refactors the Campaigns Settings page to use the new `PluginSettings` component (requires https://github.com/Automattic/newspack-plugin/pull/1450).

Creates a new setting option to define a comma-delimited list of strings to look for when checking Mailchimp merge fields (instead of hardcoding `DONAT` as the only string to check for). Closes #684.

#### New Settings UI

1. Follow testing instructions in https://github.com/Automattic/newspack-plugin/pull/1450 to test the wizard UI experience.
2. Deactivate both `newspack-plugin` and `newspack-blocks` to enable the "fallback" settings UI.
3. You should now see a **Prompts > Settings** link in the WP admin dashboard sidebar menu. Click it.
4. Confirm that all options are available here, and that you now have to click "Save" to save your settings (previously the settings updated when an option value changes, but that doesn't work as well for text inputs like the new Mailchimp Donor Merge Fields option).
5. Change various options, save, and refresh the page to confirm that your settings are saved.

Screen Shot 2022-01-21 at 5 02 54 PM

#### Donor merge fields

This one's a bit involved to test. Before getting started, make sure you have the following prerequisites setup:

* A Mailchimp account with a list and at least one subscriber whose email address you have access to (DM me if you need access to a Mailchimp account to test with)
* A test site with Newspack Newsletters connected to and setup to send to said Mailchimp list
* A Campaigns segment set to "Subscribers" only, with at least one prompt to show to subscribers
* A Campaigns segment at the highest priority, set to "Donors" only, with at least one prompt to show to donors

Once you have the above, to test:

1. Make sure Newspack Plugin is checked out to https://github.com/Automattic/newspack-plugin/pull/1450.
2. Go to **Newspack > Campaigns > Settings**. Observe the new option to define Mailchimp merge tags, and that it defaults to `DONAT` for backwards compatibility:
 
Screen Shot 2022-01-21 at 1 53 26 PM

4. Update the field to something else (e.g. `HUB_` or `DONAT,HUB_`) and save. Refresh the page and confirm that your setting is saved.
5. In Mailchimp, go to **Audience > All contacts** and expand the **Settings** menu. Select **Audience fields and \*|MERGE|\* tags**.

Screen Shot 2022-01-21 at 1 56 19 PM

6. Click **Add a Field** at the bottom of the page. Choose "Text" as the field type. A new Untitled field will appear at the bottom of the UI.
7. Set the label to something identifiable and (important) the "Put this tag in your content" field to something that matches a string in the **Mailchimp merge fields** option you set in Campaigns settings, e.g.:

Screen Shot 2022-01-21 at 1 59 20 PM

(Note that in the above screenshot, the merge field label is `HUB_SUPPORTER` and the merge field name is `HUB_DONOR`—the latter is what's used to match to our plugin option.)

7. Save changes and go back to your All Contacts list view. Click on the name of a contact for which you have inbox access (so you can get the email when it's sent). Click "edit" in the **Profile Information** section and set the value of your new merge field to `true`. Save and exit.

Screen Shot 2022-01-21 at 4 43 23 PM

8. Back in your site, create a newsletter with a link to a post on your site, and send it to your Mailchimp list.
9. Find the newsletter email and copy the link to the post. It should look something like `https://blog.us10.list-manage.com/track/click?u=ae81335f8566492521d60e635&id=82da4b0fff&e=754835379c`.
10. Visit the link in a new incognito session. (Optional: add the `?newspack-campaigns-debug` flag to the URL to be able to view debug info on the request response in dev tools.)
11. On the first post you view, confirm that you see the prompt for subscribers (donation status is processed on that first page load, so you won't be considered a donor until subsequent page views).
12. Navigate to another post and confirm that you now see the prompt for donors.

### Other information:


aschweigert (approved):
This looks good to me! We'll want to make sure we fully cover how to set this up in our docs because it's presently a common source of confusion. Once this is ready to release, let me know and I'm happy to help out with that bit.

dkoo:
Nice, thanks for testing, @aschweigert! I'll still hold out for a code review from @Automattic/newspack-product before merging.

includes/class-newspack-popups-api.php
```diff
@@ -105,6 +102,45 @@ public function register_api_endpoints() {
...
 	}
 
+	/**
+	 * Recursively sanitize an array of arbitrary values.
+	 *
+	 * @param array $array Array to be sanitized.
+	 * @return array Sanitized array.
+	 */
+	public static function sanitize_array( $array ) {
+		foreach ( $array as $key => $value ) {
```
miguelpeixe, line 112: This loop will not update the values from the given array. This can either change to `&$value` or store the sanitized values in `$array[ $key ]`.

includes/class-newspack-popups-settings.php
```diff
@@ -60,6 +60,99 @@ public static function set_settings( $options ) {
...
+		switch ( $type ) {
+			case 'int':
+			case 'integer':
+			case 'boolean':
+				return (int) $value;
+			case 'float':
+				return (float) $value;
+			case 'string':
+				return sanitize_text_field( $value );
+			case 'select':
```
miguelpeixe, line 130: Nit: since `select` is determined by the existence of `options` and not by its type, should we keep this case?

miguelpeixe (changes requested):
That is a super cool feature! Excited to see the settings working on another UI context too 🎉 

With `newspack-plugin` and `newspack-blocks` deactivated I'm getting the following error when I uncheck and save any of the first two checkboxes. Weirdly, it's not the case when I toggle the third checkbox:

```
{code: 'newspack_popups_settings_error', message: 'Error updating the settings.', data: null}
```

I'd also expect to have the error message on the UI, I got a frozen state. Likely due to an unresolved loading state.

---

The donor merge fields appear to be working well, although I didn't get the donor prompt. It first prompted the subscriber prompt, then a generic prompt that I have for "everyone" and no other popup prompted. My setup is:

 - **Mailchimp merge donor field**: `HUB_`
 - Contact set with `HUB_SUPPORTER: true`

includes/class-newspack-popups-api.php
dkoo: Whoops, good catch! I forgot a line. Fixed in 3a2fb9e.

includes/class-newspack-popups-settings.php
dkoo: Guess it's not necessary—removed in ab48f5a.

dkoo:
> I'd also expect to have the error message on the UI, I got a frozen state. Likely due to an unresolved loading state.

21b921a adds error handling and disables the "Save" button while the Settings component is `inFlight` to avoid queuing requests.

> With `newspack-plugin` and `newspack-blocks` deactivated I'm getting the following error when I uncheck and save any of the first two checkboxes. 

Upon further testing, this is caused by a very specific situation. The `Suppress newsletter prompts for subscribers` and `Suppress newsletter prompts if dismissed` options are `true` by default, whereas the `Suppress donation prompts for donors` is `false` by default. So on a new site, the first two checkboxes are checked. Unchecking one of these checkboxes sends a request to `update_option` with a falsy value. However, because the option previously didn't exist, `update_option` returns `false` because the requested falsy value is equivalent to the non-existent value—and so it doesn't create or update the option. Then, the `set_settings` function returns an error if `update_option` returns `false`. 

I fixed this in 58d1d76 by calling `add_option` with the option name and value before `update_option`. If the option doesn't already exist, it will be created with the given value (even if falsy) and we can skip the `update_option` call entirely. If it does already exist, `add_option` will do nothing and return `false`, so we can safely call `update_option` instead.

miguelpeixe (changes requested):
> However, because the option previously didn't exist, update_option returns false because the requested falsy value is equivalent to the non-existent value—and so it doesn't create or update the option.

Ah, this is an annoying behavior of the `update_option()`. The way I solved it for a singular boolean setting is to [always cast it to integer](https://github.com/Automattic/newspack-popups/pull/777/files#diff-4a460f2519dbf25be336ce7612fd14b882b7486dc7d988266070ed97766d31bcR122-R135) so it's never handled as boolean on the wp option level.

~~I'm still getting the error with a disabled UI while unchecking the first two checkboxes, though:~~

image

~~The API error is:~~

```json
{"code":"rest_no_route","message":"No route was found matching the URL and request method.","data":{"status":404}}
```

**Nevermind!** I don't know why it happened again with the recent changes pulled and built. It's working as expected!

miguelpeixe (approved):

dkoo:
Aha, the answer was in the same class the whole time! 😆 9a058d3 updates the `set_settings` method to use the same `sanitize_setting_option` as the wizard UI, and 044d779 ensures `inFlight` is set to `false` whether updating the option is successful or not.

includes/class-newspack-popups-settings.php
```diff
@@ -50,16 +50,134 @@ public static function create_admin_page() {
...
+		$setting       = array_reduce(
+			$settings_list,
+			function( $acc, $config ) use ( $option_name ) {
+				if ( $option_name === $config['key'] ) {
+					$acc = $config;
+				}
+				return $acc;
+			},
+			false
+		);
```
miguelpeixe, line 64: I believe there's a duplicate effort here. Shouldn't the `get_setting_config()` method handle this? See the `update_setting()` method for reference on usage.

miguelpeixe (changes requested):

includes/class-newspack-popups-settings.php
dkoo: It is a bit redundant, but the difference is that without the main Newspack plugin and its components, we don't have a `$section` to pass to these existing methods. I'm trying to avoid doing a big refactor for the sake of this "non-Newspack-UI"  settings page, which should hopefully be seen by only a tiny fraction of OSS users who choose to use Campaigns without the main Newspack plugin.

miguelpeixe: Maybe rename the method to something like `set_setting_standalone()` and update its inline documentation so its purpose is more clear?

miguelpeixe (approved):

includes/class-newspack-popups-settings.php
dkoo: That's a good idea. Updated in 06bd589. To further avoid confusion I also updated some method names on the Newspack Plugin side: https://github.com/Automattic/newspack-plugin/pull/1450#issuecomment-1033176235

Could use one last review of both commits just to make sure it all makes sense.

PR #796, opened 2022-02-08
fix(editor): update meta values when custom fields panel is active

### LINKED ISSUE #25: Popup Sidebar Options Don't Update If Custom Fields Are Enabled
If custom fields are enabled in the post type options, updates to the sidebar fields do not stick. 

### Testing

1. Create a new Pop-up.
2. Click the `More Tools & Options` icon in the  upper right (three vertical dots).
3. Click Options. Check "Custom Fields" then Enable & Reload.
4. Make  changes to the Pop-up Settings in  the document sidebar.
5. Save the Pop-up and reload.
6. Observe the changes have been lost.

dkoo:
### All Submissions:

I'm actually unsure of whether this should even be fixed, because it seems to be somewhat intended behavior in the block editor.

To summarize: if you open the block editor's built-in Custom Fields panel by clicking the three-vertical-dots icon in the upper-right corner of the editor, then clicking Preferences, then enabling the Custom Fields panel as shown:

SXODCI

This enables a Custom Fields panel inside the block editor which shows all of the meta fields and their values for the post:

Screen Shot 2022-02-08 at 11 29 26 AM

However, these inputs seem to override any controls that edit the same meta fields via `dispatch` methods. If you change sidebar controls that update meta values, but don't update the same value in the Custom Fields panel, the unchanged Custom Fields value will take precedence when saving the post. Conversely, if you update the Custom Fields value and save the post, this value will get saved.

This proof-of-concept is kind of a hacky way we could override this default behavior, by watching for meta changes in the `dispatch` call and automatically updating the value of the matching Custom Fields input. However, there are problems with this approach:

* It's a hacky solution, and relies on specific markup structure in the editor that may be subject to change in future WP versions
* Solving it in Campaigns would mean that the behavior differs from that of other plugins/themes. The Custom Fields panel shows **all** meta fields registered for the post, not just those registered by this plugin. So we'd be updating our custom fields for Campaigns, but any other plugin's UI that updates its own meta fields won't do the same. A better solution would be to somehow hook into the `dispatch` function whenever any meta value is updated in the block editor (regardless of the plugin's source), but I'm not sure how to do this without deeper research, and it still seems to fix intended WP editor behavior, in my opinion.
* I somehow doubt that almost anyone would want to use this Custom Fields panel over our UI. And if they do, it seems like an opportunity to educate them on what seems like intended WP editor behavior.

@Automattic/newspack-product thoughts? I would prefer to mark this issue as "Do not fix" and treat it as intended (if unintuitive) behavior instead of as a bug.

Closes #25.

Verify that #25 can't be reproduced. When updating prompt options in the sidebar UI, confirm that values for those fields in the Custom Fields panel are updated automatically, and that they persist after saving and refreshing the editor.

### Other information:

### DIFF
--- a/src/editor/index.js
+++ b/src/editor/index.js
@@ -32,6 +32,19 @@ const mapDispatchToProps = dispatch => {
 	const { createNotice, removeNotice } = dispatch( 'core/notices' );
 	return {
 		onMetaFieldChange: ( key, value ) => {
+			const customFieldLabel = document.querySelector( `input[value="${ key }"` );
+			let customFieldValue;
+
+			if ( customFieldLabel ) {
+				customFieldValue = customFieldLabel.parentElement.nextElementSibling.querySelector(
+					'textarea'
+				);
+			}
+
+			if ( customFieldValue ) {
+				customFieldValue.value = value;
+			}
+
 			dispatch( 'core/editor' ).editPost( { meta: { [ key ]: value } } );
 		},
 		createNotice,


adekbadek (commented):
Is the "Custom fields" panel visible by default? If not, then I'd also opt for "won't fix".

dkoo:
> Is the "Custom fields" panel visible by default?

It's not, and it's hidden enough that you would need to be very intentional about turning it on.

dkoo:
I'm going to close this as "won't fix", per the above discussion.

PR #798, opened 2022-02-09
fix: don't strip URL query params if there's a hash in the URL

### LINKED ISSUE #745: Query params in URL stripped out if '#' in URL and non-AMP
On a site with Campaigns active and prompts, on non-AMP pages it looks like query params in the URL get stripped out a little too aggressively.

To reproduce:
1. Create a popup prompt on all pages
2. Log out, so that you get served popups
3. On a non-AMP page visit `the-current-non-amp-page.com/#test?foo=bar`. Observe that after the page is loaded the URL becomes `the-current-non-amp-page.com/#test`
4. On an AMP page visit `the-current-amp-page.com/#test?foo=bar`. Observe that the URL stays the same.

This appears to be [something with the cleanURL handling here](https://github.com/Automattic/newspack-popups/blob/5a087a2537869590308a01a5aea4239fa126c9f8/src/view/utils/index.js#L108-L113).

dkoo:
### All Submissions:

The [URI specification](https://www.rfc-editor.org/rfc/rfc3986#section-3) dictates that the query string in a URL must precede the hash (`#`) fragment. Accordingly, as soon as the `#` character is encountered in a URL, JavaScript will interpret everything after it as part of the fragment, even if it includes a query string. However, browsers are a bit more lenient about this spec, so in a URL like `https://url/#fragment?query=test&param` JS will see a hash fragment of `#fragment?query=test&param` and no query string, whereas the browser will see a hash fragment of `#fragment` and a query string of `?query=test&param`. We want the latter behavior, not the former.

Closes #745.

1. On `master`, visit a post with AMP off and with a hash fragment preceding a query string, e.g. `https://url/#fragment?query=test&param`. Observe that everything after the `#fragment` is stripped from the URL.
2. Check out this branch and revisit the URL with both fragment and query string. Confirm that the URL is restructured with the hash fragment coming after the query string, but with both still intact. (Note that boolean query params without a value may have an `=` appended, but this should have the same meaning/behavior as a query param without any value.)

### Other information:

### DIFF
--- a/src/view/utils/index.js
+++ b/src/view/utils/index.js
@@ -82,6 +82,13 @@ export const processFormData = ( data, formElement ) => {
 	return data;
 };
 
+// Get the hash from a URL without any query strings.
+const getHash = url => {
+	const hash = new URL( url ).hash.split( /\?|\&/ );
+
+	return hash[ 0 ];
+};
+
 /**
  * Given an amp-analytics configuration, a current url, and cookies,
  * retrieve client ID related linker param to be inserted into site cookies.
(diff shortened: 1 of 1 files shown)


adekbadek (approved):
Works well! I've added a couple of tests in 60a3d6432bafe5b1b7bd404da51cd2ec7b2b9c50 while trying to understand what that URL-cleaning was about (a year after writing it).

PR #806, opened 2022-02-17
fix: prevent default category on term deletion

### LINKED ISSUE #530: Prompts: handle removed categories
From https://github.com/Automattic/newspack-popups/pull/529#pullrequestreview-643621764

1. Apply a category as the only category to a prompt
2. Delete the category
3. Observe that the prompt is automatically assigned "Uncategorized" or whatever your site's default category is

### Expected

If the default category is the only category assigned to the prompt when the category is deleted, it shouldn't be applied.

dkoo:
### All Submissions:

When a category is deleted, prevents any prompts that had only that category from being assigned the default category.

Also slightly changes behavior in an [existing callback](https://github.com/Automattic/newspack-popups/blob/master/includes/class-newspack-popups.php#L951-L959) that happens on prompt publication to remove the default category only if it's the only category assigned to that prompt. This means it's possible to intentionally assign the default category to a prompt if it's assigned along with at least one other category, which could be a valid use case scenario (although maybe unlikely).

An alternative solution might be to always remove the default category from prompts, which would basically prevent it from ever being assigned to prompts, but that seemed heavy-handed to me. Thoughts?

Closes #530.

1. Create a new category and assign it to a new draft prompt. Make sure the prompt has no other categories assigned.
2. On `master`, delete the new category.
3. Observe that the prompt has the default category assigned automatically (usually "Uncategorized").
4. Check out this branch, repeat steps 1 and 2. Confirm that the prompt has no categories assigned.
5. Publish the prompt and refresh the editor. Confirm that the prompt still has no categories assigned.
6. Create another new category and assign it to the prompt along with the default category.
7. Delete the new category and confirm that the prompt still has the default category assigned.
8. Revert the prompt to draft status.
9. Ensure that the prompt has the default category and at least one other category assigned.
10. Publish the prompt, refresh the editor, and confirm that the prompt still has all assigned categories, including the default category.

### Other information:

### DIFF
--- a/includes/class-newspack-popups.php
+++ b/includes/class-newspack-popups.php
@@ -76,8 +76,8 @@ final class Newspack_Popups {
 			add_action( 'customize_controls_enqueue_scripts', [ __CLASS__, 'enqueue_customizer_assets' ] );
 			add_filter( 'display_post_states', [ __CLASS__, 'display_post_states' ], 10, 2 );
 			add_action( 'save_post_' . self::NEWSPACK_POPUPS_CPT, [ __CLASS__, 'popup_default_fields' ], 10, 3 );
-			add_action( 'transition_post_status', [ __CLASS__, 'remove_default_category' ], 10, 3 );
-
+			add_action( 'transition_post_status', [ __CLASS__, 'prevent_default_category_on_publish' ], 10, 3 );
+			add_filter( 'delete_category', [ __CLASS__, 'prevent_default_category_on_term_delete' ] );
 			add_filter( 'show_admin_bar', [ __CLASS__, 'show_admin_bar' ], 10, 2 ); // phpcs:ignore WordPressVIPMinimum.UserExperience.AdminBarRemoval.RemovalDetected
 
 			include_once dirname( __FILE__ ) . '/class-newspack-popups-model.php';
@@ -942,20 +942,57 @@ final class Newspack_Popups {
 	}
 
 	/**
+	 * Remove the default category from the given post, if it's the only category applied to that post.
+	 *
+	 * @param int $post_id ID of the post.
+	 */
+	private static function remove_default_category( $post_id ) {
+		$default_category_id = (int) get_option( 'default_category', false );
+		$post_categories     = wp_get_post_categories( $post_id );
+		if ( 1 === count( $post_categories ) && reset( $post_categories ) === $default_category_id ) {
+			wp_remove_object_terms( $post_id, $default_category_id, 'category' );
+		}
+	}
+
+	/**
 	 * Prevent setting the default category when publishing.
 	 *
 	 * @param string $new_status New status.
 	 * @param string $old_status Old status.
 	 * @param bool   $post Post.
 	 */
-	public static function remove_default_category( $new_status, $old_status, $post ) {
+	public static function prevent_default_category_on_publish( $new_status, $old_status, $post ) {
 		if ( self::NEWSPACK_POPUPS_CPT === $post->post_type && 'publish' !== $old_status && 'publish' === $new_status ) {
-			$default_category_id = (int) get_option( 'default_category', false );
-			$popup_has_category  = has_category( $default_category_id, $post->ID );
-			if ( $popup_has_category ) {
-				wp_remove_object_terms( $post->ID, $default_category_id, 'category' );
-			}
+			self::remove_default_category( $post->ID );
 		}
+	}
+
+	/**
+	 * When a category is deleted, any posts that have only that category assigned
+	 * are automatically assigned the site's default category (usually "Uncategorized").
+	 * We want to prevent this behavior for prompts, as prompts with the default
+	 * category will only appear on posts with that category.
+	 *
+	 * @return int The number of prompts affected by this callback.
+	 */
+	public static function prevent_default_category_on_term_delete() {
+		$prompts_updated          = 0;
+		$default_category_id      = (int) get_option( 'default_category', false );
+		$default_category_prompts = get_posts(
+			[
+				'cat'            => $default_category_id,
+				'post_status'    => 'any',
+				'post_type'      => self::NEWSPACK_POPUPS_CPT,
+				'posts_per_page' => 100, // Assumes a site won't have more than 100 prompts.
+			]
+		);
+
+		foreach ( $default_category_prompts as $prompt ) {
+			self::remove_default_category( $prompt->ID );
+			$prompts_updated ++;
+		}
+
+		return $prompts_updated;
 	}
 
 	/**
--- a/includes/class-newspack-popups.php
+++ b/includes/class-newspack-popups.php
@@ -984,7 +984,11 @@ final class Newspack_Popups {
 			return false;
 		}
 
-		$default_category_id           = (int) get_option( 'default_category', false );
+		$default_category_id = (int) get_option( 'default_category', false );
+		if ( false === $default_category_id ) {
+			return false;
+		}
+
 		$prompts_with_deleted_category = get_posts(
 			[
 				'category__in'     => $deleted_term,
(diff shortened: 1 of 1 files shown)


includes/class-newspack-popups.php
```diff
@@ -941,23 +941,60 @@ public static function archive_campaign( $id, $status ) {
...
+	 */
+	public static function prevent_default_category_on_term_delete() {
+		$prompts_updated          = 0;
+		$default_category_id      = (int) get_option( 'default_category', false );
+		$default_category_prompts = get_posts(
+			[
+				'cat'            => $default_category_id,
+				'post_status'    => 'any',
+				'post_type'      => self::NEWSPACK_POPUPS_CPT,
+				'posts_per_page' => 100, // Assumes a site won't have more than 100 prompts.
```
miguelpeixe, line 986: If the goal is to fetch all prompts, why not use [`-1`](https://developer.wordpress.org/reference/classes/wp_query/#pagination-parameters)?

miguelpeixe (changes requested):
I failed to reproduce step 7 of the instructions. The deletion of the category when a prompt has been previously set with that category **and** the default one resulted in a prompt without any categories.

If the hook that removes the object terms is executed before the deletion of the category, perhaps a more specific`tax_query` could avoid this behavior?

includes/class-newspack-popups.php
dkoo: I was taught to never use `-1` in case there are a crazy number of results, but it's probably not really a concern when it comes to prompts. Updated in f4ecf1b.

dkoo:
> I failed to reproduce step 7 of the instructions. The deletion of the category when a prompt has been previously set with that category **and** the default one resulted in a prompt without any categories.

Hmm, I couldn't replicate this behavior. When I deleted one category, the other remained assigned. But does 8ddcceb fix the issue on your end?

includes/class-newspack-popups.php
miguelpeixe: It's a valid concern, but since it's executed on a very specific and non-regular occasion (the deletion of a category), I think the eventual slow query is better than the unexpected behavior of not querying all prompts.

miguelpeixe:
I'm still getting the same result, my suggestion on the use of an advanced `tax_query` is so that you exclude posts that had the deleted category. That would only work if this hook is executed prior to the term relation deletion, though.

Here are my steps:

1. Create draft prompt with `Uncategorized` and recently created `My Prompt Category`
2. Delete `My Prompt Category`
3. Confirm that the prompt now has no categories when it should've remained with `Uncategorized`

I can reproduce that with any prompt that is intentionally categorized as `Uncategorized`, after the deletion of a category they all get the term relation removed.

includes/class-newspack-popups.php
```diff
@@ -941,21 +941,62 @@ public static function archive_campaign( $id, $status ) {
...
+		$prompts_updated          = 0;
+		$default_category_id      = (int) get_option( 'default_category', false );
+		$default_category_prompts = get_posts(
+			[
+				'post_status'    => 'any',
+				'post_type'      => self::NEWSPACK_POPUPS_CPT,
+				'posts_per_page' => -1,
+				'tax_query'      => [ // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_tax_query
+					'taxonomy'         => 'category',
+					'terms'            => $default_category_id,
```
miguelpeixe, line 988:
```suggestion
					'terms'            => [ $default_category_id, $deleted_category_id ],
```

I did not test this, but it's somewhat the idea I had so it doesn't remove the term relation of every prompt with the default category.

miguelpeixe (changes requested):

includes/class-newspack-popups.php
dkoo: Unfortunately this won't work because the [`delete_{$taxonomy}`](https://developer.wordpress.org/reference/hooks/delete_taxonomy/) hook fires _after_ the term is deleted, so running a `WP_Query` with its term ID will return every post.

dkoo: Looks like we can use [`pre_delete_term`](https://developer.wordpress.org/reference/hooks/pre_delete_term/) instead: I'll try that!

dkoo:
@miguelpeixe Thanks for the replication steps! That helped clarify to me what you were seeing, which I was able to reproduce on my end. I think 778d7b6 should handle this instance. It only removes the default category when added to posts that both had the deleted category and did NOT have the default category already. I couldn't think of a cleaner way to do this without the anonymous closure, so open to ideas for further improvement.

includes/class-newspack-popups.php
```diff
@@ -941,23 +941,90 @@ public static function archive_campaign( $id, $status ) {
...
+	 * @return int The number of prompts affected by this callback.
+	 */
+	public static function prevent_default_category_on_term_delete( $deleted_term, $taxonomy ) {
+		// We only care about categories.
+		if ( 'category' !== $taxonomy ) {
+			return false;
+		}
+
+		$default_category_id = (int) get_option( 'default_category', false );
+		if ( false === $default_category_id ) {
```
miguelpeixe, line 988: This variable is being casted to `int` so it'll never be `false` for this check. Maybe `empty()` is a safe assumption?

includes/class-newspack-popups.php
```diff
@@ -76,8 +76,8 @@ public function __construct() {
 			add_action( 'customize_controls_enqueue_scripts', [ __CLASS__, 'enqueue_customizer_assets' ] );
 			add_filter( 'display_post_states', [ __CLASS__, 'display_post_states' ], 10, 2 );
 			add_action( 'save_post_' . self::NEWSPACK_POPUPS_CPT, [ __CLASS__, 'popup_default_fields' ], 10, 3 );
-			add_action( 'transition_post_status', [ __CLASS__, 'remove_default_category' ], 10, 3 );
-
+			add_action( 'transition_post_status', [ __CLASS__, 'prevent_default_category_on_publish' ], 10, 3 );
+			add_filter( 'pre_delete_term', [ __CLASS__, 'prevent_default_category_on_term_delete' ], 10, 2 );
```
miguelpeixe, line 80: [`pre_delete_term`](https://developer.wordpress.org/reference/hooks/pre_delete_term/) is an action hook

includes/class-newspack-popups.php
```diff
@@ -941,23 +941,90 @@ public static function archive_campaign( $id, $status ) {
 		update_term_meta( $id, self::NEWSPACK_POPUPS_TAXONOMY_STATUS, $status ? 'archive' : '' );
 	}
 
+	/**
+	 * Remove the default category from the given post, if it's the only category applied to that post.
+	 *
+	 * @param int $post_id ID of the post.
+	 */
+	private static function remove_default_category( $post_id ) {
+		$default_category_id = (int) get_option( 'default_category', false );
```
miguelpeixe, line 950: Casting to `int` with a `$default` set to `false` looks a bit confusing.

includes/class-newspack-popups.php
```diff
@@ -941,23 +941,90 @@ public static function archive_campaign( $id, $status ) {
...
+		if ( false === $default_category_id ) {
+			return false;
+		}
+
+		$prompts_with_deleted_category = get_posts(
+			[
+				'category__in'     => $deleted_term,
+				'category__not_in' => $default_category_id, // We don't want to remove the default category if it was intentionally added.
+				'fields'           => 'ids',
```
miguelpeixe, line 996: TIL!

miguelpeixe (changes requested):
Nice find on the `pre_delete_term` action! Thank you for revising!

It's working as expected now, just a few requests.

includes/class-newspack-popups.php
dkoo: Good call, updated the default to `0` in 692ac31.

includes/class-newspack-popups.php
dkoo: Ack, thanks! Updated in 692ac31. I was also wondering why PHPCS was telling me that callback needed a return value. 😆 I removed those in df0ddef as it doesn't need to return anything.

includes/class-newspack-popups.php
dkoo: Good call, updated in 692ac31.

miguelpeixe (approved):
🙌

PR #807, opened 2022-02-17
fix: count inner blocks' content toward post length

### LINKED ISSUE #659: Inline placement at the top & large group
## To reproduce

1. Create an inline prompt to be inserted at 0%
2. Create a post consisting of a group block containing 10 paragraphs
3. Observe that the prompt is inserted _after_ the group

## Expected result

The prompt should be inserted before _any_ post content

dkoo:
### All Submissions:

Fixes an edge case that happens if an inline prompt with 0% scroll position is placed in a post consisting of only a single Group block. Also improves the insertion behavior surrounding blocks that contain inner blocks (not just Group blocks) so that the inner blocks' content is counted toward the block's total length. In practical terms this means that "container" blocks like Group and Column blocks will now have a more accurate content length when it comes to inserting inline prompts around them. These "container" blocks are still treated as single blocks, so prompts won't be inserted in between their inner blocks, which gives editors a way to ensure that prompts won't be inserted between certain content.

On top of this, also adds a failsafe condition where if the position for an inline prompt is set to `0`, it will be inserted first regardless of other conditions.

Closes #659.

1. Create an inline prompt with "Insertion position" set to **Percentage** and "Approximage position" set to **0**.
2. Create a post whose content is only a single Group block with any number of blocks inside. Ideally, include some nested Group blocks with their own content. You can use the following test content if needed:

<details>
<summary>Expand for test content</summary>

```html
<!-- wp:group -->
<div class="wp-block-group"><!-- wp:paragraph -->
<p>Paragraph 1</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph -->
<p>Paragraph 2</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph -->
<p>Paragraph 3</p>
<!-- /wp:paragraph -->

<!-- wp:group -->
<div class="wp-block-group"><!-- wp:heading -->
<h2 id="inner-group">Inner group</h2>
<!-- /wp:heading -->

<!-- wp:paragraph -->
<p>Inner Paragraph 1</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph -->
<p>Inner Paragraph 2</p>
<!-- /wp:paragraph --></div>
<!-- /wp:group -->

<!-- wp:paragraph -->
<p>Paragraph 4</p>
<!-- /wp:paragraph -->

<!-- wp:paragraph -->
<p>Paragraph 5</p>
<!-- /wp:paragraph --></div>
<!-- /wp:group -->
```
</details>

3. On `master`, observe that the prompt appears after the Group block.
4. On this branch, confirm that it appears before the Group block.
5. Add some other content around the outermost Group block and smoke test inline prompt insertion with various values. Note that prompt insertion behavior will be a bit different now because we're counting inner blocks' content toward the total post length—if anything the behavior should be more visually accurate.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -207,6 +207,45 @@ final class Newspack_Popups_Inserter {
 	}
 
 	/**
+	 * Get content from a given block's inner blocks, and recursively from those blocks' inner blocks.
+	 *
+	 * @param object $block A block.
+	 *
+	 * @return string The block's inner content.
+	 */
+	public static function get_inner_block_content( $block ) {
+		$inner_block_content = '';
+
+		if ( 0 < count( $block['innerBlocks'] ) ) {
+			foreach ( $block['innerBlocks'] as $inner_block ) {
+				$inner_block_content .= $inner_block['innerHTML'];
+
+				// Recursively get content from nested inner blocks.
+				if ( 0 < count( $inner_block['innerBlocks'] ) ) {
+					$inner_block_content .= self::get_inner_block_content( $inner_block );
+				}
+			}
+		}
+
+		return $inner_block_content;
+	}
+
+	/**
+	 * Get content from given block, including content from the block's inner blocks, if any.
+	 *
+	 * @param object $block A block.
+	 *
+	 * @return string The block's content.
+	 */
+	public static function get_block_content( $block ) {
+		$is_classic_block = null === $block['blockName'] || 'core/freeform' === $block['blockName']; // Classic block doesn't have a block name.
+		$block_content    = $is_classic_block ? force_balance_tags( wpautop( $block['innerHTML'] ) ) : $block['innerHTML'];
+		$block_content   .= self::get_inner_block_content( $block );
+
+		return $block_content;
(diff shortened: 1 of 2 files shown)


includes/class-newspack-popups-inserter.php
```diff
@@ -334,18 +372,18 @@ function ( $block_groups, $block ) use ( &$block_index, $parsed_blocks, $max_ind
...
+					continue;
+				}
+
+				$position          = $inline_popup['precise_position'];
+				$trigger_type      = $inline_popup['options']['trigger_type'];
+				$insert_at_zero    = 0 === $position; // If the position is 0, the prompt should always appear first.
+				$insert_for_scroll = 'scroll' === $trigger_type && $pos > $position;
+				$insert_for_blocks = 'blocks_count' === $trigger_type && $block_index >= $position;
+
+				if ( $insert_at_zero || $insert_for_scroll || $insert_for_blocks ) {
```
miguelpeixe, line 386: So much easier to read! 👌

miguelpeixe (approved):
Working as described! 🙌

PR #809, opened 2022-02-23
[HOLD] feat: default segments

### LINKED ISSUE #673: Provide base campaign segments (or just good docs) by user journey
Asana: https://app.asana.com/0/1200942827698993/1201001433518149/f
Details: https://docs.google.com/document/d/1yaRwAQirSrCyNdjqlyGdeALLwrNfrJEkZsKvwM66e_w/edit#heading=h.wzhi18vbtekt

Documentation would help triage this issue, but I think having some default segments would be a better way to go. 

I generally agree with the basic set of segments Brian outlined in the doc:

> Donors
> Frequent readers
> Newsletter subscribers
> Referred by [Smartnews, or another aggregator, etc.]
> Everyone else

We should make sure we have a solid understanding of who would fall into these segments and then build at least a few default segments directly into the plugin and think about how best to guide new publishers to setting these up in a way that will work well in most cases.

dkoo:
### All Submissions:

Starts the "reasonable defaults" part of the roadmap. Allows site admins to generate a set of default segments with a button click. If they've already been created and edited, clicking the button will restore the default segments to the default configuration.

The reason for requiring some kind of user action (in this case a button click) is to accommodate sites that may have been up and running with Campaigns for a while now—we don't want to automatically generate new segments which may conflict with existing segments that are running live prompts.

Closes #673.

Before testing, make sure your Newspack Plugin is running https://github.com/Automattic/newspack-plugin/pull/1524.

1. On a fresh or existing site, visit **Newspack > Campaigns > Segments**. Confirm there's a new button at the bottom of the segments list:

Screen Shot 2022-02-22 at 5 04 31 PM

2. Click the button. Confirm you see a modal with a bit of explanation of what you're doing (@aschweigert: could use your help with reviewing/updating this copy):

Screen Shot 2022-02-22 at 5 15 15 PM

3. Click "OK". Confirm that a set of default segments (with prefix "[Newspack]") is generated. If you already had existing segments on the site, confirm that the default segments are generated with lower priority than the existing segments.
4. Edit some of the default segment names and settings, and reshuffle the priority. Delete at least one default segment.
5. Click "Reset default segments" and click "OK". Confirm that:
  - All existing segments retain their sorted priority
  - The default segment(s) you edited are now reverted to their original names and settings
  - The default segment(s) you deleted are restored

### Other information:

### DIFF
--- a/includes/class-newspack-popups-segmentation.php
+++ b/includes/class-newspack-popups-segmentation.php
@@ -509,9 +509,9 @@ final class Newspack_Popups_Segmentation {
 	 * Create a segment.
 	 *
 	 * @param object  $segment A segment.
-	 * @param boolean $get_id If true, return only the created segment ID. Otherwise, return all segments.
+	 * @param boolean $return_id If true, return the created segment ID. Otherwise, return all segments.
 	 */
-	public static function create_segment( $segment, $get_id = false ) {
+	public static function create_segment( $segment, $return_id = false ) {
 		$segments              = self::get_segments();
 		$segment['id']         = uniqid();
 		$segment['created_at'] = gmdate( 'Y-m-d' );
@@ -520,7 +520,7 @@ final class Newspack_Popups_Segmentation {
 
 		update_option( self::SEGMENTS_OPTION_NAME, $segments );
 
-		if ( $get_id ) {
+		if ( $return_id ) {
 			return $segment['id'];
 		}
 
--- a/includes/class-newspack-popups-segmentation.php
+++ b/includes/class-newspack-popups-segmentation.php
@@ -426,25 +426,25 @@ final class Newspack_Popups_Segmentation {
 		$default_segments = get_option( self::DEFAULT_SEGMENTS_OPTION_NAME, [] );
 		$default_options  = [
 			'donors'           => [
-				'name'          => __( '[Newspack] Donors', 'newspack-popups' ),
+				'name'          => __( ' Donors', 'newspack-popups' ),
 				'configuration' => [
 					'is_donor' => true,
 				],
 			],
 			'frequent_readers' => [
-				'name'          => __( '[Newspack] Frequent Readers', 'newspack-popups' ),
+				'name'          => __( ' Frequent Readers', 'newspack-popups' ),
 				'configuration' => [
 					'min_posts' => 5,
 				],
 			],
 			'subscribers'      => [
-				'name'          => __( '[Newspack] Newsletter Subscribers', 'newspack-popups' ),
+				'name'          => __( ' Newsletter Subscribers', 'newspack-popups' ),
 				'configuration' => [
 					'is_subscribed' => true,
 				],
 			],
 			'everyone_else'    => [
-				'name'          => __( '[Newspack] Everyone Else', 'newspack-popups' ),
+				'name'          => __( ' Other Readers', 'newspack-popups' ),
 				'configuration' => [
 					'is_not_donor'      => true,
 					'is_not_subscribed' => true,
--- a/includes/class-newspack-popups-segmentation.php
+++ b/includes/class-newspack-popups-segmentation.php
@@ -36,6 +36,28 @@ final class Newspack_Popups_Segmentation {
 	const SEGMENTS_OPTION_NAME = 'newspack_popups_segments';
 
 	/**
+	 * Default config for a segment.
+	 */
+	const DEFAULT_SEGMENT_CONFIG = [
+		'min_posts'           => 0,
+		'max_posts'           => 0,
+		'min_session_posts'   => 0,
+		'max_session_posts'   => 0,
+		'is_subscribed'       => false,
+		'is_donor'            => false,
+		'is_not_subscribed'   => false,
+		'is_not_donor'        => false,
+		'favorite_categories' => [],
+		'referrers'           => '',
+		'referrers_not'       => '',
+	];
+
+	/**
+	 * Name of the option to store default segment IDs.
+	 */
+	const DEFAULT_SEGMENTS_OPTION_NAME = 'newspack_popups_default_segments';
+
+	/**
 	 * Main Newspack Segmentation Plugin Instance.
 	 * Ensures only one instance of Newspack Segmentation Plugin Instance is loaded or can be loaded.
 	 *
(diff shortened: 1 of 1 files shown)


includes/class-newspack-popups-segmentation.php
```diff
@@ -424,16 +508,22 @@ function( $segment ) {
 	/**
 	 * Create a segment.
 	 *
-	 * @param object $segment A segment.
+	 * @param object  $segment A segment.
+	 * @param boolean $get_id If true, return only the created segment ID. Otherwise, return all segments.
```
adekbadek, line 512: It's weird that this method is returning all segments, and not the created segment. But refactoring it is maybe not worth the effort. 
Nit:
```suggestion
	 * @param boolean $return_id If true, return the created segment ID. Otherwise, return all segments.
```

adekbadek (changes requested):
> The reason for requiring some kind of user action (in this case a button click) is to accommodate sites that may have been up and running with Campaigns for a while now—we don't want to automatically generate new segments which may conflict with existing segments that are running live prompts.

What if there are no segments on the site – should defaults be set up then? This would be a surprise in the "Campaigns" tab, but there would be no disruption to the existing campaigns setup. WDYT @dkoo @aschweigert ? 

image

dkoo:
> What if there are no segments on the site – should defaults be set up then? This would be a surprise in the "Campaigns" tab, but there would be no disruption to the existing campaigns setup. WDYT @dkoo @aschweigert ?

I can get behind this. Maybe if there are no segments AND no active prompts? This would avoid creating confusion for sites that may have active prompts running for "everyone" and don't intend to use the segmentation features.

includes/class-newspack-popups-segmentation.php
```diff
@@ -386,6 +408,68 @@ function( $segment ) {
...
+	/**
+	 * Generate default segments. If they've already been generated, rebuild them with default parameters.
+	 *
+	 * @return array Array of all segments.
+	 */
+	public static function create_default_segments() {
+		$default_segments = get_option( self::DEFAULT_SEGMENTS_OPTION_NAME, [] );
+		$default_options  = [
+			'donors'           => [
+				'name'          => __( '[Newspack] Donors', 'newspack-popups' ),
```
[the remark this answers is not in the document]
dkoo: Sounds good to me—I felt it might be useful to be able to differentiate visually, but it's probably not important to do so. Updated in 2c4d923.

includes/class-newspack-popups-segmentation.php
```diff
@@ -386,6 +408,68 @@ function( $segment ) {
...
+				],
+			],
+			'subscribers'      => [
+				'name'          => __( '[Newspack] Newsletter Subscribers', 'newspack-popups' ),
+				'configuration' => [
+					'is_subscribed' => true,
+				],
+			],
+			'everyone_else'    => [
+				'name'          => __( '[Newspack] Everyone Else', 'newspack-popups' ),
```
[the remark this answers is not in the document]
dkoo: Good call—updated to "Other Readers" in 2c4d923.

includes/class-newspack-popups-segmentation.php
```diff
@@ -386,6 +408,68 @@ function( $segment ) {
...
+				'configuration' => [
+					'is_not_donor'      => true,
+					'is_not_subscribed' => true,
+				],
+			],
+		];
+
+		foreach ( $default_options as $key => $segment ) {
+			// Ensure all config objects contain all option keys.
+			$segment['configuration'] = wp_parse_args( $segment['configuration'], self::DEFAULT_SEGMENT_CONFIG );
```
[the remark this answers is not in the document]
dkoo: `canonize_segment` didn't work for me. I saw two issues:

1. The object returned by that method is cast as a `stdClass Object` instead of an array, so that causes the segments to be saved to the options table in a way that can't be read by the Campaigns wizard UI.
2. Even if we cast the object to an array before saving, it doesn't seem to actually save the config settings—all segments are saved with default settings.

I think we should leave this as-is for now rather than refactor `canonize_segment` to be used here, as that method is pretty important for other features and I don't want to cause any regressions there.

includes/class-newspack-popups-segmentation.php
dkoo: Updated the var name in debe5f6. I believe the method returns all segments because the Campaigns wizard UI expects all segments in the API response after creating a segment, and uses the response to re-render the segment list.

dkoo:
@adekbadek I pushed some changes to address your feedback. I think the auto-generation [we're discussing](https://github.com/Automattic/newspack-popups/pull/809#issuecomment-1049120275) can probably live in the main plugin if we trigger it upon rendering the Campaigns wizard.

includes/class-newspack-popups-segmentation.php
```diff
@@ -386,6 +408,68 @@ function( $segment ) {
...
+				],
+			],
+			'subscribers'      => [
+				'name'          => __( ' Newsletter Subscribers', 'newspack-popups' ),
+				'configuration' => [
+					'is_subscribed' => true,
+				],
+			],
+			'everyone_else'    => [
+				'name'          => __( ' Other Readers', 'newspack-popups' ),
```
adekbadek, line 447: Nit: there are extra leading spaces in the segments names:
```suggestion
				'name'          => __( 'Other Readers', 'newspack-popups' ),
```

includes/class-newspack-popups-segmentation.php
adekbadek: Yes, that's the reason. It's just unintuitive when looking at the function itself. But that's a nitpick.

includes/class-newspack-popups-segmentation.php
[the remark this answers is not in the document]
adekbadek: > Even if we cast the object to an array before saving, it doesn't seem to actually save the config settings—all segments are saved with default settings.

Casting to array _did_ work for me (also had to `require_once` the module at the top of the function, [like here](https://github.com/Automattic/newspack-popups/blob/debe5f6873c566ad801d25976a84fb2499552ab9/includes/class-newspack-popups-segmentation.php#L610)) –
 
```suggestion
			$segment['configuration'] = (array) Campaign_Data_Utils::canonize_segment( $segment['configuration'] );
```

I don't understand what's meant by "it doesn't seem to actually save the config settings" – the `Campaign_Data_Utils::canonize_segment` method is not supposed to save data, only process a configuration object.

Could you try again with the code here, unless it's exactly what you've already tried?

adekbadek:
> Maybe if there are no segments AND no active prompts? This would avoid creating confusion for sites that may have active prompts running for "everyone" and don't intend to use the segmentation features.

Adding the default segments won't disrupt their prompt delivery in any way, and I think the user should be encouraged to make use of segmentation. So I'd say the defaults should be created if there are no segments, regardless of whether there are prompts.

> we could trigger the auto-generation when you navigate to the Campaigns wizard, this way we could show a short message explaining what happened. And if a site admin never visits Campaigns, we don't need to run the extra functions.

Seems to me it'd be simpler to create the default segments on WP initialisation, without the extra API call. But it's not a strong opinion.

includes/class-newspack-popups-segmentation.php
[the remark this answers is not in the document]
dkoo: Strange, I thought this is exactly what I tried, but I must have done something wrong because it's now working. Previously, when I tried to implement it would save all of the segments with default settings values. Anyway, fixed in 86337ae.

includes/class-newspack-popups-segmentation.php
```diff
@@ -386,6 +401,70 @@ function( $segment ) {
...
+				],
+			],
+			'frequent_readers' => [
+				'name'          => __( ' Frequent Readers', 'newspack-popups' ),
+				'configuration' => [
+					'min_posts' => 5,
+				],
+			],
+			'subscribers'      => [
+				'name'          => __( ' Newsletter Subscribers', 'newspack-popups' ),
```
adekbadek, line 434: There are extra spaces also here and in other segment names ;) 

```suggestion
				'name'          => __( 'Newsletter Subscribers', 'newspack-popups' ),
```

includes/class-newspack-popups-segmentation.php
```diff
@@ -343,6 +349,15 @@ public static function create_database_table() {
 		}
 	}
 
+	/**
+	 * Create default segments for sites that don't have any existing segments.
+	 */
+	public static function new_site_setup() {
+		if ( empty( self::get_segments() ) ) {
```
adekbadek, line 356: Perhaps I'm overthinking this, but some users might just not want any segments. After the defaults are created, they'd delete those, and expect to have only the "Everyone" pseudo-segment in the UI. To make it possible, the check should be on an option that's saved when the segments are created for the first time. If they were created once, site should re-create the segments.

adekbadek (changes requested):

includes/class-newspack-popups-segmentation.php
dkoo: Gah! Fixed those in 794ac04!

includes/class-newspack-popups-segmentation.php
dkoo: I think you're right. 0aaae26 updates the function so that the default segments are created only if there are no segments AND the default segments haven't been created yet (i.e. the option doesn't exist).

dkoo:
Note to @thomasguillot: after some discussion we've decided we'd like to build this out into a more guided onboarding/wizard experience to guide users toward one of several sets of default segments, based on their input. @aschweigert will work on a suggested flow for this, but let's hold these changes until we can flesh that out a bit more.

adekbadek (approved):
Approving in the current state – as noted above it's likely to change, though.

dkoo:
After some internal discussion we're going to be moving in a different direction for this. I'm going to close this for now—we can reopen if we change our minds about wanting to release it as an interim step.

PR #813, opened 2022-02-28
fix: prevent non-page post types from appearing in settings dropdown

dkoo:
### All Submissions:

Ensures that the "Donor landing page" dropdown only shows pages, even if another plugin is runs a `pre_get_posts` callback to add additional post types. Generally, query alterations like this should be done in very specific scenarios, such as only on the front-end and only with specific query params, but occasionally some plugins alter the query without doing the necessary checks to ensure that the query isn't altered when it shouldn't be.

1. Temporarily add a callback to alter the query in your theme's `functions.php`. You can use any post type, but I'm using the Campaigns CPT in this example:

```php
function newspack_test_pre_get_posts( $query ) {
	$existing_post_types = $query->get( 'post_type' );

	if ( empty( $existing_post_types ) ) {
		$existing_post_types = [ 'post' ];
	}

	if ( ! is_array( $existing_post_types ) ) {
		$existing_post_types = [ $existing_post_types ];
	}

	if ( ! in_array( 'page', $existing_post_types, true ) ) {
		return;
	}

	// Add another post type to the query.
	$custom_post_type = [ 'newspack_popups_cpt' ];
	$post_types       = array_values( array_merge( $existing_post_types, $custom_post_type ) );
	$query->set( 'post_type', $post_types );
}
add_action( 'pre_get_posts', 'newspack_test_pre_get_posts' );
```

2. On `master`, visit the **Newspack > Campaigns > Settings** page. In the "Donor landing page" dropdown, observe that the options contain both pages and prompts (or whatever additional post type you added in step 1).
3.  Check out this branch and refresh. Confirm that the dropdown now only contains pages.
4. Temporarily disable the main Newspack Plugin. In the WP dashboard, visit **Prompts > Settings** and confirm that the dropdown here also only contains pages.

### Other information:

---
  - https://app.asana.com/0/0/1201796786147498

### DIFF
--- a/includes/class-newspack-popups-settings.php
+++ b/includes/class-newspack-popups-settings.php
@@ -200,20 +200,25 @@ class Newspack_Popups_Settings {
 	 * @return array Array of settings objects.
 	 */
 	public static function get_settings( $assoc = false, $get_segments = false ) {
-		$donor_landing_options       = [
+		$donor_landing_options = [
 			[
 				'label' => __( '-- None --', 'newspack-popups' ),
 				'value' => '',
 			],
 		];
+
+		// Before executing the query, make sure we can filter it to remove any CPTs that might be added by other plugins.
+		add_action( 'pre_get_posts', [ __CLASS__, 'prevent_other_post_types_in_page_query' ], PHP_INT_MAX );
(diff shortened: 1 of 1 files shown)


adekbadek (approved):
👌

PR #814, opened 2022-02-28
fix: only count donation products for donor segmentation

dkoo:
### All Submissions:

Currently, when segmenting readers as donors, we check if the reader is logged in and if they have any WooCommerce orders purchased under their user. However, we don't check whether those orders have anything to do with Newspack donations, so if a site is using WooCommerce to offer any other types of products, all customers are marked as donors whether or not they've actually donated.

This PR does an extra step of ensuring that the reader has purchased Newspack donation products before considering them donors. This won't fix things for readers who have already been marked donors, but it will fix things for new readers going forward.

1. On your test site, make sure your Reader Revenue platform is set to Newspack and that you have the required WooCommerce plugins installed and active.
2. Go to Products and create some additional products, separate from the Newspack Donate products.
3. Publish a prompt to a segment targeted to Donors only.
4. In a new incognito session, purchase one of the non-Newspack products (you may need to set up a shop page and use that in order to complete a purchase).
5. In the same session, browse through the site and observe that you see the prompt for Donors.
6. Check out this branch, and repeat steps 4-5 in a new incognito session.
7. This time, confirm that you don't see the prompt for Donors even after purchasing the non-Newspack product.
8. Complete a donation via a Newspack Donate block.
9. Browse through the site again and confirm that you see the prompt for Donors after purchasing the Newspack Donation product.

### Other information:

---
  - https://app.asana.com/0/0/1201693377396073

### DIFF
--- a/includes/class-newspack-popups-segmentation.php
+++ b/includes/class-newspack-popups-segmentation.php
@@ -248,7 +248,7 @@ final class Newspack_Popups_Segmentation {
 				if ( count( $user_orders ) ) {
 					$orders = [];
 					foreach ( $user_orders as $order ) {
-						$order_data = $order->get_data();
+						$order_data  = $order->get_data();
 						$order_items = array_map(
 							function( $item ) {
 								return $item->get_product_id();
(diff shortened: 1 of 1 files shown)


adekbadek (approved):
Works great!

PR #821, opened 2022-03-10
fix: prefix preview query param keys to avoid collisions

dkoo:
### All Submissions:

Fixes an issue with previewing single prompts from both the Campaigns wizard and block editor. Apparently the #732 shortened the query param keys used to render single prompt previews, but apparently one or more of these param keys was conflicting with something else in WordPress or another plugin. This PR adds an `n_` prefix to each query param to avoid such collisions.

This PR also proposes allowing prompts with categories and/or tags to appear on any post in the preview window—this might mitigate confusion when previewing a prompt and it doesn't appear in the window because the post loaded doesn't match the prompt's categories or tags. I'm open to discussing this behavior further if we think we should keep the current behavior, in which prompts with categories or tags will only appear on posts with matching terms within the preview window (thus sometimes requiring the user to navigate through the site within the preview window in order to find a post that will allow them to see the prompt).

1. On `master`, go to **Newspack > Campaigns** and attempt to preview any single prompt using the popover menu next to that prompt. Observe that for overlay prompts, you may see a strange XML error instead of the site loading within the preview; for inline prompts, the prompt just never appears.
2. Edit some prompts in the block editor and click "Preview". Observe the same behavior here.
3. Check out this branch and repeat steps 1–2. Confirm that the preview window loads the site and that the previewed prompts now appear as expected.

### Other information:


miguelpeixe (approved):
Can confirm this fixes the prompt preview!

Non-blocking suggestion, instead of adding `n_` to every parameter, why not contain the parameters inside a `newspack_prompt` array? Could also be reduced to something like `_np`:

```json
{
  "_np": {
    "bc": "#000"
  }
}
```

I wonder why the CI tasks did not trigger 🤔

dkoo:
> Non-blocking suggestion, instead of adding `n_` to every parameter, why not contain the parameters inside a `newspack_prompt` array? Could also be reduced to something like `_np`:

This is a good suggestion! However, it would require a concurrent fix in the main Newspack plugin, and I'd rather limit the scope of this fix since we've confirmed it works. If we start to see errors in Chrome previews again, your suggestion would be the first thing to try to further shorten the preview URLs.

dkoo:
> I wonder why the CI tasks did not trigger 🤔

Great question—maybe @adekbadek would know? I can confirm the tests and PHPCS pass locally.

PR #824, opened 2022-03-16
feat: show prompts to logged-in admin and editor users

### LINKED ISSUE #687: Make it easier for logged-in users to preview and QA prompts
UPDATE (3/9/22): per the discussion below and the many requests we've received from publishers, let's remove the logged-in admin check and make the behavior the same regardless of whether a user is logged in or not.

---

As @adekbadek noted, this might already be working fine but seems to be confusing so we'd probably want to start with some user-testing here and see if we can improve the segment preview UI to make it clearer how this is working.

Asana: https://app.asana.com/0/1200942827698993/1201001433518127
Details: https://docs.google.com/document/d/1yaRwAQirSrCyNdjqlyGdeALLwrNfrJEkZsKvwM66e_w/edit#heading=h.dt0seyczvadv

dkoo:
### All Submissions:

Changes prompt insertion logic so that logged-in admin and editor users see prompts just like regular readers. Like regular readers, admins and editors will only see prompts assigned to segments they match (so if an editor hasn't made any donations, they won't see donor-only prompts, for example).

Also prevents GA tracking code from firing for admins and editors, to hopefully prevent pollution of campaigns analytics data.

Closes #687.

1. Publish several prompts to several segments.
2. On `master`, as an admin user, browse the site's front-end. Observe that no prompts are displayed.
3. Check out this branch and browse again. Now confirm that prompts are displayed, as long as your reader activity matches their segments (so you may not see prompts shown to donors only, or to new readers).

#### Analytics

1. While browsing the site front-end as a logged-in admin user with this branch checked out, view source. Confirm that no `amp-analytics` tags are rendered with each prompt. This should prevent prompt interactions from being tracked while logged in.

### Other information:

### DIFF
--- a/includes/class-newspack-popups.php
+++ b/includes/class-newspack-popups.php
@@ -849,7 +849,15 @@ final class Newspack_Popups {
 	 * will not be fired for them.
 	 */
 	public static function is_user_admin() {
-		return is_user_logged_in() && current_user_can( 'edit_posts' );
+		/**
+		 * Filter to allow other plugins to decide which capability should be checked
+		 * to determine whether a user's activity should be tracked via Google Analytics.
+		 *
+		 * @param string $capability Capability to check. Default: edit_others_pages.
+		 * @return string Filtered capability string.
+		 */
+		$capability = apply_filters( 'newspack_popups_admin_user_capsbility', 'edit_others_pages' );
+		return is_user_logged_in() && current_user_can( $capability );
 	}
 
 	/**
--- a/includes/class-newspack-popups.php
+++ b/includes/class-newspack-popups.php
@@ -844,7 +844,7 @@ final class Newspack_Popups {
 	}
 
 	/**
-	 * Is the user an admin/editor/author/contributor user?
+	 * Is the user an admin or editor user?
 	 * If so, prompts will be shown to these users while logged in, but analytics
 	 * will not be fired for them.
 	 */
--- a/includes/class-newspack-popups.php
+++ b/includes/class-newspack-popups.php
@@ -856,7 +856,7 @@ final class Newspack_Popups {
 		 * @param string $capability Capability to check. Default: edit_others_pages.
 		 * @return string Filtered capability string.
 		 */
-		$capability = apply_filters( 'newspack_popups_admin_user_capsbility', 'edit_others_pages' );
+		$capability = apply_filters( 'newspack_popups_admin_user_capability', 'edit_others_pages' );
 		return is_user_logged_in() && current_user_can( $capability );
 	}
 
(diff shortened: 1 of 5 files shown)


yogeshbeniwal:
@dkoo Will it be better to not fire GA tracking code on basis of user capabilities i.e. User that can write post rather than just for admin and editor role. This will then cover authors/contributors/custom role which has post write capability and will be logged in most of the time on website while reading content.

dkoo:
@yogeshbeniwal Good suggestion. It's somewhat unclear to me whether this should apply to all Newspack sites—I'm not sure how all our different publishers use author/contributor users and whether they expect their interactions to be tracked or not. That said, it does seem likely that the default should be to not track for those users, so I updated in 2a3dd66.

cc @Automattic/newspack-product for your thoughts as well.

includes/class-newspack-popups.php
```diff
@@ -844,10 +844,12 @@ public static function api_config_missing_notice() {
...
 
 	/**
-	 * Is the user an admin user?
+	 * Is the user an admin/editor/author/contributor user?
+	 * If so, prompts will be shown to these users while logged in, but analytics
+	 * will not be fired for them.
 	 */
 	public static function is_user_admin() {
-		return is_user_logged_in() && current_user_can( 'edit_others_pages' );
+		return is_user_logged_in() && current_user_can( 'edit_posts' );
```
miguelpeixe, line 852: I wonder if we should skip analytics for anyone other than admin and editor. Listings customers, for example, have `edit_posts` capability.

miguelpeixe (approved):
Working as described! Only a non-blocking question, in case this has already been considered.

includes/class-newspack-popups.php
dkoo: I had the same concern. In the case of listings customers, it's a bit more complicated—they actually only have `edit_posts` when it comes to accessing the block editor, otherwise we block that capability for all other purposes (so that they don't have access to the WP admin dashboard or other parts of the site that `edit_posts` would normally grant). But I am concerned that custom user types on other sites may have `edit_posts` but should be tracked as readers.

What if we were to add a filter to allow other plugins to alter the capability we check to determine which users shouldn't be tracked?

dkoo: I think that's the way to go. For most sites `edit_others_pages` should be the right capability to check, but sites can choose to block analytics for a wider set of capabilities if needed. 2f8f3fe

miguelpeixe: Thank you for clarifying the listings customer mechanism!

> For most sites edit_others_pages should be the right capability to check, but sites can choose to block analytics for a wider set of capabilities if needed.

I agree that's the best solution!

miguelpeixe (approved):

PR #832, opened 2022-03-23
feat: add "undismissible" option for inline prompts

### LINKED ISSUE #822: Inline placements should not be dismissable
From Reveal: 

We expected the newsletter prompt on the mid-homepage to always appear, regardless if you're a subscriber, because it's targeted to the "everyone" group. But that's not what we see -- once you've used the prompt, it goes away and we just see a blue rectangle.

Here's our hypothesis: Once someone has used a prompt (eg signed up for our newsletter, or made a donation), it's considered "dismissed", and the *frequency* setting on the campaign prompt disables it.

If our hypothesis checks out, it would be great to have a new frequency setting different than "dismissed" for non-popup prompts, like "always" -- prompts that are baked into the UX of the site that don't go away after use.

Here's the blue rectangle we see where the midhomepage prompt should appear:

image

Here's what we expected to see on the midhomepage prompt:

image-1

dkoo:
### All Submissions:

Current UI allows editors to make "virtually" undismissible prompts by simply not entering any text in the Dismiss Button Text field. In this case, the prompt can't be dismissed with a button click, but it will still be suppressed based on other factors such as newsletter subscription, donation, etc. This is not ideal for publishers who include prompts as more or less static design elements in specific areas, such as on the homepage.

This PR adds a new option to explicitly make inline, Custom Placement, and Manual-Only prompts "undismissible" prompts. There's a new toggle switch, which is ON by default, which allows these prompts to be dismissed. This represents the status quo/current behavior, so sites already using prompts won't experience any change in behavior for existing prompts.

Screen Shot 2022-03-23 at 11 29 06 AM

Toggling this option OFF will make the prompt "undismissible," which results in the following behavior:

- The prompt will not display any "dismiss" button.
- The prompt will not follow the usual permanent suppression rules. For example, even if an undismissible prompt contains a newsletter signup form, and the reader has dismissed another newsletter prompt, the undismissible prompt will continue to be shown. I've added unit tests to cover all of these suppression cases.
- Even if a prompt was previously dismissed by the reader before this new option existed, if the new dismiss option is turned off the prompt will now continue to be shown.
- Undismissible prompts will still follow the rules of frequency and segmentation, as before. So an undismissible prompt with the "once" frequency will only be shown once, then will be suppressed as usual. An undismissible prompt targeted to a specific segment will still only be shown to readers matching that segment.
- Overlay prompts are always dismissible.

Closes #822.

1. Check out this branch and edit an existing inline, custom placement, or manual-only prompt. Expand the "Dismiss Button Settings" sidebar and confirm the new toggle option, which should be on by default for existing and new prompts.
2. Do a smoke test to confirm that existing prompts continue to behave as before, with no change in expected behavior. Make sure to keep one prompt around that has a newsletter signup block and is still dismissible with a visible dismiss button.
3. Create a new prompt for each inline placement (inline, custom placement, and manual), or edit the existing prompts. Make sure they contain newsletter signup blocks, and ensure that each will render on a particular page. Toggle OFF the new option to "allow this prompt to be dismissed" and confirm that the dismiss button fields (and preview in the block area) disappear. Save.
4. In a new incognito window, view the page where the undismissible prompts are shown. confirm that they're shown without a dismiss button.
5. Dismiss the dismissible newsletter prompt and refresh the page with the undismissible prompts. Confirm that the ones with newsletter signup blocks are still shown, even if your "Suppress newsletter prompts if dismissed" option is enabled.
6. If you like, you can repeat steps 4–5 but with `utm_medium=email` in the URL, Donate blocks for a donor reader, and with a UTM suppression value and confirm that the undismissible prompts are never automatically suppressed. These cases should all be covered by unit tests.
7. Confirm all unit tests are still passing. 👇 

### Other information:

### DIFF
--- a/api/campaigns/class-maybe-show-campaign.php
+++ b/api/campaigns/class-maybe-show-campaign.php
@@ -229,22 +229,25 @@ class Maybe_Show_Campaign extends Lightweight_API {
 		if ( false === $now ) {
 			$now = time();
 		}
+
 		$campaign_data      = $this->get_campaign_data( $client_id, $campaign->id );
 		$init_campaign_data = $campaign_data;
 
-		if ( $campaign_data['suppress_forever'] ) {
+		// Undismissible prompts are never automatically suppressed except by frequency settings.
+		$is_undismissible = isset( $campaign->u ) && $campaign->u;
+
+		if ( ! $is_undismissible && $campaign_data['suppress_forever'] ) {
 			self::add_suppression_reason( $campaign->id, __( 'Prompt dismissed permanently.', 'newspack-popups' ) );
 			return false;
 		}
 
-		$should_display = true;
-
-		$has_newsletter_prompt = $campaign->n;
+		$should_display        = true;
+		$has_newsletter_prompt = ! $is_undismissible && $campaign->n;
 		// Suppressing based on UTM Medium parameter in the URL.
 		$has_utm_medium_in_url = Campaign_Data_Utils::is_url_from_email( $referer_url );
 
 		// Handle referer-based conditions.
-		if ( ! empty( $referer_url ) ) {
+		if ( ! $is_undismissible && ! empty( $referer_url ) ) {
 			// Suppressing based on UTM Source parameter in the URL.
 			$utm_suppression = ! empty( $campaign->utm ) ? urldecode( $campaign->utm ) : null;
 			if ( $utm_suppression && stripos( urldecode( $referer_url ), 'utm_source=' . $utm_suppression ) ) {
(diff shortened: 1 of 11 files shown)


aschweigert (approved):
This seems like a reasonable compromise to me. We'll want to make sure we also update the docs to cover these options and make it clearer how/why a publisher might want to use them (the Reveal homepage example would be a good one to capture and explain once we get this fix implemented for them).

yogeshbeniwal:
> "Suppress newsletter prompts if dismissed"

@dkoo Now does it make more sense to update option in settings to permanent disable rather than dismissed only.

Screenshot

miguelpeixe:
@yogeshbeniwal help texts have been updated to only refer to "dismissable prompts":

image

miguelpeixe (approved):
Working as described! 🙌

PR #838, opened 2022-04-05
fix: report views when seen, not just interacted with

dkoo:
### All Submissions:

Apparently, prompt views are only being reported on `POST` requests, which means that prompts were only reported as viewed if the reader interacted with the prompt (e.g. by dismissing it). This caused prompts with the "once" or "once a day" frequency to not be properly suppressed after being viewed (but not dismissed) once.

1. On `master`, publish two inline prompts, one with the "Once" frequency and the other with the "Once a day" frequency.
2. In a new incognito session, navigate through several posts without dismissing the prompts. Observe that they appear on every post.
3. Check out this branch, repeat step 2, and confirm that the prompts are shown only once.

### Other information:

### DIFF
--- a/api/campaigns/class-maybe-show-campaign.php
+++ b/api/campaigns/class-maybe-show-campaign.php
@@ -152,6 +152,18 @@ class Maybe_Show_Campaign extends Lightweight_API {
 				}
 			}
 
+			// If prompt was shown, report a view.
+			if ( $campaign_should_be_shown ) {
+				$campaign_data = $this->get_campaign_data( $client_id, $campaign->id );
+				$campaign_data['count']++;
+				$campaign_data['last_viewed'] = time();
+				$client_data_update           = [
+					'prompts' => [ "$campaign->id" => $campaign_data ],
+				];
+
+				$this->save_client_data( $client_id, $client_data_update );
+			}
+
 			$response[ $campaign->id ] = $campaign_should_be_shown;
 		}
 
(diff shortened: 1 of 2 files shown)


adekbadek (commented):
I was not able to reproduce the issue – the inline "once"/"once a day" prompts are not displayed after being seen once. 

> Apparently, prompt views are only being reported on POST requests, which means that prompts were only reported as viewed if the reader interacted with the prompt

Not exactly – `POST` requests also are made [when the prompt enters the viewport](https://github.com/Automattic/newspack-popups/blob/f766dac3d0e15b89a1d2d92dc0827654bb1d70b6/includes/class-newspack-popups-model.php#L798-L813). The view reporting code is in [`Report_Campaign_Data::report_campaign` method](https://github.com/Automattic/newspack-popups/blob/f766dac3d0e15b89a1d2d92dc0827654bb1d70b6/api/campaigns/class-report-campaign-data.php#L44-L45).

miguelpeixe:
#824 implements visibility of prompts for authenticated users and also [disables the tracking code for admins and editors](https://github.com/Automattic/newspack-popups/pull/824/files#diff-c7efd1b032b601286ed42b644beb0443eaa33ce019c42dc068c443adaf1028fdR872-R882). Being tied to the tracked campaign report, could that be related?

adekbadek:
Ah, that's true – I did not yet get used to this change. #839 is an alternative proposition – this allows _internal_ tracking (which serves segmentation/frequency purposes) for admins, while keeping the analytics-related tracking disabled for admins.

dkoo:
Closing in favor of #839. After a bit more testing, I found that this PR fixes an issue which seemingly only happens in certain browsers.

PR #843, opened 2022-04-08
fix: hide empty paragraph blocks

thomasguillot:
### All Submissions:

Sometimes the Editor adds an empty paragraph block at the end of a page. When looking at the front-end we end up with an empty `<p/>` that adds some unwanted extra margin.

1. Create/Edit a prompt
2. Insert an empty Paragrpah block
3. Check the front-end, the empty `<p/>` should have some sort of margin
4. Switch to this branch and refresh
5. Check the front-end, the empty `<p/>` should not be an issue anymore

### Other information:

### DIFF
--- a/src/view/style.scss
+++ b/src/view/style.scss
@@ -373,6 +373,14 @@ $size__large: 1264px;
 	}
 }
 
+// Hide empty Paragraph Blocks
+.newspack-lightbox,
+.newspack-inline-popup {
+	p:empty {
+		display: none;
+	}
+}
+
 // This is needed for the scroll-triggered popups to function properly.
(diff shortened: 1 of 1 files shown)


miguelpeixe (changes requested):
I was not able to reproduce the issue with an extra margin due to empty paragraphs. I do get an unwanted margin with inline prompts when there's no dismiss button:

image

The example above presents the same margin on master and this PR, with or without an empty paragraph at the end.

The prompt could probably use a `*:last-child` margin rule for such cases.

Can you provide a screenshot of the extra margin you are seeing due to empty paragraphs?

PR #844, opened 2022-04-08
Add specific class to editor when prompt is inline

thomasguillot:
### All Submissions:

By adding a class of `is-placement-inline` to the Editor, we will able to provide specific CSS for some patterns, whether they are inline or not.

1. Switch to this branch
2. Check the list of classes for `block-editor-block-list__layout`
3. Toggle between inline and overlay (play with overlay sizes as well)

### Other information:

### DIFF
--- a/src/editor/EditorAdditions.js
+++ b/src/editor/EditorAdditions.js
@@ -62,18 +62,22 @@ const EditorAdditions = () => {
 		}
 	}, [ undismissible_prompt, dismiss_text, dismiss_text_alignment ] );
 
-	// Setting editor size as per the popup size.
+	// Setting editor size as per the overlay size or add inline placement.
 	useEffect( () => {
 		const blockEditor = document.querySelector( '.block-editor-block-list__layout' );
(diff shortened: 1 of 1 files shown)


src/editor/EditorAdditions.js
```diff
@@ -62,18 +62,22 @@ const EditorAdditions = () => {
...
-	// Setting editor size as per the popup size.
+	// Setting editor size as per the overlay size or add inline placement.
 	useEffect( () => {
 		const blockEditor = document.querySelector( '.block-editor-block-list__layout' );
 		if ( blockEditor ) {
 			blockEditor.classList.forEach( className => {
 				if ( className.startsWith( 'is-size-' ) ) {
 					blockEditor.classList.remove( className );
+				} else {
+					blockEditor.classList.remove( 'is-placement-inline' );
```
adekbadek, line 73: This can be done outside of the `blockEditor.classList` loop, without conditions (below line 68 for example). The intention will be more readable this way.

PR #846, opened 2022-04-12
fix: make "scroll" the default inline trigger_type

### LINKED ISSUE #842: Inline prompts with invalid insertion position
#755 added the ability to insert inline prompts by either content percentage or absolute block count. However, some prompts that pre-date this change seem to be stuck in a limbo state, where the `RangeControl` to select the percentage or block count isn't shown.

https://d.pr/i/ZykjXp

When this happens, the block is always inserted at the very end of the content.

Expected: we should decide on a "default" behavior and fall back to that state so that the `RangeControl` is always shown and used to calculate insertion. I'm thinking "percentage" should be the fallback since that was the only option available prior to #755 and most of the affected prompts are probably intended to be inserted based on percentage.

dkoo:
### All Submissions:

Fixes an edge case where an inline prompt can get stuck with an invalid "insertion position" or `trigger_type` value, which causes the `RangeControl` used to set the insertion position to not appear at all in the editor, and the prompt to always appear at the very end of the content.

Prior to #755 `scroll` was really the only valid `trigger_type` value for inline prompts, so all inline prompts behaved the same way regardless of what that value was. After #842, a second `blocks_count` option was added, but it's still possible for a prompt to be saved as an overlay, then converted to an inline prompt, and have `trigger_type` be set to `time` as a result. This is the edge case fixed by this PR. The fix is to make any value other than `blocks_count` behave like `scroll`, effectively making this the default/fallback behavior.

Closes #842.

1. Publish an overlay prompt with the "Trigger" option set to "Timer".
2. After the post is saved, change the "Prompt Type" to "Inline". This puts the the prompt into a state where it's inserted as an inline prompt, but its `trigger_type` value is still set to `time`, an invalid value for inline prompts.
3. On `master`, in the editor under the "Prompt Settings" sidebar, observe that the `RangeControl` to set the precise insertion position is not rendered at all.
4. On the front-end, observe that the prompt gets inserted like an inline prompt at the end of every article.
5. Check out this branch. Confirm that in the editor, the "Insertion Position" is set to "Scroll" and the `RangeControl` appears with the default value of 30%.
6. Confirm on the front-end that the prompt gets inserted at 30% of the content, which is the default inline prompt behavior.
7. Conversely, set the "Insertion Position" option to "Blocks Count" and save, then convert the "Prompt Type" to an Overlay. This puts the prompt into a state where it's an overlay with a `trigger_type` of `blocks_count`, an invalid value for overlays.
8. Confirm that in the editor, the "Trigger" is shown as the default value "Timer".
9. Confirm on the front-end that the overlay appears after a 3 second delay, the default delay.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -380,7 +380,7 @@ final class Newspack_Popups_Inserter {
 				$position          = $inline_popup['precise_position'];
 				$trigger_type      = $inline_popup['options']['trigger_type'];
 				$insert_at_zero    = 0 === $position; // If the position is 0, the prompt should always appear first.
-				$insert_for_scroll = 'scroll' === $trigger_type && $pos > $position;
+				$insert_for_scroll = 'blocks_count' !== $trigger_type && $pos > $position;
 				$insert_for_blocks = 'blocks_count' === $trigger_type && $block_index >= $position;
 
 				if ( $insert_at_zero || $insert_for_scroll || $insert_for_blocks ) {
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -606,6 +606,7 @@ final class Newspack_Popups_Model {
 				case 'scroll':
 					$popup['options']['trigger_delay'] = 0;
 					break;
+				case 'blocks_count':
 				case 'time':
 				default:
 					$popup['options']['trigger_scroll_progress'] = 0;
--- a/src/editor/Sidebar.js
+++ b/src/editor/Sidebar.js
@@ -176,21 +176,20 @@ const Sidebar = props => {
 						] }
 						onChange={ value => onMetaFieldChange( 'trigger_type', value ) }
 					/>
-					{ 'scroll' === trigger_type && (
+					{ 'blocks_count' === trigger_type ? (
+						<RangeControl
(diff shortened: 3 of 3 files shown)


miguelpeixe (changes requested):
It's working perfectly for switching from overlay to inline.

After saving the inline prompt to use "block count" and switching back to overlay, the "Trigger" select displays "Time" but I'm unable to select the time and the overlay prompt does not render on the frontend:

image

The delay `RangeControl` does appear if I change the trigger options manually after selecting overlay.

dkoo:
> After saving the inline prompt to use "block count" and switching back to overlay, the "Trigger" select displays "Time" but I'm unable to select the time

Ah, good catch—this required the same fix for making the default trigger type "Timer" for overlays: ccd8827

> and the overlay prompt does not render on the frontend

This wasn't the case for me, at least when I viewed it in a new incognito window. Overlays can only be shown with "Once" or "Once per day" frequency, so if you viewed that prompt as an inline prompt, converted it to overlay, then tried to view it again, it would be dismissed as already seen. Can you confirm that you see the overlay in a new session after converting from inline to overlay?

miguelpeixe:
Thank you for revisign! Switching prompt types now behave as expected! 💯 

> This wasn't the case for me, at least when I viewed it in a new incognito window. 

Yep! Must've been the frequency config.

miguelpeixe (approved):

Issue #864, opened 2022-05-05
Dismiss button in wrong location in WordPress 6.0 RC


laurelfulford:
When using the WP 6.0 RC, if you create a new prompt, the Dismiss button previews above the content as soon as you start editing it:

![image](https://user-images.githubusercontent.com/177561/166989475-efd916af-265e-4ca5-9feb-a349dd3e77dc.png)

It displays in the correct location on the front-end, and this issue doesn't appear to happen when editing existing prompts. If you publish your new prompt and exit and reenter the editor, the issue corrects itself.

## Steps to recreate

1. Start on a test site running the latest WP 6.0 RC
2. Create a new Overlay prompt
3. Edit the empty paragraph block that's there. 
4. Note that the Dismiss text and paragraph block swap locations.

dkoo:
Closing as https://github.com/Automattic/newspack-popups/pull/866 has been merged.

PR #866, opened 2022-05-13
feat: remove permanent suppression features

dkoo:
### All Submissions:

This PR proposes to remove all features related to permanent suppression. The concept of permanent suppression was built into the plugin prior to the introduction of segmentation features, and in many cases permanent suppression logic acts at odds with expected prompt behavior with segmentation. 

This also removes the following plugin suppression settings, which are better handled via segmentation:

* Suppress all newsletter-related prompts if a reader has suppressed a single newsletter-related prompt
* Suppress all newsletter-related prompts if a reader is known to have subscribed
* Suppress all newsletter-related prompts if a reader arrives at the site with a referrer URL containing `utm_medium=email` - we shouldn't assume that this means the reader is a subscriber.
* Suppress all donation-related prompts if a reader is known to have donated

After this change, readers can no longer permanently suppress any prompt. Prompts will be displayed as long as the reader matches both the segmentation and frequency criteria for the prompt. For example: given an overlay with frequency setting "Daily" and assigned to a segment for non-subscribers only: 

* A new reader will see this prompt on the first page view each day.
* The prompt may still be dismissed via the standard ✖ "close" button or by clicking on the overlay background, but the prompt will be displayed again after 24 hours.
* If the reader subscribes to a newsletter list via an onsite signup prompt, they will no longer see the prompt.

The above changes should make the prompt display logic easier to understand. It should also result in increased prompt views and conversions, since readers will be more likely to encounter prompts at appropriate times.

1. Check out this branch and build.
2. Visit **Newspack > Campaigns > Settings**. Confirm that the "Suppression Settings" section no longer exists.
3. Edit or create a prompt. Confirm that the "Dismiss Button Settings" sidebar panel no longer exists, and that no dismiss button preview is rendered in the block preview window.
4. Test the prompt display:
  - For an inline prompt with no segment ("Everyone"), the prompt should appear on every article (or archive, if an archive prompt) with no option to dismiss it.
  - For an overlay prompt with no segment, it should appear on the first page view of the session or day. It should still be dismissible via the ✖ "close" button or by clicking on the overlay background.
  - Segmentation and frequency options should continue behaving as expected, without the extra layers of logic around permanent suppression and dismissal.
5. Confirm that the unit tests still pass, and that the tests that have been altered or removed make sense according to the new prompt display logic.

### Other information:


yogeshbeniwal:
Thank you for making it easier, it was becoming tricky to track.

adekbadek (approved):
This looks great and removes a major source of confusion arising from the now-removed settings 🙌 

> BREAKING CHANGE: This PR should trigger a new major version bump since it introduces breaking changes.

The `BREAKING CHANGE: …`, along with a description of what's non-backwards-compatible, [should be a part of the commit message](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#footer).

adekbadek:
This PR closes https://github.com/Automattic/newspack-popups/issues/864, since the affected UI is removed here.

PR #869, opened 2022-05-18
refactor: segmentation data

dkoo:
### All Submissions:

This PR implements a significant refactor of segmentation data storage and retrieval for improved performance. It does the following:

* Creates two new tables with a column and index structure optimized for performance:
  - The "readers" table replaces the legacy "transients" table and normalizes data that was previously serialized in a single column into separate, indexable columns
  - The "reader events" table replaces the legacy "events" table and stores info on not just `post_read` events, but also non-article page views, prompt views, donations, and newsletter subscriptions. In the future, when we're able to support a JS-based alternative to `amp-access`, this table may be offloaded from the server to `localStorage`.
  - Because the above are stored as events, we no longer replicate the data in the readers table, except for keeping a count of read categories, article views, and non-article page views. A reader is now considered a donor if they have at least one "donation" event associated with their unique client ID, for example.
  - The legacy readers table is retained, but existing readers who visit the site will be migrated to the new model and their legacy data deleted. Migrated data includes a lifetime count of posts and categories read, donations, and subscriptions. Prompt views and dismissals will not be migrated as we no longer need to store permanent suppression data. Old data lacking donations and subscriptions will be purged after 30 days.
  - The legacy events table can be dropped if it exists. (**Note:** I didn't implement this in case we need to roll back the refactor. We can push a new PR to drop these tables once we confirm the refactor is successful.) Since `post_read` events are no longer needed after 1 hour, this table can be safely deprecated. It will result in a 1-hour period where all returning readers are treated as not having seen any prompts, after which the new data model will start tracking prompt views as usual. I doubt this will be noticeable to most readers.
  - The previous legacy handling (v1 transients data) is deprecated, since no existing Newspack sites should have any data in the original model at this point.
* Improved data purging logic: we purge article and page view events every hour (as opposed to every 30 days) as we only need to keep a record of unique post IDs and timestamps during a single session. Prompt view events are kept for 30 days (to facilitate upcoming new frequency options for daily/weekly/monthly periods), and donation/subscription events are kept permanently.

The above changes should give us some significant performance boosts:

* We avoid the use of `LIKE` queries and instead base all queries on indexed columns
* We avoid unlimited `DELETE` queries and instead chunk rows to delete into groups of 1000, to avoid replication lag
* We avoid potentially unlimited row lengths caused by the current serialized data model
* We should hopefully limit the growth of the events table by purging article/page view events more often, although this may be offset by moving prompt views/donations/subscriptions to this table
* We still rely on the persistent WP cache (if available) for retrieving reader and events data per page view. Events are logged to a flat file and parsed to the DB every 15 minutes. This behavior is unchanged from the current.

The refactor should result in few if any reader-facing changes, except for faster site performance at scale. The exceptions:

* A "session" is now considered one whole hour instead of 45 minutes. This is simply because WP cron supports the `hourly` period, but it shouldn't make a huge difference in segmentation behavior. 45 minutes was an arbitrary choice to begin with.
* `post_read` events are renamed as `article_view` for clarity. The behavior is the same—a view of a singular post of any public-facing post type is considered an article view.
* Non-article page views (archives, search results) are now tracked as `page_view`. This will become important for upcoming prompt frequency options based on total page views.
* Article/page views are now only de-deduplicated for the current session, instead of for the lifetime of the reader. For example, if you visit or refresh the same post multiple times within an hour, that will still count as only a single article view. If you visit that same article again the next day, it will count as a new article view. This is closer to how Google Analytics counts page views per session, and will be important for upcoming prompt frequency options based on page views.
* We no longer store subscription/donation data on the reader data itself, instead storing these as separate events that can be cross-queried using the client ID. In the future, if we move the events data to `localStorage`, we may want to retain subscription and donation data on the server so we don't lose that data if the browser cache is purged.

It's hard to quantify the performance improvements that this refactor should enable without running the plugin at scale, so we should plan to test this refactor with a small number of sites before rolling out completely. Other than performance issues:

* This refactor should result in **no reader-facing changes**. So please thoroughly test all important prompt and segmentation features as a reader and ensure that nothing unexpected happens.
* Also test the editor experience, particularly the Campaigns Analytics dashboard, and make sure you see nothing unexpected there, too.
* Due diligence: test the data migration handling by running the plugin on `master` first. 
  1. In a new incognito window, visit several posts with categories, subscribe via a newsletter signup prompt, and donate via a donate prompt. Note which posts/categories you viewed so you can confirm those views are migrated to the new model.
  2. inspect your DB tables directly via Adminer or Sequel Pro and note the client ID for your reader activity as created in the `wp_newspack_campaigns_transients` table.
  3. Switch to this branch and **in the same incognito window** (so that your client ID remains the same) continue navigating the site and confirm that you still see the prompts you'd expect to see based on your segment.
  4. Inspect your DB tables and confirm that a row is created for your client ID in the `wp_newspack_campaigns_readers` table that carries over the previously viewed posts and categories. Also confirm that events were created for your client ID in the `wp_newspack_campaigns_events` table for your subscription and donation.
* Also make sure to test in both AMP and non-AMP mode.

### Other information:

### DIFF
--- a/api/campaigns/class-campaign-data-utils.php
+++ b/api/campaigns/class-campaign-data-utils.php
@@ -114,7 +114,9 @@ class Campaign_Data_Utils {
 			array_filter(
 				$reader_events,
 				function ( $event ) {
-					return 'article_view' === $event['type'];
+					$hour_ago   = strtotime( '-1 hour', time() );
+					$event_time = strtotime( $event['date_created'] );
+					return 'article_view' === $event['type'] && $event_time > $hour_ago;
 				}
 			)
 		);
--- a/api/campaigns/class-campaign-data-utils.php
+++ b/api/campaigns/class-campaign-data-utils.php
@@ -10,30 +10,44 @@
  */
 class Campaign_Data_Utils {
 	/**
-	 * Is client a subscriber?
+	 * Is reader a newsletter subscriber?
 	 *
-	 * @param object $client_data Client data.
+	 * @param object $reader_events Reader events.
 	 */
-	public static function is_subscriber( $client_data ) {
-		return ! empty( $client_data['email_subscriptions'] );
+	public static function is_subscriber( $reader_events ) {
+		return 0 < count(
+			array_filter(
+				$reader_events,
+				function( $event ) {
+					return 'subscription' === $event['type'];
+				}
+			)
+		);
 	}
 
 	/**
-	 * Is client a donor?
+	 * Is reader a donor?
 	 *
-	 * @param object $client_data Client data.
+	 * @param object $reader_events Reader events.
 	 */
-	public static function is_donor( $client_data ) {
-		return ! empty( $client_data['donations'] );
+	public static function is_donor( $reader_events ) {
+		return 0 < count(
+			array_filter(
+				$reader_events,
+				function( $event ) {
+					return 'donation' === $event['type'];
+				}
+			)
+		);
 	}
 
 	/**
-	 * Is client logged in?
+	 * Does reader have a WP user account?
 	 *
-	 * @param object $client_data Client data.
+	 * @param object $reader_data Reader data.
 	 */
-	public static function is_logged_in( $client_data ) {
-		return ! empty( $client_data['user_id'] );
+	public static function has_user_account( $reader_data ) {
+		return ! empty( $reader_data['user_id'] );
 	}
 
 	/**
@@ -83,68 +97,52 @@ class Campaign_Data_Utils {
 	 * Given a segment and client data, decide if the prompt should be shown.
 	 *
 	 * @param object $campaign_segment Segment data.
-	 * @param object $client_data Client data.
+	 * @param string $reader_data Reader data for the given client ID.
+	 * @param string $reader_events Reader events for the given client ID.
 	 * @param string $referer_url URL of the page performing the API request.
 	 * @param string $page_referrer_url URL of the referrer of the frontend page that is making the API request.
 	 * @return bool Whether the prompt should be shown.
 	 */
-	public static function does_client_match_segment( $campaign_segment, $client_data, $referer_url = '', $page_referrer_url = '' ) {
-		$should_display = true;
-		// Posts read.
-		$posts_read_count = count( $client_data['posts_read'] );
-		// Posts read in the current session.
-		$session_start            = strtotime( '-45 minutes', time() );
-		$posts_read_count_session = count(
+	public static function does_reader_match_segment( $campaign_segment, $reader_data, $reader_events, $referer_url = '', $page_referrer_url = '' ) {
+		$should_display              = true;
+		$is_subscriber               = self::is_subscriber( $reader_events );
+		$is_donor                    = self::is_donor( $reader_events );
+		$has_user_account            = self::has_user_account( $reader_data );
+		$campaign_segment            = self::canonize_segment( $campaign_segment );
+		$article_views_count         = $reader_data['article_views'];
+		$article_views_count_session = count(
 			array_filter(
-				$client_data['posts_read'],
-				function ( $post_data ) use ( $session_start ) {
-					if ( ! isset( $post_data['created_at'] ) ) {
-						return false;
-					}
-					return strtotime( $post_data['created_at'] ) > $session_start;
+				$reader_events,
+				function ( $event ) {
+					return 'article_view' === $event['type'];
 				}
 			)
 		);
-		$is_subscriber            = self::is_subscriber( $client_data );
-		$is_donor                 = self::is_donor( $client_data );
-		$is_logged_in             = self::is_logged_in( $client_data );
-		$campaign_segment         = self::canonize_segment( $campaign_segment );
 
 		// Read counts for categories.
-		$categories_read_counts = array_reduce(
-			$client_data['posts_read'],
-			function ( $read_counts, $read_post ) {
-				foreach ( explode( ',', $read_post['category_ids'] ) as $cat_id ) {
-					if ( isset( $read_counts[ $cat_id ] ) ) {
-						$read_counts[ $cat_id ]++;
-					} else {
-						$read_counts[ $cat_id ] = 1;
-					}
-				}
-				return $read_counts;
-			},
-			[]
-		);
-		arsort( $categories_read_counts );
-		$favorite_category_matches_segment = in_array( key( $categories_read_counts ), $campaign_segment->favorite_categories );
+		if ( $reader_data['categories_read'] ) {
+			$categories_read_counts = $reader_data['categories_read'];
+			arsort( $categories_read_counts );
+			$favorite_category_matches_segment = in_array( key( $categories_read_counts ), $campaign_segment->favorite_categories );
+		}
 
 		/**
-		 * By posts read count.
+		 * By article view count.
 		 */
-		if ( $campaign_segment->min_posts > 0 && $campaign_segment->min_posts > $posts_read_count ) {
+		if ( $campaign_segment->min_posts > 0 && $campaign_segment->min_posts > $posts_rearticle_views_countad_count ) {
 			$should_display = false;
 		}
-		if ( $campaign_segment->max_posts > 0 && $campaign_segment->max_posts < $posts_read_count ) {
+		if ( $campaign_segment->max_posts > 0 && $campaign_segment->max_posts < $article_views_count ) {
 			$should_display = false;
 		}
 
 		/**
-		 * By posts read in session count.
+		 * By article views in past hour.
 		 */
-		if ( $campaign_segment->min_session_posts > 0 && $campaign_segment->min_session_posts > $posts_read_count_session ) {
+		if ( $campaign_segment->min_session_posts > 0 && $campaign_segment->min_session_posts > $article_views_count_session ) {
 			$should_display = false;
 		}
-		if ( $campaign_segment->max_session_posts > 0 && $campaign_segment->max_session_posts < $posts_read_count_session ) {
+		if ( $campaign_segment->max_session_posts > 0 && $campaign_segment->max_session_posts < $article_views_count_session ) {
 			$should_display = false;
 		}
 
@@ -171,10 +169,10 @@ class Campaign_Data_Utils {
 		/**
 		 * By login status.
 		 */
-		if ( $campaign_segment->is_logged_in && ! $is_logged_in ) {
+		if ( $campaign_segment->is_logged_in && ! $has_user_account ) {
 			$should_display = false;
 		}
-		if ( $campaign_segment->is_not_logged_in && $is_logged_in ) {
+		if ( $campaign_segment->is_not_logged_in && $has_user_account ) {
 			$should_display = false;
 		}
 
--- a/api/campaigns/class-campaign-data-utils.php
+++ b/api/campaigns/class-campaign-data-utils.php
@@ -120,6 +120,7 @@ class Campaign_Data_Utils {
 		);
 
 		// Read counts for categories.
+		$favorite_category_matches_segment = false;
 		if ( $reader_data['categories_read'] ) {
 			$categories_read_counts = $reader_data['categories_read'];
 			arsort( $categories_read_counts );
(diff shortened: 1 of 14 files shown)


miguelpeixe (changes requested):
Thank you for putting together this massive amount of work so quickly! It's looking very promising! 🎉 

I ran into a few issues and have some questions and suggestions about some of the architectural decisions.

I'm not sure "article" is the best name for any public-facing post type. It's hard to determine a generic enough term for this, considering it can be an article, a page, an event, a newsletter, or a listing, I believe `post` is already the widely accepted generic term in the WordPress community.

A new session landing on the homepage counts as `article_views` on the reader data table and an `article_read` event on the events table. With the current terminology, I'd expect a `page_view` or nothing at all, since I haven't read any articles.

I had an issue with the "Prompt dismissed" event. It was triggered twice (2 identical events with the same timestamp) on an overlay prompt at the homepage.

Another issue with the `donation` event. Donating on a prompt with the WC Stripe donation block generates a malformed row:

image

Weirdly, this row was generated after searching for a random term:

image

Every new search generates a new row like that for that client and every view (post/page) for that client generates a new malformed row.

The client has indeed donated but the donation row was never created by donating.

### Architecture

I worry that we currently do not have a strategy that scales persistent data. The persistent/aggregated data is only stored as columns in a reader row. As we move forward, we're likely to increase the collection of data we'd like to retain for a reader and it shouldn't require a migration effort to support such enhancements.

To maintain the current amount of tables, I suggest we rename `reader_events` to `reader_data` and add a boolean indexed column to determine if its an event, in order to appropriately implement APIs for each type of data:

- `update_reader_data( $reader_id, $data_name, $value )`: Upserts a non-event row for a reader - `event = false`
- `add_reader_event( $reader_id, $event_name, $value)`: Add a new event row for a reader. - `event = true`

With that approach, `article_views`, `page_views`, and `categories_reader` would no longer be columns on the readers table.

#### Composite indexes

As discussed in the first post about the refactor, the existing queries can be optimized by using composite indexes:
- `client_id`, `type`
- `type`, `created_at`

With those indexes, there's no need for single indexes for `client_id`, `type`, or `created_at`.

dkoo:
Pausing on this while we discuss the data model changes suggested by @miguelpeixe [above](https://github.com/Automattic/newspack-popups/pull/869#pullrequestreview-990420370).

I also found that there are some issues with reporting `prompt_seen` events while in non-AMP mode, so I'll investigate that in the meantime. I added a testing step to make sure we repeat testing steps in both AMP and non-AMP modes to the original PR testing instructions.

dkoo:
@miguelpeixe in 14c0c66...36efb08 I started with your suggestions as outlined above:

> I worry that we currently do not have a strategy that scales persistent data. The persistent/aggregated data is only stored as columns in a reader row. As we move forward, we're likely to increase the collection of data we'd like to retain for a reader and it shouldn't require a migration effort to support such enhancements.
> 
> To maintain the current amount of tables, I suggest we rename reader_events to reader_data and add a boolean indexed column to determine if its an event, in order to appropriately implement APIs for each type of data:
> 
>     update_reader_data( $reader_id, $data_name, $value ): Upserts a non-event row for a reader - event = false
>     add_reader_event( $reader_id, $event_name, $value): Add a new event row for a reader. - event = true
> 
> With that approach, article_views, page_views, and categories_reader would no longer be columns on the readers table.

But I found it added some complexity to the API to upsert persistent/aggregated data, because of the way we're caching events to the flat file and only writing to the DB once per 15 minutes. In 29de88e I tried moving that data back to the `readers` table, but as JSON-encoded data in a new `reader_data` column instead of as separate columns. I think this will allow us to easily add more types of aggregated data in the future if needed, while also maintaining some simplicity in the `reader_events` table and how we write to it. As a failsafe, 29de88e also adds a feature where we can increment a `TABLE_VERSION` constant in the `Newspack_Popups_Segmentation` class, which will trigger a `dbDelta` query that will let us add new columns to either table if needed. I hope this will be used rarely if ever, but it does give us the option to add table columns without a full migration in the future. WDYT?

api/classes/class-lightweight-api.php
```diff
@@ -227,161 +226,580 @@ public function delete_transient( $name ) {
...
+		if ( ! $this->ignore_cache ) {
+			$events = wp_cache_get( 'reader_events', $client_id );
+			if ( ! empty( $events ) ) {
+				$this->debug['reader_events'] = $events;
+				return $this->filter_events_by_type( $events, $type, $context );
+			}
+		}
+
+		// If no data in the cache, get it from the DB.
+		$events = $this->get_reader_events_from_db( $client_id );
```
miguelpeixe, line 630: Storing the reader events with proper structure and indexes allows us to have a more powerful API to perform optimized queries. Shouldn't we start using those queries instead of always fetching all events and processing/filtering on the PHP?

api/classes/class-lightweight-api.php
```diff
@@ -227,161 +226,580 @@ public function delete_transient( $name ) {
...
 	 */
-	public function get_all_clients_data() {
+	public function get_all_readers_data() {
 		global $wpdb;
-		$events_table_name = Segmentation::get_events_table_name();
+		$readers_table_name = Segmentation::get_readers_table_name();
 
 		// Results are limited to the 1000 most recent rows for performance reasons.
-		$all_client_ids_rows = $wpdb->get_results( "SELECT DISTINCT client_id,id FROM $events_table_name ORDER BY id DESC LIMIT 1000" ); // phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching, WordPress.DB.PreparedSQL.InterpolatedNotPrepared
+		$all_client_ids_rows = $wpdb->get_results( "SELECT DISTINCT client_id,date_modified FROM $readers_table_name WHERE is_preview IS NULL ORDER BY date_modified DESC LIMIT 1000" ); // phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching, WordPress.DB.PreparedSQL.InterpolatedNotPrepared
```
miguelpeixe, line 788: Are we ok not retrieving all data when the total exceeds 1000?

includes/class-newspack-popups-segmentation.php
```diff
@@ -307,58 +328,58 @@ function( $item ) {
...
+		$reader_events_sql = "CREATE TABLE $reader_events_table_name (
+			id bigint(20) NOT NULL AUTO_INCREMENT,
+			client_id varchar(100) NOT NULL,
+			date_created datetime NOT NULL DEFAULT CURRENT_TIMESTAMP,
+			type varchar(20) NOT NULL,
+			context varchar(20) NOT NULL,
+			value longtext,
+			PRIMARY KEY  id (id),
+			KEY client_id_type_context (client_id, type, context),
+			KEY date_created (date_created)
```
miguelpeixe, line 377: `client_id` index is not required with the composite `client_id, type, context`

miguelpeixe (changes requested):
The structure appears to be working well and I was able to test Mailchimp subscription and regular navigation with different clients.

I was not able to test `donation`. It didn't trigger an event when using the streamline Donation block with Stripe on a prompt. Is there another way of testing this?

includes/class-newspack-popups-segmentation.php
miguelpeixe: I just now realized that the index is created automatically due to the `NOT NULL` rule.

miguelpeixe (changes requested):

api/classes/class-lightweight-api.php
dkoo: This function is only used to calculate approximate reach/viewability of a segment. The query is limited to 1000 to avoid performance issues on very large sites that will have many thousands of readers.

api/classes/class-lightweight-api.php
```diff
@@ -227,161 +226,580 @@ public function delete_transient( $name ) {
...
+							],
+						];
+					}
+				}
 
-		if ( $data && isset( $data['prompts'] ) && isset( $data['prompts'][ $campaign_id ] ) ) {
-			// NEW FORMAT - one row per reader.
-			$prompt = $data['prompts'][ $campaign_id ];
+				if ( ! empty( $reader_events ) ) {
+					$this->save_reader_events( $client_id, $reader_events );
```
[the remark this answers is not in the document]
dkoo: This is an excellent point—will refactor this migration feature to take that into account. More to come.

api/classes/class-lightweight-api.php
```diff
@@ -227,161 +226,580 @@ public function delete_transient( $name ) {
...
-	 * @param string $campaign_id Prompt ID.
-	 * @return object Prompt data.
+	 * @param string $client_id Client ID to check.
+	 *
+	 * @return boolean True if a preview.
 	 */
-	public function get_campaign_data_legacy( $client_id, $campaign_id ) {
-		return $this->get_transient( $this->get_transient_name_legacy( $client_id, $campaign_id ) );
+	public function is_preview( $client_id ) {
+		return 'preview' === substr( $client_id, 0, 7 );
```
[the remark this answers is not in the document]
dkoo: We use the `is_preview` column in order to query these rows, but we need this algorithm to decide which rows get flagged as `is_preview` in the first place (before they're written to the DB).

api/campaigns/class-campaign-data-utils.php
```diff
@@ -52,6 +85,63 @@ public static function does_referrer_match( $page_referrer_url, $referrers_list_
...
 
+	/**
+	 * Given a reader's data, get the total view count of singular posts.
+	 * Articles are any singular post type in Newspack sites.
+	 *
+	 * @param array $reader Reader data.
+	 *
+	 * @return int Total view count of singular posts.
+	 */
+	public static function get_post_view_count( $reader ) {
```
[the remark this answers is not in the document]
dkoo: Whoops, good catch—that's a remnant of a prior refactor. Fixed in 20de3e7.

api/campaigns/class-maybe-show-campaign.php
```diff
@@ -37,40 +35,30 @@ public function __construct() {
...
-				];
-				$this->save_client_data(
-					$client_id,
-					[
-						'posts_read' => $posts_read,
-					]
-				);
+		// Log an article or page view event.
+		if ( $visit && ( ! defined( 'DISABLE_CAMPAIGN_EVENT_LOGGING' ) || true !== DISABLE_CAMPAIGN_EVENT_LOGGING ) ) {
+			$read_event = [
```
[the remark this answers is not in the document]
dkoo: Good call—updated all instances of `$read_event` to `$view_event` in 163f059.

api/classes/class-lightweight-api.php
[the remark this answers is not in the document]
miguelpeixe: Got it! Thanks!

api/classes/class-lightweight-api.php
miguelpeixe: In that case, it is more costly to filter the data on PHP than to use indexed and optimized queries to fetch the exact data that is going to be used.

The number of DB hits should be the same, with the difference that we'd be skipping a PHP processing/filtering step, making it generally lighter. Fewer items to retrieve and fewer data to process.

We can optimize that on another PR, but I'd still prioritize it for performance sake.

miguelpeixe (changes requested):
I'm still not able to test the `donation` event, from the previous review:

> I was not able to test `donation`. It didn’t trigger an event when using the streamline Donation block with Stripe on a prompt. Is there another way of testing this?

While testing other event types/contexts I noticed the event data storing double-encoded JSON:

image

dkoo:
> I was not able to test donation. It didn't trigger an event when using the streamline Donation block with Stripe on a prompt. Is there another way of testing this?

I tested with both WooCommerce and Stripe platforms, submitting one-time and recurring donations via a block in post content as well as inside a prompt. In all cases I saw a `donation` event logged with the order details, which caused my session to match a "donors" segment on subsequent pageviews.

That said, [1ce8003](https://github.com/Automattic/newspack-popups/pull/869/commits/1ce8003cc2ed6f6531e53d895c4fd82dac430b9d) does add logic which logs the platform of the donation (`stripe` or `woocommerce`) as the `context` value.

Could you try again with the latest changes?

dkoo:
> While testing other event types/contexts I noticed the event data storing double-encoded JSON:

6db4883 fixes the double-encoding, and also fixes a bug where multiple visits to the same non-singular page in the same session aren't properly deduplicated.

api/classes/class-lightweight-api.php
dkoo: I [created an issue](https://github.com/Automattic/newspack-popups/issues/907) to tackle this in a separate PR because I don't want to refactor this data handling again in this PR!

dkoo:
@miguelpeixe 4a82ba0 is (hopefully 🤞) the last large change on this PR. It does the following:

* Fixes the issue with repeated `donation` events you [highlighted above](https://github.com/Automattic/newspack-popups/pull/869#pullrequestreview-1030658832). This was happening for logged-in readers with WooCommerce orders attached to their user ID—the API was not checking whether the orders were already logged as donation events before logging them.
* Changed the `get_reader_events` API to only query by the needed types and contexts, instead of querying all events and filtering via PHP (as [suggested above](https://github.com/Automattic/newspack-popups/pull/869#discussion_r907669777)). This does result in a few more DB queries than before, but hopefully the more precise queries will offset that, as you mentioned.
* Changed the use of the `context` column for `subscription` and `user_account` type events. Now, the reader's email address and WP user ID are stored in the `context` column, instead of in the `value`. This is so that we can more easily query on those values, which will be important when we need to be able to look up whether a new reader subscribes with a previously known email address or logs in with a known user account.
  - Because of this, I bumped up the length of the `context` column to `varchar(100)` from `varchar(20)`, to accommodate long email address strings.
  - I also moved the ESP info to the event `value`. It's still valuable to store this data if we have it, but seems less important to me to be able to query on it compared with the actual email address.
  - I thought of doing the same with `donation` type events, but that's a little trickier because `donation` events can come from various platforms (Stripe or WC to start), so there's no guarantee that their IDs will be unique or if the donation data will even have an ID value to store as `context`. So for now, I'm leaving `donation` events with the platform in the `context` column and all transaction details in `value`.

miguelpeixe (approved):
🙌 🎉

PR #879, opened 2022-06-03
feat: disable prompts by default for protected pages

dkoo:
### All Submissions:

Disables prompts by default on special/protected pages, such as My Account, Privacy Policy, Donation, etc. other than the homepage and blog posts page.

Also makes the "Disable prompts on this post or page" option available to all post types that can possibly show prompts.

Requires https://github.com/Automattic/newspack-plugin/pull/1686.

1. Using WP CLI, ensure that the `newspack_popups_has_disabled_popups` meta value doesn't exist on a protected page by running `wp post meta delete newspack_popups_has_disabled_popups <post_id>`. See `\Newspack\Patches` class for a list of protected pages.
2. Edit the page and confirm that by default, the "Disable prompts on this post or page" is turned on. Also confirm that this behavior is true on the front-end and that no prompts are displayed.
3. Confirm that you can still toggle the option off to intentionally force prompts to appear on those protected pages.
4. Confirm that behavior for non-protected pages and posts remains the same—"Disable prompts on this post or page" should be `false` by default, and prompts should load as usual unless the option is turned on.
5. Bonus: confirm that the "Disable prompts on this post or page" option now exists for any public-facing post type, not jsut posts and pages.

### Other information:

### DIFF
--- a/includes/class-newspack-popups.php
+++ b/includes/class-newspack-popups.php
@@ -71,6 +71,7 @@ final class Newspack_Popups {
 			add_action( 'init', [ __CLASS__, 'register_cpt' ] );
 			add_action( 'init', [ __CLASS__, 'register_meta' ] );
 			add_action( 'init', [ __CLASS__, 'register_taxonomy' ] );
+			add_action( 'init', [ __CLASS__, 'disable_prompts_for_protected_pages' ] );
 			add_action( 'enqueue_block_editor_assets', [ __CLASS__, 'enqueue_block_editor_assets' ] );
 			add_action( 'customize_controls_enqueue_scripts', [ __CLASS__, 'enqueue_customizer_assets' ] );
 			add_filter( 'display_post_states', [ __CLASS__, 'display_post_states' ], 10, 2 );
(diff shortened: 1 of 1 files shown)


includes/class-newspack-popups.php
```diff
@@ -71,6 +71,7 @@ public function __construct() {
 			add_action( 'init', [ __CLASS__, 'register_cpt' ] );
 			add_action( 'init', [ __CLASS__, 'register_meta' ] );
 			add_action( 'init', [ __CLASS__, 'register_taxonomy' ] );
+			add_action( 'init', [ __CLASS__, 'disable_prompts_for_protected_pages' ] );
```
miguelpeixe, line 74: Non-blocking: could be hooked to a less broad hook to not fire so often without need. `template_redirect` should be an early enough hook

dkoo:
> If the intention to use the meta is so the editor can change it

This is the intention—I didn't want to take away the ability to make prompts appear on the page. Initially I tried to dynamically set the `default` value on the `register_meta` function based on the post ID, but meta registration happens before `get_the_ID()` is available. The whole feature could be simplified a lot if we agree that taking away that ability is okay, in which case we could simply use the hook as you suggest.

> But I was never able to make a prompt appear on that page, I can't figure out why.

I'm not able to replicate this—if I disable the option and save, then view the page front-end, I see prompts. I'm testing with the Donate page on my site. Could you try appending `?newspack-campaigns-debug` to the URL and inspecting the `debug` object on the `campaigns` API request response? It may have a `suppression` key with an explanation as to why the prompt was suppressed.

miguelpeixe:
Don't know what happened then, it's working now! Reenabling prompts for a protected page behave as expected.

dkoo:
> Don't know what happened then, it's working now! Reenabling prompts for a protected page behave as expected.

Strange! Glad it's working as expected now. What do you think about the idea of retaining the ability to change the setting for protected pages? Is there a use case there?

miguelpeixe (approved):
It's a good idea to allow a publisher to change this default setting. I can't think of a reason to, but campaigns can be very broad and there's no telling what strategies someone may have 😄

PR #892, opened 2022-06-22
fix: check subscriber status earlier and restore utm_medium=email logic

dkoo:
### All Submissions:

#866 removed permanent suppression features as well as the logic that treated visitors coming to the site via any URL with a `utm_medium=email` param as a subscriber. However, removing both has resulted in some unwanted behavior for readers who visit sites via a newsletter link, but in a new or cache-cleared browser session. For these readers, prompt display logic is executed before their Mailchimp subscriber or donor status is evaluated, meaning they will be segmented as a subscriber/donor only after the first page load.

This PR evaluates the Mailchimp subscriber and donor status before executing segmentation logic, so that by the time the plugin decides which segment's prompts to display, a reader who arrives via a Mailchimp URL should already be known to be a subscriber/donor.

This PR also restores the logic that looks for the `utm_medium=email` param in the referrer URL. However, instead of using this info to permanently suppress all newsletter-related prompts, it more simply causes the reader to be segmented as a subscriber. This does mean that readers who are not actually subscribers but who visit the site via a forwarded newsletter email, or who simply append the `utm_medium=email` param to a URL, will be considered subscribers. But this is already true for any visitor who visits the site via a forwarded Mailchimp email, as their Mailchimp subscriber status is parsed from Mailchimp URL params upon the first reported prompt view. So this plugs a functionality hole for readers who arrive at the site in a new session from a non-Mailchimp email newsletter, who would previously not be segmented as subscribers.

Note that if this gets merged, #869 will need to reimplement these changes in the new data model.

1. Check out this branch on a site that's connected to Mailchimp.
2. Set up two segments, one for non-subscribers and one for subscribers, and publish a prompt in each. Make sure the prompt for non-subscribers includes a Mailchimp signup form.
3. In a new browser session, visit the site and append `?utm_medium=email` to the URL.
4. Confirm that you see the prompt for subscribers on the first page load, and on every subsequent page load.
5. Send a newsletter from the site to yourself via Mailchimp.
6. Copy a URL from the sent newsletter to a new browser session. Confirm that you see the prompt for subscribers on the first page load, and on every subsequent page load.
7. Send a newsletter from the site to yourself via another ESP, such as Campaign Monitor.
8. Copy a URL from the sent newsletter to a new browser session. Confirm that you see the prompt for subscribers on the first page load, and on every subsequent page load. This is because all ESPs append `?utm_medium=email` to URLs within the email.
9. In a new browser session, sign up via the Mailchimp signup form in the prompt for non-subscribers.
10. Navigate through the site and confirm that you see the prompt for subscribers on subsequent page loads.

### Other information:

### DIFF
--- a/api/classes/class-lightweight-api.php
+++ b/api/classes/class-lightweight-api.php
@@ -6,6 +6,10 @@
  * @phpcs:disable WordPress.DB.RestrictedClasses.mysql__PDO
  */
 
+require_once __DIR__ . '/../../vendor/autoload.php';
+
+use \DrewM\MailChimp\MailChimp;
+
 /**
  * API endpoints
  */
@@ -394,6 +398,26 @@ class Lightweight_API {
 	}
 
 	/**
+	 * Get URL query param.
+	 *
+	 * @param string $param Param name.
+	 * @param string $url URL to parse.
+	 *
+	 * @return string|boolean Value of the param, or false if it's not in the URL.
+	 */
+	public function get_url_param( $param, $url ) {
+		$parsed_url = parse_url( $url ); // phpcs:ignore WordPress.WP.AlternativeFunctions.parse_url_parse_url
+		if ( ! empty( $parsed_url['query'] ) ) {
+			parse_str( $parsed_url['query'], $query );
+			if ( ! empty( $query[ $param ] ) ) {
+				return $query[ $param ];
+			}
+		}
+
+		return false;
+	}
+
+	/**
 	 * Get POST request payload.
 	 */
 	public function get_post_payload() {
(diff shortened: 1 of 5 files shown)


api/segmentation/class-segmentation-client-data.php
```diff
@@ -72,66 +68,7 @@ public function report_client_data( $request ) {
...
-
-						if ( $has_donated_according_to_mailchimp ) {
-							$client_data_update['donations'][] = [
-								'mailchimp_has_donated' => true,
-							];
-						}
-					}
-				}
-			}
+			$this->report_mailchimp( $client_id, $mailchimp_campaign_id, $mailchimp_subscriber_id );
```
adekbadek, line 71: Isn't this redundant now? If the MC API is queried when our API decides whether to display the prompts, why should it be queried again when the API is getting client data report?

api/classes/class-lightweight-api.php
```diff
@@ -413,6 +440,81 @@ public function get_post_payload() {
 		return $payload;
 	}
 
+	/**
+	 * Report Mailchimp data.
```
adekbadek, line 444: I'd say it's rather querying MC data, this (and the method name) suggests that data is reported _to_ MC.

adekbadek (changes requested):

api/classes/class-lightweight-api.php
dkoo: My thinking is that we're fetching data from MC and reporting it to Campaigns data, similar to how we "report" a prompt view or dismissal as an event. But I can see it both ways. feb6b0a renames the method to `get_mailchimp_client_data` instead.

api/segmentation/class-segmentation-client-data.php
dkoo: I think you're right. My initial thought was that the subscriber/donor status may change if the user takes action on that first page, but we capture that activity directly in those cases (if the reader subscribes or donates via a form, for example). 893864e removes the second call on client data report.

adekbadek (approved):

PR #897, opened 2022-06-24
feat: use existing options to restrict display size of overlay prompts

dkoo:
### All Submissions:

Uses the existing "hide border" and "overlay size" options to improve the display of overlay prompts in small viewports. Currently there's no way to prevent an overlay prompt from taking up the entire viewport at smaller screen sizes if its content is long enough, so these changes provide some means of ensuring that the prompt size can be limited independently of content length.

The "hide border" option was originally only available for inline prompts, but it makes sense to allow it for overlays as well. This PR enables that option for overlays and applies appropriate CSS to hide the white border around the overlay content, replacing it with a transparent margin of the same size.

The "overlay size" option now restricts the max-height of overlays as a percentage of viewport height. If the height of the prompt content exceeds this max height, the content will become scrollable.

* `x-small` maxes out at 50% of the viewport height.
* `small` maxes out at 60%.
* `medium` maxes out at 80%.
* `large` and `full-width` max out at 100%, so they can potentially fill the entire viewport if desired.

1. Check out this PR.
2. Create an overlay prompt with a lot of content. In the **Prompt Settings** sidebar, confirm that the "Hide Prompt Border" option is now available for overlays, off by default. Turn it on.
3. Preview the prompt. Confirm that the prompt is displayed without a white border wrapper.
4. Change the viewport size to mobile and confirm that the prompt content does not extend all the way to the outer edges of the viewport, but is set in 32px (the same size as the white border if the border were displayed).
5. Edit the prompt again and test each "Overlay size" option. Preview each size and confirm that the prompt height maxes out at an appropriate max-height for each size, with overflow content becoming scrollable. Test at all viewport sizes but pay special attention to the mobile experience.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -1167,7 +1167,7 @@ final class Newspack_Popups_Model {
 					action-xhr="<?php echo esc_url( $endpoint ); ?>"
 					target="_top">
 					<?php echo $hidden_fields; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>
-					<button style="opacity: <?php echo floatval( $overlay_opacity ); ?>;background-color:<?php echo esc_attr( $overlay_color ); ?>;" class="newspack-lightbox-shim" on="tap:<?php echo esc_attr( $element_id ); ?>.hide"></button>
+					<button style="opacity: <?php echo floatval( $overlay_opacity ); ?>; ?>;" class="newspack-lightbox-shim" on="tap:<?php echo esc_attr( $element_id ); ?>.hide"></button>
 				</form>
 			<?php endif; ?>
 		</amp-layout>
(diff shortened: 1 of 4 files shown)


includes/class-newspack-popups-model.php
```diff
@@ -1279,8 +1279,9 @@ public static function foreground_color_for_background( $background_color ) {
 	 * @return string Inline styles attribute.
 	 */
 	public static function container_style( $popup ) {
-		$background_color = $popup['options']['background_color'];
-		$foreground_color = self::foreground_color_for_background( $background_color );
+		$hide_border      = $popup['options']['hide_border'];
+		$background_color = $hide_border ? 'transparent' : $popup['options']['background_color'];
```
adekbadek, line 1283: With a transparent background, an overlay prompt won't look great:

image

adekbadek (changes requested):

includes/class-newspack-popups-model.php
dkoo: Ah, my test prompt happened to have a Group block with a background. 🤦 [8621638](https://github.com/Automattic/newspack-popups/pull/897/commits/86216384223d86e768f1ae7277927d71d2d0cc40) should make prompts look better without that.

adekbadek (approved):

dkoo:
Thanks, @adekbadek! Pinging @laurelfulford and @thomasguillot for the design perspective. This PR changes the way prompts are displayed on the front-end, by constraining their dimensions according to the existing "overlay size" option. This will mainly impact mobile viewports, but it may affect a lot of prompts because it's so easy to create a prompt that fills the entire mobile viewport. What do you both think about the new behavior and how it could change things for existing prompts?

thomasguillot:
I'd probably simplify this to `x-small`, `small`, `medium`, `large` with a max-height of `75vh` and only have `full-width` at `100vh`.

thomasguillot:
@dkoo never mind, it's fine as it is 😃

PR #908, opened 2022-07-06
feat: improved frequency options

### LINKED ISSUE #902: Improved frequency options
## As is

The frequency controls allow for displaying a prompt once, once daily, and until dismissed.

## To be 

The frequency options should allow for 

1. setting initial number of page views before the prompt is displayed. Arguably this can be achieved with segmentation – so this might not be necessary as a frequency option
1. setting in-between page view count, e.g. displaying a prompt every three views.

dkoo:
### All Submissions:

~**Note:** will require a rebase after #869 is merged.~ ✅ 

Adds new frequency options for displaying prompts based on pageviews and configurable time period if the `NEWSPACK_EXPERIMENTAL_READER_ACTIVATION` is enabled. Includes the following changes:

* **Frequency:** this existing dropdown lets you select several frequency presets, similarly to before. A few changes:
  - Changed "Once" preset to "Once a month". Because prompt interaction data is purged monthly, there's now no difference between "once" (implicitly per session) and "once per month".
  - Added a "Once a week" preset, because why not?
  - Removed the restriction that disallowed overlay prompts from using the "Every pageview" preset. Now, overlays can be shown to the same reader more frequently than once per 24 hours, so there's no reason to block the use of this preset.
  - Added an "Every 4th page view, up to 5x per month" preset, which matches the desired default behavior for overlay prompts as specified by Blue Lena.
  - Added a "Custom" option to let advanced users set the following new options manually. Note that if one of the other frequency presets is selected, the preset will override whatever values might exist or not exist in the various frequency option meta fields. This is so that sites that are already running prompts with one of the existing presets won't experience any unexpected changes in prompt behavior.
* **Max number of displays:** This new option defines the maximum number of times the prompt can be displayed within the specified time period. The default is `0`, which means no maximum (status quo).
* **Start after pageview:** This new option lets users specify a minimum number of pageviews a reader must have before the prompt will be displayed. For example, if this option is set to `2`, the reader will only start seeing the prompt on the third pageview. The default is `0`, which means the prompt will be displayed immediately (status quo).
* **Pageviews between displays:** This new option lets users specify the number of pageviews that must be logged in between prompt displays. For example, if this option is set to `3`, once the prompt is displayed, the reader must view three pages before the prompt can be displayed again (on the fourth view after last prompt display). The default is `0`, meaning the prompt will be displayed every page (status quo).
* **Reset counter per:** This new option lets users specify the time period after which the "Max number of displays" counter is reset. For example, if the prompt is set to show a max of 3 times per week, readers will only see the prompt a maximum of three times in a given one-week period. The default is `month`, which is status quo because prompt interaction data is purged monthly.
  - **Note:** Because specific pageview data (such as timestamp for each view) is purged hourly for performance reasons, the "Start after pageview" and "Pageviews between displays" options are based off of the reader's lifetime pageview counter. Consider the following example: a prompt is set to "Start after pageview" `2` and "Reset counter" per `month`. A reader visits the site only once, and then again after 32 days. On that second pageview after 32 days, they will NOT see the prompt because that's only the reader's second pageview. If they then visit the site again within the same month, they will see the prompt.
  - Another option is to keep detailed pageview event data for each reader for 30 days so we can reset pageview-related behavior as well, but this may result in bloated events table data.

Closes #902.

1. On `master`, publish some prompts (overlay and inline) with "Once", "Once a day", and "Every pageview" frequency options.
2. With the `NEWSPACK_EXPERIMENTAL_READER_ACTIVATION` flag OFF or not present in `wp-config.php`, check out this branch and https://github.com/Automattic/newspack-plugin/pull/1779. Test the pre-existing prompts in a new incognito session and confirm that they continue to work as expected.
3. Edit or create a new overlay prompt. Note that the frequency options behave identically as in `master`, except that the "Once" frequency is renamed "Once a month".
4. Add `define( 'NEWSPACK_EXPERIMENTAL_READER_ACTIVATION', true );` to `wp-config.php` and refresh the prompt editor. Note the new and relabeled options in the "Frequency" dropdown here and in the Campaigns wizard dashboard as well.

Screen Shot 2022-07-06 at 3 53 52 PM

5. Test each preset option in a new incognito session. Note that you won't be able to really test per day/week/month presets unless you edit the `prompt_seen` events in the events table to backdate them a day/week/month.
6. Select the "Custom" frequency option and test each new option. Again, you won't be able to really test reset option unless you edit the `prompt_seen` events in the events table to backdate them a day/week/month. Note that if you select a preset, then select "Custom", the detailed frequency options will reflect the behavior of the last-selected preset.

Screen Shot 2022-07-06 at 3 56 15 PM

7. Confirm that all unit tests pass. 👇 

### Other information:


tests/test-api.php
```diff
@@ -261,6 +262,112 @@ public function test_daily_frequency() {
...
+		);
+
+		// Report a view.
+		self::$report_campaign_data->report_campaign(
+			[
+				'cid'      => self::$client_id,
+				'popup_id' => Newspack_Popups_Model::canonize_popup_id( $test_popup['id'] ),
+			]
+		);
+
```
adekbadek, line 308: There should be an assertion here to verify the prompt is not visible in-between two consecutive page views. It currently will fail, and so does a manual test:

1. Set up a prompt with the 'Every 4th page view…' preset
2. View four articles, observe prompt displayed
3. Refresh page – prompt is displayed on every following page load (till reaching the max of 5 views)

adekbadek (changes requested):

tests/test-api.php
dkoo: This is not what I'm seeing on my end, and there is also already an assertion that tests that the prompt is not visible on two consecutive pageviews if it has a "between" value of more than 0: https://github.com/Automattic/newspack-popups/pull/908/files#diff-0eb8371b72c3645321e39c8bae29c8c4ef73c133564e8d4e09cdd952aa6f5c00R315-R318

8d9819c adds another assertion for good measure and clarifies in the code comments the difference between the pageview event and the prompt view event.

If you try this again in a new session, do you see any errors happening in the `debug` object in the API response?

dkoo: I see what you mean now! If you refresh _the same_ post, the prompt is displayed again, counting toward the max number of displays. Fixing...

dkoo: 0257027 adds a check and a test to ensure that overlay prompts are NOT shown on pages that have already been viewed in the past hour. Inline prompts are still shown on consecutive views of the same page because they take up layout space.

I'm not sure if this is the right way to handle it—does it make sense to you?

api/campaigns/class-maybe-show-campaign.php
```diff
@@ -300,21 +306,107 @@ public function should_popup_be_shown( $client_id, $popup, $settings, $referer_u
...
+			$seen_events,
+			function( $event ) use ( $frequency_reset, $now ) {
+				$seen = strtotime( $event['date_created'] );
+				return $seen >= strtotime( '-1 ' . $frequency_reset, $now );
+			}
+		);
+
+		// If not displaying every pageview.
+		if ( 0 < $frequency_between ) {
+			$views_after_start = max( 0, $total_views - ( $frequency_start + 1 ) );
```
adekbadek, line 372: When setting "Start after pageview" to `1` I'd expect the prompt to be shown _on the second pageview_. Currently it will be shown on the first one.

tests/test-api.php
adekbadek: I'm not sure about the is-repeat-view approach. I can still reproduce the issue:

1. Create a prompt to be be displayed on every second pageview ("Pageviews between displays" of 1)
2. Go to homepage, then an article, then homepage – the prompt will not be visible, while it should be
3. Go to another article, then homepage again – the prompt will be visible on both consecutive page loads, which should not happen

I think the underlying issue is that the view count is only counting unique page views, which makes the `$total_views` tally inaccurate.

adekbadek (changes requested):

api/campaigns/class-maybe-show-campaign.php
dkoo: With a "Start after pageview" setting of `1`, the desired behavior should be:

1. First pageview of a session: no prompt
2. Second pageview of the session: prompt

That's what I'm seeing locally and on a new test site. Can you describe steps to replicate other behavior?

dkoo:
> I'm not sure about the is-repeat-view approach. I can still reproduce the issue:

Seems like the ideal approach would be to still display the prompt on refreshes or repeat page views, but not capture a `prompt_seen` event so that it won't count toward the max number of displays. This way the views should always sync up with the `$total_views` count—what do you think?

dkoo:
> Seems like the ideal approach would be to still display the prompt on refreshes or repeat page views, but not capture a `prompt_seen` event so that it won't count toward the max number of displays.

This might prove challenging because the first request to log the pageview event and the second request to log the prompt view event are completely different requests, so by the time we report the prompt view, the pageview has already been logged and there's no way to tell if it was the first time or not.

tests/test-api.php
dkoo: @adekbadek [56f1ef4](https://github.com/Automattic/newspack-popups/pull/908/commits/56f1ef47bf867fa32e229bdf690f8451ff4c3e61) refactors a few things to handle repeat visits a bit more intentionally.

I don't think there's a perfect solution here due to the fact that we don't log repeat visits to the same pages within the same session (~1 hour). But I think this makes the most sense and also results in fewer "extra" overlay prompt displays:

* On repeat visits within the same hour, overlay prompts are not shown nor logged. So given our overlay prompt ("Pageviews between displays" of 1) if you visit homepage, then article 1, then homepage again, then article 2, then homepage again, it will only display on the first homepage visit and the article 2 visit.
* Inline prompts are displayed on repeat pageviews, but `prompt_seen` events aren't logged. So if our every-other-pageview prompt is an inline, it will be shown if you refresh or revisit a page within the hour, but it won't count toward max displays.

This does create different behavior between overlays vs inline prompts, but I think it's necessary to avoid showing overlays too often if the reader navigates through the same pages several times in a session (such as if they visit the homepage or a specific category repeatedly). The other option would be to retain all `view` events for a full month so we can go by that count instead of the reader's total view count, but that seems like a bad tradeoff because it could bloat the DB quite a lot on high-traffic sites.

api/campaigns/class-maybe-show-campaign.php
adekbadek: Sorry, my bad! I was experimenting with the settings end confused "Start after pageview" with "Pageviews between displays".

adekbadek:
I think the is-repeat-view logic is unnecessary here; the [bug described in my last review](https://github.com/Automattic/newspack-popups/pull/908#discussion_r921989759) is still reproducible. The frequency is based on a total view count, which currently is inaccurate – it's a count of views of unique posts/pages. The API should keep a count of _all views_, which would make the sophisticated frequency options simple to implement.

api/segmentation/class-segmentation-client-data.php
```diff
@@ -120,9 +130,10 @@ function( $order_event ) {
 				function( $acc, $order ) use ( $existing_orders ) {
```
adekbadek, line 130: Missed:

```suggestion
				function( $acc, $order ) use ( $existing_orders, $timestamp ) {
```

dkoo:
> The API should keep a count of all views, which would make the sophisticated frequency options simple to implement.

This would be a change in how we track views—we currently only track _unique_ views. If we start counting all views, then refreshing a page 10x would result in 10 views. Is this change okay for the plugin's purposes?

dkoo:
@adekbadek As discussed, 258ccd3 removes all logic that deduplicates repeat pageviews from the past session, and therefore also removes special handling for prompt display logic with respect to repeat pageviews. Now, all pageviews are logged to a reader whether or not unique.

320b17a also adds the `page` post type to the list of contexts that should not be considered "articles" for purposes of segmentation ("Articles read in past month/session").

api/segmentation/class-segmentation-client-data.php
dkoo: Fixed in 258ccd3

tests/test-api.php
adekbadek: This makes sense, thanks for the detailed explanation!

PR #911, opened 2022-07-14
feat(segmentation): handle pre-authenticated users

### LINKED ISSUE #903: Segmentation: handle pre-authentication
## As is 

The segmentation can target users who have logged in.

## To be

With the addition of pre-authentication (https://github.com/Automattic/newspack-plugin/pull/1655), ~a segmentation option should be added to allow targeting pre-authenticated/non-pre-authenticated readers~ the "has user account" segment option should also target the pre-authenticated users.

adekbadek:
### All Submissions:

Closes #903.

1. Create a prompt segmented to users who have an account
2. Enable Reader Activation (`NEWSPACK_EXPERIMENTAL_READER_ACTIVATION` env. variable)
3. In a new session, observe the prompt is not displayed
4. Pre-authenticate by entering existing user's email in Registration block
5. Navigate to another page, observe the prompt displayed

### Other information:

### DIFF
--- a/api/segmentation/class-segmentation-client-data.php
+++ b/api/segmentation/class-segmentation-client-data.php
@@ -81,23 +81,12 @@ class Segmentation_Client_Data extends Lightweight_API {
 			];
 		}
 
-		$user_account_context = false;
-
-		// Get auth intention value if they are pre-authenticated.
-		$pre_auth_value = $this->get_request_param( 'pre_auth_value', $request );
-		if ( $pre_auth_value ) {
-			$user_account_context = $pre_auth_value;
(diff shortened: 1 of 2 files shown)


dkoo (approved):
Works as described!

Is there any way to process the registration such that the reader is segmented properly on the very next pageview after registering?

adekbadek:
> Is there any way to process the registration such that the reader is segmented properly on the very next pageview after registering?

@dkoo – actually, this can be handled via a WP hook! I've reverted the original commit and took this approach. I've updated the PR description to reflect that.

PR #915, opened 2022-07-15
fix: retrieve events from persistent cache, if available

dkoo:
### All Submissions:

#869 had this working at some point, but it underwent so many changes that it got lost along the way.

When fetching and storing reader data and events, if a persistent cache is available, the plugin should attempt to store and retrieve such data in the persistent cache to minimize DB interactions. This PR restores a missing `wp_cache_set` call in the `save_reader_events` method, and also ensures that cached events that lack an `id` value (meaning they haven't yet been saved to the DB) persist in the cache across pageviews.

On sites that do not have a persistent cache available, data will still be read from and written to the DB directly.

1. Install the plugin from `master` onto a site that has memcached enabled. Publish an overlay prompt with a "Once" or "Once a day" frequency.
2. In a new session, observe that the prompt is displayed on every pageview. You can also inspect the `debug` object on the Campaigns API response to observe that the `reader_events` array is always empty—this is because events are not being properly stored in the persistent cache. This effectively means that reader events aren't "real-time" and won't be read until they're bulk-inserted into the DB from the flat log file every 15 minutes.
3. Install the plugin from this branch and repeat step 2. Navigate through a few pages to ensure that several pageviews, prompt views, and dismissals are logged. This time confirm that the prompt is displayed only once, as expected, and that the `debug` object's `reader_events` is updated whenever new events are saved.

### Other information:

### DIFF
--- a/api/classes/class-lightweight-api.php
+++ b/api/classes/class-lightweight-api.php
@@ -663,6 +663,22 @@ class Lightweight_API {
 	}
 
 	/**
+	 * Get reader events from the persistent cache, if available.
+	 *
+	 * @param string $client_id Single client ID associated with the current reader.
+	 *
+	 * @return array Array of reader events for the given client ID.
+	 */
+	public static function get_reader_events_from_cache( $client_id ) {
+		$events = wp_cache_get( 'reader_events', $client_id );
+		if ( ! $events ) {
+			$events = [];
+		}
+
+		return $events;
+	}
+
+	/**
 	 * Retrieve data for a specific reader from the cache or DB.
 	 *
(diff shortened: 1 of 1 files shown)


miguelpeixe (changes requested):
With `memcached` active I'm still getting the prompt on every pageview and also this PHP notice:

```
Undefined index: id in /var/www/wp-content/plugins/newspack-popups/api/classes/class-lightweight-api.php on line 735
```

miguelpeixe (approved):
Working as expected now!

I had to tweak my environment to properly handle `WP_Object_Cache` with memcache integration.

PR #918, opened 2022-07-20
fix(segmentation): data pruning

adekbadek:
Moves the "Remove all preview sessions data." block down, so `$old_client_ids_to_delete` is available to it. Otherwise, `PHP Notice:  Undefined variable: old_client_ids_to_delete` is logged.


dkoo:
@adekbadek good catch! Actually, the value is not needed at all in that function call because we're supposed to be deleting the preview IDs there. 1b4cd69 should be sufficient to avoid the warning and also allow the preview IDs to be properly pruned.

dkoo:
Paging @miguelpeixe for a second set of eyes to review since @adekbadek and I both contributed code.

miguelpeixe (approved):

PR #922, opened 2022-07-21
fix: simplify WC donation handling

dkoo:
### All Submissions:

I recently started running into issues getting single, non-authenticated WooCommerce donation orders logged as donation events. In the process, I decided to simplify the plugin's handling of WC donation orders in general.

#### Current state

There are two ways the plugin looks for orders:

1. If the reader completes a recurring donation, they are automatically authenticated as a WooCommerce customer. For all logged-in WP users, the plugin fetches all orders associated with the logged-in user and upserts the orders as donation events. This seems wasteful and redundant as this happens on every logged-in session page load.
2. If the reader completes a single donation while not logged in, the plugin looks for a webhook callback on `order.created` and creates a donation event with the order details. This started failing for me sometime during the past week due to unauthenticated request errors on WP's part. Perhaps something has changed to cause these webhook callbacks to be treated as unauthenticated requests?

#### Proposed simplification

Instead of handling both cases separately, the plugin instead uses the ~[`woocommerce_checkout_order_created`](https://woocommerce.github.io/code-reference/files/woocommerce-includes-class-wc-checkout.html#source-view.405)~ [woocommerce_new_order](https://woocommerce.github.io/code-reference/files/woocommerce-includes-data-stores-class-wc-order-data-store-cpt.html#source-view.89) action hook to log a donation event anytime a new order is generated by WooCommerce. In my mind this is a superior way to handle donation orders as we don't need to do any deduplication or handle recurring or single donations in a different way.

Because the webhook may still be active for sites already using Campaigns, this also takes the step of deleting it, if it exists, to avoid potential data duplication.

1. On this branch, set your Reader Revenue platform to "Newspack".
4. In new incognito sessions, test multiple single and recurring donations.
5. Confirm that `donation` events are created for each completed order per client ID.
6. Confirm also that the sessions are recognized as being part of the "Donors" segment (sees prompts targeted to a "Donors only" segment).

### Other information:

### DIFF
--- a/includes/class-newspack-popups-donations.php
+++ b/includes/class-newspack-popups-donations.php
@@ -39,6 +39,7 @@ final class Newspack_Popups_Donations {
 	public function __construct() {
 		\add_action( 'admin_init', [ __CLASS__, 'delete_legacy_wc_webhook' ] );
 		\add_action( 'woocommerce_new_order', [ __CLASS__, 'create_donation_event' ], 10, 2 );
+		\add_action( 'newspack_stripe_new_donation', [ __CLASS__, 'create_donation_event_stripe' ], 10, 3 );
 	}
 
 	/**
@@ -152,7 +153,7 @@ final class Newspack_Popups_Donations {
 			}
 
 			if ( 0 < count( $donation_events ) ) {
-				Newspack_Popups_Segmentation::update_client_data(
+				\Newspack_Popups_Segmentation::update_client_data(
 					$client_id,
 					[
 						'reader_events' => $donation_events,
@@ -160,6 +161,32 @@ final class Newspack_Popups_Donations {
 				);
 			}
 		}
+	}
+
+	/**
+	 * When a new Stripe transaction is completed with a client ID,
+	 * log a new donation event to that client ID.
+	 *
+	 * @param string      $client_id Client ID.
+	 * @param array       $donation_data Info about the transaction.
+	 * @param string|null $newsletter_email If the user signed up for a newsletter as part of the transaction, the subscribed email address. Otherwise, null.
+	 */
+	public static function create_donation_event_stripe( $client_id, $donation_data, $newsletter_email ) {
+		$client_update = [
+			'reader_events' => [
+				[
+					'type'    => 'donation',
+					'context' => 'stripe',
+					'value'   => $donation_data,
+				],
+			],
+		];
+
+		if ( $newsletter_email ) {
+			$client_update['email_subscription'] = $newsletter_email;
+		}
+
+		\Newspack_Popups_Segmentation::update_client_data( $client_id, $client_update );
 	}
 }
 Newspack_Popups_Donations::instance();
--- a/includes/class-newspack-popups-donations.php
+++ b/includes/class-newspack-popups-donations.php
@@ -143,8 +143,9 @@ final class Newspack_Popups_Donations {
 	 * When a new WooCommerce order is created with the client ID meta,
 	 * log a new donation event to that client ID.
 	 *
-	 * @param int      $order_id Order ID.
-	 * @param WC_Order $order Order object.
+	 * @param int         $order_id Order ID.
+	 * @param WC_Order    $order Order object.
+	 * @param string|null $client_id Client ID to associate with the donation. If not passed, will attempt to get from the current session.
 	 */
 	public static function create_donation_event( $order_id, $order, $client_id = null ) {
 		if ( ! $client_id ) {
--- a/includes/class-newspack-popups-donations.php
+++ b/includes/class-newspack-popups-donations.php
@@ -75,25 +75,6 @@ final class Newspack_Popups_Donations {
 	}
 
 	/**
-	 * Add a hidden billing field with client id.
-	 *
-	 * @param Array $form_fields WC form fields.
-	 */
-	public static function woocommerce_billing_fields( $form_fields ) {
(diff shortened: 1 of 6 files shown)


miguelpeixe (changes requested):
Nice! This looks like a much cleaner approach!

I'm having issues generating the donation event.

I’ve tested with a streamlined donation prompt using new emails (creates account) and existing emails (pre-auth). Also single and recurring donations.

The Stripe webhook is working and the order is created on WC.

While logging in to what goes in `woocommerce_checkout_order_created` while submitting the donation, I got nothing. Maybe this is not the hook to use?

dkoo:
Ah, perhaps [`woocommerce_new_order`](https://woocommerce.github.io/code-reference/files/woocommerce-includes-data-stores-class-wc-order-data-store-cpt.html#source-view.89) is what we need? This should theoretically fire anytime a new order is created, not just from the checkout flow.

miguelpeixe:
> When you donate via Stripe, do you get any donation event occurring?

No donation event at all, just the `prompt_seen` from the donation prompt.

miguelpeixe:
I did manage to get it working with `woocommerce_new_order` but it's bailing because `Newspack_Popups_Segmentation::NEWSPACK_SEGMENTATION_CID_NAME` is not present on the order meta.

dkoo:
Thanks for the info, will revisit!

dkoo:
@miguelpeixe Did a little refactoring. I removed the handling of client ID via hidden billing fields and instead opted to set this meta value directly based on the current session's client ID, or a client ID passed as an arg to the method. This simplifies the `Donation` class a bit, but also requires that we pass the client ID when creating a WooCommerce order synced from Stripe—because that sync happens in a Stripe webhook callback, it's not happening in the same session so we can't fetch the client ID there. Instead, we can pass down the client ID from the Stripe transaction: https://github.com/Automattic/newspack-plugin/pull/1794

dkoo:
Note: 4b6e400 and [a0534f1](https://github.com/Automattic/newspack-plugin/pull/1794/commits/a0534f199453c35dd56bbfc419fd6a9192e80d74) also refactor things so that the main Plugin fires an action hook with Stripe donation details instead of calling Campaigns methods directly—this way we can keep the implementation details in the Campaigns plugin.

includes/class-newspack-popups-donations.php
```diff
@@ -37,122 +37,156 @@ public static function instance() {
...
 	 */
-	public static function woocommerce_checkout_update_order_meta( $order_id ) {
-		if ( ! empty( $_POST[ Newspack_Popups_Segmentation::NEWSPACK_SEGMENTATION_CID_NAME ] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Missing
-			update_post_meta( $order_id, Newspack_Popups_Segmentation::NEWSPACK_SEGMENTATION_CID_NAME, sanitize_text_field( $_POST[ Newspack_Popups_Segmentation::NEWSPACK_SEGMENTATION_CID_NAME ] ) ); // phpcs:ignore WordPress.Security.NonceVerification.Missing
+	public static function create_donation_event( $order_id, $order, $client_id = null ) {
+		if ( ! $client_id ) {
+			$client_id = \Newspack_Popups_Segmentation::get_client_id();
+		}
+
+		if ( $client_id ) {
```
miguelpeixe, line 137: NIT: this could return early for readability

miguelpeixe (changes requested):
Love the refactor to use a hook on `plugin` instead of calling campaigns' method there. Much better maintenance.

I'm still unable to get the donation event, though 😕 

The `create_donation_event()` gets fired twice and, at first, with `null` client ID. Than `\Newspack_Popups_Segmentation::get_client_id();` does populate the value but still no event stored.

includes/class-newspack-popups-donations.php
dkoo: Good idea, done in c009c20

dkoo:
> 

> Love the refactor to use a hook on `plugin` instead of calling campaigns' method there. Much better maintenance.
> 
> I'm still unable to get the donation event, though 😕
> 
> The `create_donation_event()` gets fired twice and, at first, with `null` client ID. Then `\Newspack_Popups_Segmentation::get_client_id();` does populate the value but still no event stored.

The `create_donation_event` callback was getting fired twice because it was getting fired upon the new order being created by the `\WooCommerce_Connection::create_transaction` method, then again later in that same method.

I'm not sure why you're not seeing any events getting logged, but let's try once more. c009c20 and [4f34f58](https://github.com/Automattic/newspack-plugin/pull/1794/commits/4f34f58f3a95e1a1e1273f485818fcccee25a38f) refactor the WooCommerce implementation in the Plugin to use a hook as well. I also changed the hook in the Campaigns plugin back to the `woocommerce_checkout_order_created` hook, which fires the same callback.

So now `woocommerce_checkout_order_created` gets fired when a new order is created via the WooCommerce checkout flow (when Woo is the Reader Revenue platform), and `newspack_new_donation_woocommerce` gets fired when a new order is created by our Newspack Stripe sync (when Stripe is the Reader Revenue platform), and they both execute the same callback to create the `donation` event.

If you're still not seeing a `donation` event getting created, can you pinpoint where the `create_donation_event_woocommerce` callback is bailing?

miguelpeixe (approved):
Using the API methods instead of calling it through `wp_safe_remote_post()` makes much more sense! Fixed my self-signed certificate issue 🎉

PR #923, opened 2022-07-21
feat: lookup unknown email addresses in ESP to determine status

### LINKED ISSUE #893: Lookup subscriber status in ESP by email, if logged in
If a WP user logs in, and they are not marked as a subscriber in segmentation, ESP should be queried to see if they are a subscriber. They might've been signed up via other medium than the site.

dkoo:
### All Submissions:

Intercept an email address provided by the user, and if that email address is not already known to be associated with a newsletter subscription for the current reader client ID, look up the subscription status in the active ESP in Newspack Newsletters. If the email address is a known contact, log a subscription event for segmentation purposes.

This could happen if the user has subscribed to a newsletter via offsite means, or via the site prior to the Reader Activation flow.

Note: may require different hooks per https://github.com/Automattic/newspack-plugin/issues/1791.

Awaiting resolution of https://github.com/Automattic/newspack-newsletters/pull/891 in order to support all ESPs (currently only ActiveCampaign supports the `get_contact_lists` method).

Closes #893.

1. Connect an ActiveCampaign account. Set ActiveCampaign as the active ESP in **Engagement > Newsletters** and enable at least one list.
2. In the ActiveCampaign dashboard, manually add an email address as a contact to one or more lists that are enabled on the site.
3. In a new incognito session, sign up with that same email address via the Newspack Registration block without signing up for any newsletters in the block.
4. Navigate to another page and confirm that a.) you're segmented as a subscriber (see prompts targeted to subscribers only) and b.) a `subscription` event is logged to the `wp_newspack_campaigns_reader_events` table with the relevant info: `context` should be the email address, and `value` should contain the ESP slug (`active_campaign`) and an array of list IDs the email address is subscribed to.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-newsletters.php
+++ b/includes/class-newspack-popups-newsletters.php
@@ -147,8 +147,6 @@ final class Newspack_Popups_Newsletters {
 
 			$api->save_reader_events( $client_id, [ $subscription_event ] );
 		}
-
-		return;
 	}
 }
 
--- a/includes/class-newspack-popups-newsletters.php
+++ b/includes/class-newspack-popups-newsletters.php
@@ -82,12 +82,16 @@ final class Newspack_Popups_Newsletters {
 				],
 			];
 
-			\Newspack_Popups_Segmentation::update_client_data(
-				\Newspack_Popups_Segmentation::get_client_id(),
-				[
-					'reader_events' => [ $subscription_event ],
-				]
-			);
+
+			$client_id = isset( $contact['client_id'] ) ? $contact['client_id'] : \Newspack_Popups_Segmentation::get_client_id();
+			$nonce     = \wp_create_nonce( 'newspack_campaigns_lightweight_api' );
+			$api       = \Campaign_Data_Utils::get_api( $nonce );
+
+			if ( ! $api || ! $client_id ) {
+				return;
+			}
+
+			$api->save_reader_events( $client_id, [ $subscription_event ] );
 		}
 	}
 
--- a/includes/class-newspack-popups-newsletters.php
+++ b/includes/class-newspack-popups-newsletters.php
@@ -112,41 +112,43 @@ final class Newspack_Popups_Newsletters {
 			return;
 		}
 		self::$is_checking_status = true;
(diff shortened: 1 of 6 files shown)


adekbadek:
(@dkoo – the corresponding issue is already [on the project](https://github.com/orgs/Automattic/projects/412), no need to add the PR there ;) )

dkoo:
Since https://github.com/Automattic/newspack-newsletters/pull/891 gracefully fails when the active ESP doesn't support contact lookups, we can proceed with this PR. It will work when the active ESP supports the lookup, and silently skip operation if it doesn't.

https://github.com/Automattic/newspack-plugin/pull/1795 also registers anonymous readers who make a single WooCommerce donation, which is the last missing scenario where the reader may provide an email address that should be checked against the ESP contact lists in this PR.

adekbadek:
Pushed some changes in @dkoo 's absence. The check will be run only on login/registration, not on every request. Hitting the ESP API on every request is an overkill, and the check should only be run _when the email is supplied_. 

Since I've contributed, asking @miguelpeixe for a review.

includes/class-newspack-popups-newsletters.php
```diff
@@ -77,6 +99,55 @@ public static function handle_newsletter_subscription( $provider, $contact, $lis
...
+	 * of the email address with the ESP currently active in Newspack Newsletters.
+	 *
+	 * @param string|null $email_address Email address or null if not set.
+	 */
+	public static function check_reader_newsletter_subscription_status( $email_address ) {
+		if ( self::$is_checking_status ) {
+			return;
+		}
+		self::$is_checking_status = true;
+		if ( $email_address && class_exists( '\Newspack_Newsletters' ) && class_exists( '\Newspack_Newsletters_Subscription' ) ) {
```
miguelpeixe, line 115: We should exit early here, following the standard practice already present in this method.

miguelpeixe (changes requested):
Just a small thing here

includes/class-newspack-popups-newsletters.php
adekbadek: Done in e711c15f13c90cd52d4a992f79c4b42369d03790

miguelpeixe (approved):

PR #927, opened 2022-07-29
feat: read ESP donation metadata on login and supply to segmentation

adekbadek:
### All Submissions:

Adds donation-related metadata fetching from the ESP, on login.

Closes https://github.com/Automattic/newspack-issues/issues/236

1. Switch Newsletters plugin to `feat/contact-data-details` branch (https://github.com/Automattic/newspack-newsletters/pull/901)
2. Manually add a contact in ActiveCampaign ESP, and fill in `NP_Last Payment Amount` or `Total Spent` fields
3. Register on the site as this user and observe prompts segmented to donors are displayed

### Other information:

### DIFF
--- a/includes/class-newspack-popups-newsletters.php
+++ b/includes/class-newspack-popups-newsletters.php
@@ -46,7 +46,7 @@ final class Newspack_Popups_Newsletters {
 	public function __construct() {
 		\add_action( 'newspack_newsletters_add_contact', [ __CLASS__, 'handle_newsletter_subscription' ], 10, 4 );
 		\add_action( 'wp_login', [ __CLASS__, 'status_check_on_login' ], 10, 2 );
-		\add_action( 'newspack_registered_reader', [ __CLASS__, 'check_reader_newsletter_subscription_status' ] );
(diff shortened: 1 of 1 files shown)


miguelpeixe (changes requested):
The events are stored but are duplicated on every `newspack_registered_reader` trigger:

image

This hook will be triggered very often due to the optimistic quality of `register_reader()`.

miguelpeixe:
Also, do we want to use the ESP's stored date to push the event, or we don't care that much about this data?

adekbadek:
> The events are stored but are duplicated on every newspack_registered_reader trigger:

> This hook will be triggered very often due to the optimistic quality of register_reader().

Both should be mitigated by dadc6b148bea95f690970a3831ba091b63b4e58b – the ESP lookup will now only happen if it's a new reader registration. If it's an existing reader, we assume the data was already tracked. 

> Also, do we want to use the ESP's stored date to push the event, or we don't care that much about this data?

I don't think we do – and date parsing would come into play here, which would unnecessarily complicate things.

miguelpeixe (approved):

PR #930, opened 2022-08-01
fix: segment matching when no segment is the best-priority

### LINKED ISSUE #928: Logged-in users receiving prompts for "does not have account" segment
image

The above prompt is segmented for anonymous readers only and this was shown to an authenticated admin. You can see that the registration block shows as resolved, because its the default display for logged in users.

dkoo:
### All Submissions:

Fixes what seems to be a long-standing bug when evaluating whether a reader matches a prompt's segments, but none of the prompt's segments is the reader's highest-priority segment. In this case, the prompt's segmentation matching is evaluated as if it has no segments (same as "Everyone").

Closes #928.

1. On `master`, create an "Anonymous" segment targeted to readers with no account, and make it the highest-priority segment. Publish a prompt to the segment (only that segment) with "Every page" frequency. Then, create a second, lower-priority segment that will also easily match readers, such as "not subscribed" or "non-donors".
2. In a new incognito session, observe that you see the "Anonymous" prompt on every pageview before and after logging into a user account (**note:** must log in via the WP login screen, the WooCommerce My Account screen, or any page where the Campaigns plugin is disabled).
3. Check out this branch, and in a new incognito session, confirm that you only see the "Anonymous" prompt before you log in. If you add `?newspack-campaigns-debug` to your URL and inspect the `/api/campaigns/index.php` XHR request, you should be also be able to see in the `debug` response object that the prompt's segment does not match after logging in.

### Other information:

### DIFF
--- a/api/campaigns/class-maybe-show-campaign.php
+++ b/api/campaigns/class-maybe-show-campaign.php
@@ -29,6 +29,7 @@ class Maybe_Show_Campaign extends Lightweight_API {
 		$visit     = json_decode( $_REQUEST['visit'], true ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized, WordPress.Security.ValidatedSanitizedInput.InputNotValidated
 		$response  = [];
 		$client_id = $_REQUEST['cid']; // phpcs:ignore WordPress.Security.NonceVerification.Recommended, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+		$user_id   = isset( $_REQUEST['uid'] ) ? absint( $_REQUEST['uid'] ) : null; // phpcs:ignore WordPress.Security.NonceVerification.Recommended, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
 
 		$view_as_spec = [];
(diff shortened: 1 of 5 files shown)


adekbadek (changes requested):
I can't reproduce the issue described in the PR description on `master`. I've added [a comment](https://github.com/Automattic/newspack-popups/issues/928#issuecomment-1202129559) to #928 with more specific reproduction steps and a suggestion on fixing the issue.

dkoo:
@adekbadek You're right, this fix doesn't handle the scenario where in a new session, you might go directly to the WP login or My Account login screen, both of which won't execute any Campaigns code and therefore won't log the user account event before the first front-end pageview.

0052dbe should fix this by moving the logic to log user account events to the reader endpoint, so it happens earlier in the flow (as soon as the prompts are checked for segmentation matching). To test:

1. In a new incognito session, go directly to the `/wp-login.php` or `/my-account` screen and log in.
2. View a front-end article next—on `master`, you won't be recognized as having a user account, and you'll see the prompt for user anonymous users.
3. On this branch, at step 2 you should be immediately recognized as having a user account and NOT see the anonymous prompt. Conversely, if you have any prompts segmented to users with accounts, you should see those on this initial pageview.

adekbadek (approved):

PR #940, opened 2022-08-17
Clarify exclusion labels

gamebits:
The "Advanced Settings" panel of the campaign editor lets users select categories and tags to exclude. However, this panel doesn't indicate that these are _exclusion_ settings, leading some users to mistakenly input their inclusion preferences into these fields, producing just the opposite results as intended.

### All Submissions:

Renames the labels for the category and tag exclusion fields, distinguishing them from the inclusion settings farther down the screen.

1. Observe the current behavior when editing a prompt:

2. Install this branch (complete with compiled JS files).
3. Open the prompt editor. The labels "Post categories" and "Post tags" should now read "Excluded categories" and "Excluded tags".

### Other information:

### DIFF
--- a/src/editor/AdvancedSidebar.js
+++ b/src/editor/AdvancedSidebar.js
@@ -17,7 +17,7 @@ import { CategoryAutocomplete } from 'newspack-components';
 const AdvancedSidebar = ( { onMetaFieldChange, excluded_categories = [], excluded_tags = [] } ) => {
 	return (
 		<Fragment>
-			<BaseControl className="newspack-popups__advanced-sidebar">
+			<BaseControl>
 				<CategoryAutocomplete
 					label={ __( 'Excluded Categories', 'newspack-popups' ) }
 					description={ __(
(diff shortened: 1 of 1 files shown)


dkoo:
Added some further explanation using the `description` prop available on `CategoryAutocomplete`. @miguelpeixe or @adekbadek, could one of you review since I contributed code?

src/editor/AdvancedSidebar.js
```diff
@@ -17,18 +17,27 @@ import { CategoryAutocomplete } from 'newspack-components';
 const AdvancedSidebar = ( { onMetaFieldChange, excluded_categories = [], excluded_tags = [] } ) => {
 	return (
 		<Fragment>
-			<BaseControl className="newspack-popups__segmentation-sidebar">
+			<BaseControl className="newspack-popups__advanced-sidebar">
```
adekbadek, line 20: Where is this class styled? I can't see anything when searching for `__advanced`.

dkoo: Oh, I originally added some CSS to create some space in between the two fields, but decided that the `<hr />` looked better. Regardless, the class name `__segmentation-sidebar` is inaccurate and potentially confusing.

adekbadek: But the `.newspack-popups__advanced-sidebar` is not styled, so there's no reason for any `className` prop here.

dkoo: Removed in 8bd6b66

adekbadek (approved):

PR #947, opened 2022-08-25
fix: remove duplicate newspack-popups class name

### LINKED ISSUE #945: Duplicated "newspack-popup" classname
## As is 

In an overlay prompt, the top-level `amp-layout`, which has a classname of `newspack-popup`, has a child element also with `newspack-popup` classname. 

## To be

`newspack-popup` classname should be used only once in a single prompt markup.

dkoo:
### All Submissions:

Removes the extra `newspack-popups` class name from a child `div` of the outermost `newspack-popups` overlayp prompt container.

Closes #945.

There should be no functional or style changes to overlay prompts.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -1167,7 +1167,7 @@ final class Newspack_Popups_Model {
 			id="<?php echo esc_attr( $element_id ); ?>"
 		>
 			<div class="<?php echo esc_attr( implode( ' ', $wrapper_classes ) ); ?>" data-popup-status="<?php echo esc_attr( $popup['status'] ); ?>" style="<?php echo ! $hide_border ? esc_attr( self::container_style( $popup ) ) : ''; ?>">
-				<div class="newspack-popup" style="<?php echo $hide_border ? esc_attr( self::container_style( $popup ) ) : ''; ?>">
+				<div style="<?php echo $hide_border ? esc_attr( self::container_style( $popup ) ) : ''; ?>">
 					<?php if ( $has_featured_image ) : ?>
 						<div class="newspack-popup__featured-image">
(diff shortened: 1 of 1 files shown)


includes/class-newspack-popups-model.php
```diff
@@ -1167,7 +1167,7 @@ class="<?php echo esc_attr( implode( ' ', $classes ) ); ?>"
 			id="<?php echo esc_attr( $element_id ); ?>"
 		>
 			<div class="<?php echo esc_attr( implode( ' ', $wrapper_classes ) ); ?>" data-popup-status="<?php echo esc_attr( $popup['status'] ); ?>" style="<?php echo ! $hide_border ? esc_attr( self::container_style( $popup ) ) : ''; ?>">
-				<div class="newspack-popup" style="<?php echo $hide_border ? esc_attr( self::container_style( $popup ) ) : ''; ?>">
```
adekbadek, line 1170: There are a few places where a nested `.newspack-popup` is used in CSS selector, e.g.:

- [Theme patterns 7](https://github.com/Automattic/newspack-theme/blob/4ae87e805c542ef88be957fd8c426fadfe079d90/newspack-theme/sass/blocks/_patterns.scss#L68-L69) & 8
- [This plugin, in sizing handling](https://github.com/Automattic/newspack-popups/blob/72474694a59b6502a052d45351c13e15e118b3bf/src/view/style.scss#L224-L226)

adekbadek (changes requested):

includes/class-newspack-popups-model.php
dkoo: Ah, I see that it's actually used for styling. 868e742 changes the name to `.newspack-popup__content-wrapper` instead to avoid the naming collision while retaining the styles, and

adekbadek (approved):

PR #948, opened 2022-08-26
feat: tapping on overlay background no longer dismisses prompt

dkoo:
### All Submissions:

This makes it slightly harder to accidentally dismiss overlay prompts by removing the "dismiss" action from the overlay background. We've gotten feedback that it's very easy to accidentally dismiss overlays, especially on touchscreen devices where scrolling might be interpreted as a tap event that closes the overlay as soon as it appears.

@thomasguillot I wanted to tag you in case you think this warrants a redesign of the close button—we want to ensure that it's still easy enough to _intentionally_ dismiss the prompt, and not make readers hunt for the button. I think it's prominent-looking enough as-is, but it's something to consider.

1. Check out this branch, publish an overlay prompt.
2. On the front-end confirm that tapping or clicking the overlay background (the translucent part behind the prompt content) doesn't dismiss the prompt, but that tapping or clicking the X close button still does.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -1191,12 +1191,7 @@ final class Newspack_Popups_Model {
 				</div>
 			</div>
 			<?php if ( ! $no_overlay_background ) : ?>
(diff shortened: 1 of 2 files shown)


adekbadek (approved):

dkoo:
Thanks, @adekbadek. Awaiting design approval from @thomasguillot before merging.

Issue #901, opened 2022-06-29
Post-login data reconciliation


adekbadek:
Consider the following scenario:

1. User logs in (as a site subscriber) on device A and reads a few articles 
2. Then they view the site, not logged in, on a different device, B, and subscribe to the newsletter
3. They log in on device B
4. When they again browse on device A

At point 3. the anonymous data (containing the newsletter subscription) should be reconciled with the identifiable data, so that at point 4. they are considered a subscriber (for segmentation)

adekbadek:
Another case: post-pre-authentication reconciliation:

1. User reads three articles, anonymously
2. User pre-authenticaticates with the email of an existing user

At this point, their segmentation data should be reconciled. In other words, the segmentation data should be reconciled anytime an anonymous reader provides and email address.

Maybe the client-id to reader relationship should be redefined as many-to-one.

dkoo:
#978 is a PR in progress that attempts to address this issue. When a reader provides a unique identifier such as an email address or user account, we look up unique client IDs from any past reader events that share the same identifier and then update all associated reader records with all client IDs. Then, whenever getting reader events, if the reader record has multiple client IDs associated with it, we get events for all of those client IDs instead of just the current one.

There are still a few kinks to work out, but it seems to solve this issue with minimal code refactoring.

PR #954, opened 2022-09-02
Segmentation: handle former donor

adekbadek:
### All Submissions:

Adds "former donor" segmentation handling, and receiving data about a cancelled recurring donation.

_Follow testing steps in https://github.com/Automattic/newspack-plugin/pull/1966_

### Other information:

### DIFF
--- a/api/campaigns/class-campaign-data-utils.php
+++ b/api/campaigns/class-campaign-data-utils.php
@@ -372,7 +372,7 @@ class Campaign_Data_Utils {
 	/**
 	 * Get all events types.
 	 */
-	public static function get_all_events_types() {
+	public static function get_reader_events_types() {
 		return array_merge( self::get_protected_events_types(), [ 'user_account', 'view' ] );
 	}
 
--- a/api/campaigns/class-campaign-data-utils.php
+++ b/api/campaigns/class-campaign-data-utils.php
@@ -81,6 +81,29 @@ class Campaign_Data_Utils {
 	}
 
 	/**
+	 * Is reader a former donor?
+	 *
+	 * @param object $reader_events Reader data.
+	 */
+	public static function is_former_donor( $reader_events ) {
+		$donation_related_events = array_values(
+			array_filter(
+				$reader_events,
+				function( $event ) {
+					return stripos( $event['type'], 'donation' ) !== false;
+				}
+			)
+		);
+		if ( 0 < count( $donation_related_events ) ) {
+			// The donation cancellation event must be the latest one.
+			// If they've donated again, they're not a former donor.
+			$latest_donation_related_event = $donation_related_events[0];
+			return 'donation_cancelled' === $latest_donation_related_event['type'];
+		}
+		return false;
+	}
+
+	/**
 	 * Does reader have a WP user account?
 	 *
 	 * @param object $reader_events Reader data.
@@ -186,6 +209,7 @@ class Campaign_Data_Utils {
 				'is_not_subscribed'   => false,
 				'is_donor'            => false,
 				'is_not_donor'        => false,
+				'is_former_donor'     => false,
 				'has_user_account'    => false,
 				'no_user_account'     => false,
 				'referrers'           => '',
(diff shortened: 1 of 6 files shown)


miguelpeixe (changes requested):
I successfully created a segment and a prompt for "Former donors". Canceling immediately through the Stripe dashboard triggers the hook and the `donation_cancelled` event is stored to `wp_newspack_campaigns_reader_events` but the prompt never shows up. I did receive a prompt targeted to "Donors" after canceling the subscription.

api/campaigns/class-campaign-data-utils.php
```diff
@@ -340,4 +368,18 @@ public static function is_inline( $popup ) {
 	public static function is_above_header( $popup ) {
 		return isset( $popup->t ) && 'a' === $popup->t;
 	}
+
+	/**
+	 * Get all events types.
+	 */
+	public static function get_all_events_types() {
```
dkoo, line 375: "all event types" isn't entirely accurate here, as there are also `prompt_seen` and `prompt_dismissed` types. Maybe `get_all_reader_events_types`?

dkoo (changes requested):
One reason this can easily fail is if the cancellation happens in a separate session (with a different client ID) from the original donation. This is probably more likely to happen when cancelling a subscription since it might happen weeks or months later after the first donation. #901 would fix that issue.

Following the testing steps by donating in a new session, cancelling the subscription in Stripe and then continuing to navigate the site in the same session, I see that this is working.

One nit below about a method name from me, but I'd also await confirmation from @miguelpeixe on whether it works for him in the same session as well.

api/campaigns/class-campaign-data-utils.php
adekbadek: Refactored to `get_reader_events_types` in f379d0b803eefd128911c0bcbb533ec819427320

adekbadek:
> One reason this can easily fail is if the cancellation happens in a separate session (with a different client ID) from the original donation. 

Session is not a factor here – the client ID is a metadata on the Stripe customer. The session during which the cancellation happens does not matter at all.

dkoo:
Session does matter somewhat because until we solve #901 (slated for RAS phase 2 in the near future), we don't yet have a means to tie different client IDs together even if they're tied to the same email. This is the same issue as not knowing about past subscriptions/donations if you log into the same account with a different device or in a different browser session, I'm just highlighting that the particular use case of cancelling recurring donations is more likely to encounter this issue because a cancellation is likely happening weeks or months after the initial donation.

adekbadek:
I think we were talking about two different things :) By "this can easily fail" do you mean a following scenario:

1. reader donates on device A
2. they log in on device B
3. they cancel the subscription – now they're recognised as a former donor only on device B

?

dkoo (approved):
That's the scenario I was talking about, yes. It's not just different devices, but different client IDs—expected given our single client ID logic, but I think it's more likely to occur with cancellations because browser sessions/client IDs may not be likely to persist for months between donation and cancellation.

Approving because it works with the current logic, but I'd also like @miguelpeixe to confirm if it works for him or if there's another scenario that I didn't encounter.

miguelpeixe:
@dkoo did you successfully render a prompt segmented to former donors? I'm still having the exact same issue [reported before](https://github.com/Automattic/newspack-popups/pull/954#pullrequestreview-1095381630):

> I successfully created a segment and a prompt for "Former donors". Canceling immediately through the Stripe dashboard triggers the hook and the `donation_cancelled` event is stored to `wp_newspack_campaigns_reader_events` but the prompt never shows up. I did receive a prompt targeted to "Donors" after canceling the subscription.

dkoo:
> @dkoo did you successfully render a prompt segmented to former donors?

I did, which is why I approved it. The "former donors" prompt only appears in the same session as the original donation.

miguelpeixe:
I was in the same session 🤔

miguelpeixe (approved):

PR #958, opened 2022-09-09
feat: remove wp-dom-ready, medialement dependencies for frontend script

adekbadek:
### All Submissions:

Removes enqueueing of `wp-dom-ready` and `mediaelement-core` dependencies: the first is replaced by an inline function, second simply removed. 

The `mediaelement-core` dependency was added in #771, but it's not clear how the fix there works. The podcast player block is broken for me with or without the `mediaelement-core` enqueued, so the fix seems not to work anymore. I've found that the podcast player block is also broken 

- whenever Newspack Newsletters' Subscribe block assets are loaded
- on one test site, when Newspack Plugin is active at all

Which makes me think that the code of the podcast player block is too brittle to function property with Newspack. 

View a page without AMP and confirm the prompts load as expected 

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -541,7 +541,7 @@ final class Newspack_Popups_Inserter {
 			wp_register_script(
 				'newspack-popups-view',
 				plugins_url( '../dist/view.js', __FILE__ ),
-				[ 'wp-dom-ready', 'wp-url', 'mediaelement-core' ],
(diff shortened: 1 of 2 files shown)


dkoo (approved):
I can confirm that the `mediaelement-core` fix from #771 doesn't seem to work anymore. I couldn't get the podcast player to load with AMP off, even without the Campaigns plugin. CC @allykmyta who reported the original issue.

PR #951, opened 2022-08-30
feat: log reg form info to analytics events

dkoo:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Captures form info to analytics events for forms submitted via prompts. This lets us distinguish between subscribe vs register forms, and tells us whether the reader subscribed to any newsletter lists (and which ones).

### How to test the changes in this Pull Request:

1. Check out this branch, as well as https://github.com/Automattic/newspack-plugin/pull/1953 and https://github.com/Automattic/newspack-newsletters/pull/943
2. Ensure that your site is connected to Google Analytics, and monitor the Realtime events dashboard for incoming events.
3. Publish a prompt containing a registration form.
4. On the front-end, register a new email address and optionally subscribe to one or more newsletter lists.
5. In Google Analytics, look for new events with the event category `Newspack Announcement` and event action `Form Submission`. Confirm that the event label for these events contains not just the prompt info, but also the registration info, in this format:

> Prompt position: Prompt Name (prompt ID) - Segment name | Form ID - signup email - list IDs

6. Publish a prompt containing a subscribe form and repeat steps 4–5.
7. Disable the AMP plugin and repeat steps 4-6.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [ ] Have you successfully ran tests with your changes locally?

PR #960, opened 2022-09-12
fix(analytics): only add form info to form submission events

dkoo:
### All Submissions:

#951 added form info (form and list IDs) to analytics events that occur from inside a prompt. However, the placeholders for this info were added to _all_ events that are triggered by the prompt, even events that don't have anything to do with the form. This results in unprocessed placeholders being sent to GA. This PR only adds the form info to form submission events.

1. On `master`, on a site connected to Google Analytics, publish a prompt containing a registration or subscribe block.
2. On the front-end view the prompt and submit it.
3. On the Analytics side, in the Realtime Events report, observe that the `Form Submission` event has an Event Label with properly processed placeholders (e.g. `Prompt position: Prompt Name (prompt ID) - Segment name | Form ID - list IDs`), but that the `Seen` and `Load` events do not—they might appear with placeholders intact, like ``Prompt position: Prompt Name (prompt ID) - Segment name | ${formId} - ${formFields[lists[]]}`.
4. Check out this branch and repeat steps 2-3. Confirm that the `Form Submission` event still has the form ID and field IDs in the Event Label, but that all other events related to the prompt do not.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -976,7 +976,7 @@ final class Newspack_Popups_Model {
 			$analytics_events[] = [
 				'amp_on'     => 'amp-form-submit',
 				'on'         => 'submit',
-				'element'    => '#' . esc_attr( $element_id ) . ' form:not(.' . self::get_form_class( 'action', $element_id ) . ')', // Not an 'action' (dismissal) form.
+				'element'    => '#' . esc_attr( $element_id ) . ' form:not(.popup-dismiss-form)', // Not a dismissal form.
 				'event_name' => __( 'Form Submission', 'newspack-popups' ),
 			];
 		}
@@ -984,7 +984,7 @@ final class Newspack_Popups_Model {
 			$analytics_events[] = [
 				'amp_on'          => 'amp-form-submit-success',
 				'on'              => 'submit',
-				'element'         => '.' . self::get_form_class( 'dismiss', $element_id ),
+				'element'         => '#' . esc_attr( $element_id ) . ' .popup-dismiss-form',
 				'event_name'      => __( 'Dismissal', 'newspack-popups' ),
 				'non_interaction' => true,
 			];
@@ -992,7 +992,7 @@ final class Newspack_Popups_Model {
 
 		foreach ( $analytics_events as &$event ) {
 			// If a form submission and the form contains registration + list info, append that to the event label.
-			if ( 'submit' === $event['on'] && $has_register_form ) {
+			if ( isset( $event['amp_on'] ) && 'amp-form-submit' === $event['amp_on'] && $has_register_form ) {
 				$event_label .= __( ' | ${formId} - ${formFields[lists[]]}' );
 			}
 
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -972,11 +972,6 @@ final class Newspack_Popups_Model {
 			];
 		}
 
-		// If the form contains registration + list info, append that to the event label.
-		if ( $has_register_form ) {
-			$event_label .= __( ' | ${formId} - ${formFields[lists[]]}' );
-		}
-
 		if ( $has_form ) {
 			$analytics_events[] = [
 				'amp_on'     => 'amp-form-submit',
@@ -996,6 +991,11 @@ final class Newspack_Popups_Model {
 		}
 
 		foreach ( $analytics_events as &$event ) {
+			// If a form submission and the form contains registration + list info, append that to the event label.
+			if ( 'submit' === $event['on'] && $has_register_form ) {
+				$event_label .= __( ' | ${formId} - ${formFields[lists[]]}' );
+			}
+
 			$event['id']             = self::get_uniqid();
 			$event['event_category'] = esc_attr( $event_category );
 			$event['event_label']    = esc_attr( $event_label );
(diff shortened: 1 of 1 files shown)


includes/class-newspack-popups-model.php
```diff
@@ -993,6 +988,11 @@ function( $acc, $segment_id ) {
 		}
 
 		foreach ( $analytics_events as &$event ) {
+			// If a form submission and the form contains registration + list info, append that to the event label.
+			if ( 'submit' === $event['on'] && $has_register_form ) {
+				$event_label .= __( ' | ${formId} - ${formFields[lists[]]}' );
```
adekbadek, line 993: I've seen two issues: 

1. This string is attached also to the form dismissal event label, which is unnecessary
2. The above happens twice, somehow. 

Here are the last two items on `$analytics_events` in my testing:

```
    [3] => Array
        (
            [amp_on] => amp-form-submit
            [on] => submit
            [element] => #c529 form:not(.c529)
            [event_name] => Form Submission
            [id] => c27c
            [event_category] => Newspack Announcement
            [event_label] => Center: Register (343) - Anons | ${formId} - ${formFields[lists[]]}
        )

    [4] => Array
        (
            [amp_on] => amp-form-submit-success
            [on] => submit
            [element] => .c5290
            [event_name] => Dismissal
            [non_interaction] => 1
            [id] => cda1
            [event_category] => Newspack Announcement
            [event_label] => Center: Register (343) - Anons | ${formId} - ${formFields[lists[]]} | ${formId} - ${formFields[lists[]]}
        )
```

adekbadek (commented):
Nice catch!

adekbadek (commented):
And a third issue, though not directly related to these changes – a Registration form might not have the `lists[]` field (it can only be used to register, not to subscribe). There should be check for the `lists[]` field before appending the ` - ${formFields[lists[]]}` part.

includes/class-newspack-popups-model.php
dkoo: @aschweigert are these two events considered redundant? If so, which one should we keep for dismissal? We use the `Form Submission` one for non-dismissal form submissions.

aschweigert: They don't appear to be redundant...we definitely wouldn't want to consider a dismissal a successful form submission, is that what's happening currently?

dkoo: Correct, a prompt dismissal is technically considered a form submission. So it looks like we're capturing that erroneously.

aschweigert: Yep, that's very much the opposite of a successful form submission :) so we should definitely track these as separate events

dkoo: Both issues should be fixed by 1e1da62.

dkoo:
> And a third issue, though not directly related to these changes – a Registration form might not have the `lists[]` field (it can only be used to register, not to subscribe). There should be check for the `lists[]` field before appending the ` - ${formFields[lists[]]}` part. 

6e3bac5 only adds the `lists[]` field value if the prompt contains a `lists[]` field.

includes/class-newspack-popups-model.php
adekbadek: > Correct, a prompt dismissal is technically considered a form submission. So it looks like we're capturing that erroneously.

But these are two different events. On dismissal, a `Dismissal` event is sent, and on form submission a `Form Submission` event is sent. So all was well, the change in 1e1da62fbc920b77c7c1bd2862b4455c56627423 should not be needed. Or am I missing something?

adekbadek: Also, the first issue from my original comment is still reproducible – the `Dismissal` event still get the `| ${formId} - ${formFields[lists[]]}` appended. This is probably because `$event_label` is declared outside of the loop, so on iterations after it's first modified, it will still be modified.

dkoo:
@adekbadek could you have another look at this one, please?

adekbadek:
I did, and responded in [the review thread](https://github.com/Automattic/newspack-popups/pull/960#discussion_r971737957)

includes/class-newspack-popups-model.php
dkoo: > But these are two different events. On dismissal, a `Dismissal` event is sent, and on form submission a `Form Submission` event is sent. So all was well, the change in 1e1da62fbc920b77c7c1bd2862b4455c56627423 should not be needed. Or am I missing something?

@aschweigert what are your thoughts on this? It sounded to me like you were not expecting dismissals to also result in a form submission event. What is desired from an analytics standpoint?

aschweigert: @dkoo Sorry if that wasn't clear! The dismissal event should *not* be a form submission event, it should be tracked separately.

dkoo: That is currently the case, but what @adekbadek explains above is that a dismissal is technically also a form submission, so it results in two events: dismissal, and form submission. Should the form submission event not happen at all in this case?

aschweigert: ah gotcha, yeah we should definitely not fire the form submission event on the dismissal, that should make the analysis a bit easier

dkoo: @adekbadek 363829b fixes the definition of `$event_label` so that the extra form info isn't appended to non-form-submission events.

adekbadek (approved):

PR #961, opened 2022-09-13
fix: fallback to AMP script polyfills if AMP scripts are blocked

### LINKED ISSUE #939: `prompt_seen` events not logged with ad blocker extensions
Some ad blocker extensions (AdBlock, uBlock Origin) block the AMP Analytics JS from running. `prompt_seen` events are logged on the AMP Analytics custom `visible` event. So if AMP Analytics is blocked, `prompt_seen` events are never logged, and prompts will therefore be shown at a greater frequency than intended for readers using an ad blocker.

To replicate:

1. Publish an overlay prompt set to show only once per month.
2. Install and set up uBlock Origin on your browser with the default settings.
3. In a new session, visit your site and navigate around.
4. Observe that the overlay prompt is shown on every pageview rather than just once.

dkoo:
### All Submissions:

We've gotten a growing number of reports that prompts are being displayed more often than they should be. Most of these reports can likely be attributed to the AMP Analytics script being blocked by any means (browser extensions such as uBlock Origin or Ghostery, strict security settings, firewalls, etc.)—because we rely on AMP Analytics' visibility triggers to report `prompt_seen` events, if this script is blocked, the plugin will always assume that each prompt view is the reader's first. This only affects sites that are running on AMP or AMP Plus, as we have polyfills to handle the reporting for non-AMP.

This PR attempts to mitigate the issue by enqueuing the polyfill scripts in both non-AMP and AMP Plus settings, and falling back to the polyfills if the AMP scripts fail to load for any reason. This doesn't address the issue on non-AMP Plus sites because the polyfills rely on JS APIs that are not supported in `amp-script`'s [limited version of JS](https://amp.dev/documentation/components/amp-script/#supported-apis). But as most Newspack sites are moving toward AMP Plus or non-AMP, this should become less of an issue over time, and eventually when AMP is deprecated entirely, the current polyfill JS will become the de facto standard for all Newspack sites.

Closes #939 for non-AMP and AMP Plus sites.

1. Install uBlock Origin and/or Ghostery on your browser. Publish one or more prompts.
2. Enable AMP Plus with the `define( 'NEWSPACK_AMP_PLUS_ENABLED', true );` environmenet constant.
3. On `master`, view the prompts on the front-end. Observe that:
  a. In Dev Tools > Network, the `amp-analytics` script is blocked from loading.
  b. No `prompt_seen` events are logged to the `wp_newspack_campaigns_reader_events` table. (Note: if testing on a site with persistent caching enabled, change the value on [this line](https://github.com/Automattic/newspack-popups/blob/master/api/classes/class-lightweight-api.php#L39) to `true` in order to log events to the DB immediately, otherwise you'll have to wait ~15 minutes before the events move from the cache to the DB).
4. Check out this branch, repeat step 3, and confirm that even though the `amp-analytics` is still blocked, `prompt_seen` events are logged when the prompts are displayed in the viewport.
5. Turn off AMP and repeat step 4.
6. Turn off the browser extension blocking `amp-analytics` and repeat step 4, and confirm that only one `prompt_seen` event is logged per prompt view.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -78,6 +78,7 @@ final class Newspack_Popups_Inserter {
 
 		// Always enqueue scripts, since this plugin's scripts are handling pageview sending via GTAG.
 		add_action( 'wp_enqueue_scripts', [ $this, 'enqueue_scripts' ] );
+		add_filter( 'script_loader_tag', [ __CLASS__, 'mark_view_script_as_amp_plus_allowed' ], 10, 2 );
 
 		add_filter(
 			'newspack_popups_assess_has_disabled_popups',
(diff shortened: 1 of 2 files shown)


includes/class-newspack-popups-inserter.php
```diff
@@ -522,6 +523,37 @@ public static function insert_inline_prompt_in_archive_pages( $post_count ) {
...
 
+	/**
+	 * Modify the view script tag to allow it as an "AMP Plus" script.
+	 *
+	 * @param string $tag HTML of the script tag.
+	 * @param string $handle The script handle.
+	 */
+	public static function mark_view_script_as_amp_plus_allowed( $tag, $handle ) {
+		if ( 'newspack-popups-view' === $handle ) {
+			return str_replace( '<script', '<script data-amp-plus-allowed', $tag );
```
adekbadek, line 534: Nit: this could be done in `enqueue_scripts` using the script data, [like here](https://github.com/Automattic/newspack-newsletters/blob/2b52e1a0890af8680778c080de61eeffff729cc2/src/blocks/subscribe/index.php#L53). This is processed by Newspack plugin to add the inline attribute.

adekbadek (approved):

includes/class-newspack-popups-inserter.php
dkoo: Good call! b4af145 refactors to use `wp_script_add_data` and 7a69c83 makes the script `async`, which doesn't seem to have any negative effect on runtime.

PR #963, opened 2022-09-14
feat: improved frequency options no longer experimental; clearer UI

dkoo:
### All Submissions:

Moves the improved frequency options out from behind the `NEWSPACK_EXPERIMENTAL_READER_ACTIVATION` flag, replaces the preset with the one we actually went with for RAS sites, and makes the UI for custom frequency options a little bit more human-readable.

1. Check out this branch and https://github.com/Automattic/newspack-plugin/pull/2000.
2. Turn off the `NEWSPACK_EXPERIMENTAL_READER_ACTIVATION` flag in wp-config.
3. Create a new overlay prompt. In the Frequency Settings sidebar, confirm that you can select "every pageview" and "custom" options.
4. For "custom" options, confirm that there's a new toggle to enable or disable max displays. Disabling it forces the `frequency_max` meta value to `0`, or no limit, while enabling it forces the value to `1` and reveals a number control to change it.
5. Confirm that the "Start at pageview" and "Pageviews between displays" options are now no longer zero-indexed. These settings still work the same on the backend, but we simply add `1` to the value in the UI display so it's more human-readable. "Start at pageview 1" means the prompt will be displayed on the first pageview of a session (index `0`) and "Pageviews between displays 2" means the prompt will be displayed every other pageview.
6. Test the custom settings on the front-end and confirm that they work as expected.
7. Also test the new preset option "Every 4 pageviews" and confirm that it displays on the first pageview of a session, then every fourth pageview thereafter with no limit.
8. Also test the UI in the Campaigns wizard, single prompt card. The frequency options should match the ones in the single prompt editor sidebar:

Screen Shot 2022-09-14 at 3 07 27 PM

### Other information:

### DIFF
--- a/api/campaigns/class-maybe-show-campaign.php
+++ b/api/campaigns/class-maybe-show-campaign.php
@@ -325,10 +325,10 @@ class Maybe_Show_Campaign extends Lightweight_API {
 			$frequency_reset   = 'month';
 		}
 		if ( 'preset_1' === $frequency ) {
-			$frequency_max     = 5;
-			$frequency_start   = 3;
+			$frequency_max     = 0;
+			$frequency_start   = 0;
 			$frequency_between = 3;
-			$frequency_reset   = 'month';
+			$frequency_reset   = 'day';
 		}
 
 		if ( false === $now ) {
--- a/includes/class-newspack-popups.php
+++ b/includes/class-newspack-popups.php
@@ -653,7 +653,6 @@ final class Newspack_Popups {
 					)
 				),
 				'preview_query_keys'           => self::PREVIEW_QUERY_KEYS,
-				'experimental'                 => class_exists( '\Newspack\Reader_Activation' ) ? \Newspack\Reader_Activation::is_enabled( false ) : false,
 			]
 		);
 		\wp_enqueue_style(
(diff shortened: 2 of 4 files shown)


aschweigert (commented):
This isn't a very strongly held opinion, but I don't know that we necessarily need the "every 4 pageviews" option in the dropdown since that's easy enough to do with the "custom" option. Once we settle on a default I think it might make sense to add that to the dropdown, but initially here I think I'd err on the side of simplicity/fewer options.

dkoo:
> This isn't a very strongly held opinion, but I don't know that we necessarily need the "every 4 pageviews" option in the dropdown since that's easy enough to do with the "custom" option. Once we settle on a default I think it might make sense to add that to the dropdown, but initially here I think I'd err on the side of simplicity/fewer options.

That's fine—ed064f6 removes the preset option from the dropdown. I left the logic to handle it in place, though, since we're going to want to reimplement once we know what the preset values should be.

aschweigert:
Sounds good. Excited to get this out there, I think people are going to be really excited about these new options!

dkoo:
Had to close in order to allow a release build job to succeed. Will reopen.

PR #966, opened 2022-09-15
feat: improved frequency options no longer experimental; clearer UI (v2)

dkoo:
Reopening #963 which was closed due to a conflict with an ongoing `release` build.

### All Submissions:

Moves the improved frequency options out from behind the `NEWSPACK_EXPERIMENTAL_READER_ACTIVATION` flag, replaces the preset with the one we actually went with for RAS sites, and makes the UI for custom frequency options a little bit more human-readable.

1. Check out this branch and https://github.com/Automattic/newspack-plugin/pull/2000.
2. Turn off the `NEWSPACK_EXPERIMENTAL_READER_ACTIVATION` flag in wp-config.
3. Create a new overlay prompt. In the Frequency Settings sidebar, confirm that you can select "every pageview" and "custom" options.
4. For "custom" options, confirm that there's a new toggle to enable or disable max displays. Disabling it forces the `frequency_max` meta value to `0`, or no limit, while enabling it forces the value to `1` and reveals a number control to change it.
5. Confirm that the "Start at pageview" and "Pageviews between displays" options are now no longer zero-indexed. These settings still work the same on the backend, but we simply add `1` to the value in the UI display so it's more human-readable. "Start at pageview 1" means the prompt will be displayed on the first pageview of a session (index `0`) and "Pageviews between displays 2" means the prompt will be displayed every other pageview.
6. Test the custom settings on the front-end and confirm that they work as expected.
7. Also test the new preset option "Every 4 pageviews" and confirm that it displays on the first pageview of a session, then every fourth pageview thereafter with no limit.
8. Also test the UI in the Campaigns wizard, single prompt card. The frequency options should match the ones in the single prompt editor sidebar:

Screen Shot 2022-09-14 at 3 07 27 PM

### Other information:

### DIFF
--- a/api/campaigns/class-maybe-show-campaign.php
+++ b/api/campaigns/class-maybe-show-campaign.php
@@ -324,12 +324,6 @@ class Maybe_Show_Campaign extends Lightweight_API {
 			$frequency_between = 0;
 			$frequency_reset   = 'month';
 		}
-		if ( 'preset_1' === $frequency ) {
-			$frequency_max     = 0;
-			$frequency_start   = 0;
-			$frequency_between = 3;
-			$frequency_reset   = 'day';
-		}
 
 		if ( false === $now ) {
 			$now = time();
--- a/api/campaigns/class-maybe-show-campaign.php
+++ b/api/campaigns/class-maybe-show-campaign.php
@@ -325,10 +325,10 @@ class Maybe_Show_Campaign extends Lightweight_API {
 			$frequency_reset   = 'month';
 		}
 		if ( 'preset_1' === $frequency ) {
-			$frequency_max     = 5;
-			$frequency_start   = 3;
+			$frequency_max     = 0;
+			$frequency_start   = 0;
 			$frequency_between = 3;
-			$frequency_reset   = 'month';
+			$frequency_reset   = 'day';
 		}
 
 		if ( false === $now ) {
--- a/src/editor/FrequencySidebar.js
+++ b/src/editor/FrequencySidebar.js
@@ -10,6 +10,7 @@ import { Fragment } from '@wordpress/element';
 import {
 	SelectControl,
 	TextControl,
+	ToggleControl,
 	// eslint-disable-next-line @wordpress/no-unsafe-wp-apis
 	__experimentalNumberControl as NumberControl,
 } from '@wordpress/components';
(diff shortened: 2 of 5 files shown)


yogeshbeniwal:
> 5\. Pageviews between displays 2" means the prompt will be displayed every other pageview

This may be confusing for end users, "**Pageviews between displays 1 instead of 2**, means the prompt will be displayed every other pageview" will be more clear.

dkoo:
It is kind of confusing! The English language is pretty imprecise when it comes to the word "between" for counting things. We got feedback that publishers thought "Pageviews between displays 4" would mean:

Pageview 1: display
Pageview 2: no display
Pageview 3: no display
Pageview 4: no display
Pageview 5: display

Meaning that the first pageview is counted in the "between." What would be a clearer wording, in your opinion? Ideas:

* "Display every n pageviews"
* "Pageviews in between displays" - 1 would mean every other pageview. "In between" is more precise in that it indicates that the first instance is not included
* Something else?

cc @aschweigert

aschweigert:
"Display every n pageviews" is probably clearest

dkoo:
Updated to "Display every n pageviews" in f2d12f7.

api/campaigns/class-maybe-show-campaign.php
```diff
@@ -325,10 +325,10 @@ public function should_popup_be_shown( $client_id, $popup, $settings, $referer_u
 			$frequency_reset   = 'month';
 		}
 		if ( 'preset_1' === $frequency ) {
-			$frequency_max     = 5;
-			$frequency_start   = 3;
+			$frequency_max     = 0;
+			$frequency_start   = 0;
 			$frequency_between = 3;
-			$frequency_reset   = 'month';
+			$frequency_reset   = 'day';
```
adekbadek, line 331: Just noting that this will affect any prompts already set to this preset-frequency. But since the setting was experimental up till now, this is acceptable.

api/campaigns/class-maybe-show-campaign.php
```diff
@@ -325,10 +325,10 @@ public function should_popup_be_shown( $client_id, $popup, $settings, $referer_u
 			$frequency_reset   = 'month';
 		}
 		if ( 'preset_1' === $frequency ) {
```
adekbadek, line 327: Since the `preset_1` is now removed in the options, there should be a comment indicating that this is a legacy setting.

src/editor/FrequencySidebar.js
```diff
@@ -94,38 +74,47 @@ const FrequencySidebar = ( {
 					}
 
 					if ( 'preset_1' === value ) {
```
adekbadek, line 76: `preset_1` is removed from the options, so this code will never be executed and can be removed.

src/editor/FrequencySidebar.js
```diff
@@ -94,38 +74,47 @@ const FrequencySidebar = ( {
...
+						className="newspack-popups__frequency-toggle-control"
+						label={ __( 'Limit number of displays', 'newspack-popups' ) }
+						checked={ 0 < frequency_max }
+						onChange={ value => onMetaFieldChange( { frequency_max: value ? 1 : 0 } ) }
 					/>
+					{ 0 < frequency_max && (
+						<NumberControl
+							className="newspack-popups__frequency-number-control"
+							disabled={ 0 === frequency_max }
+							label={ __( 'Max number of displays', 'newspack-popups' ) }
```
adekbadek, line 99: Suggestion: put this setting after the "start" and "every" settings. This would follow the logical flow of the interaction with a prompt:

image

adekbadek (changes requested):
Note that this is a hotfix PR, while the [related Plugin PR](https://github.com/Automattic/newspack-plugin/pull/2000) is a regular one. After the hotfix is released, the UI in the Plugin will be out-of-sync until the next release.

api/campaigns/class-maybe-show-campaign.php
dkoo: This is true, but I don't believe any site was actually using the preset as it was more or less for our own convenience during beta testing.

dkoo:
> Note that this is a hotfix PR, while the [related Plugin PR](https://github.com/Automattic/newspack-plugin/pull/2000) is a regular one. After the hotfix is released, the UI in the Plugin will be out-of-sync until the next release.

Oh, bother—you're right! Since this is something we wanted to roll out to some publishers ASAP, I'll push the related Plugin PR as a new hotfix PR based against `release`.

src/editor/FrequencySidebar.js
dkoo: Good call—updated in 1f0edcd.

src/editor/FrequencySidebar.js
dkoo: I wanted to keep this code around because we do plan to re-implement presets soon, but I can understand it might cause confusion in the interim. Removed in 65f546.

adekbadek (approved):

PR #979, opened 2022-09-28
feat: treat logged-in admins and editors as preview sessions

dkoo:
### All Submissions:

Currently, logged-in admins and editor users are treated the same as readers for segmentation purposes. That is, their browsing activity and logged-in state determine which segments they match and which prompts they see when browsing the site front-end. However, admins and editors often exhibit very different browsing behavior compared to real readers, so the prompts they see while browsing can feel unexpected or incongruous.

This PR proposes that logged-in admins and editors should be treated as preview users. That is, when browsing the site, their client IDs are mocked with a `preview-` prefix, and their logged-in status is forced to not logged in for purposes of showing prompts and RAS login states. Their browsing activity (post views, term views, prompt interactions, donations, subscriptions, etc.) will be purged hourly, just like in preview sessions started from the Campaigns wizard. For RAS purposes, they will be segmented as non-registered, non-subscribers and non-donors, and will see RAS registration prompts in a pre-logged-in state (however, if they try to register or log in using the RAS UI, they will get an error message). This does mean that logged-in admins and editors will never be segmented as registered users, but this should still hopefully make the prompts shown a bit less alarming and easier to comprehend.

1. Log into a site as an admin user.
2. Browse the site front-end while logged in. Confirm that you see prompts that a "new" unknown reader would see, but that otherwise your browsing activity will change your segmentation as expected (with the exception of registering—see note above about forcing the non-logged-in state for these users).
3. Come back to the session after at least one hour and confirm that your segmentation status has been "reset" due to your admin session data being purged along with other preview data.
4. Do a smoke test of segmentation and RAS functionality in an incognito session to ensure that behavior is unchanged for non-admin users.

### Other information:

### DIFF
--- a/includes/class-newspack-popups.php
+++ b/includes/class-newspack-popups.php
@@ -739,7 +739,7 @@ final class Newspack_Popups {
 		$is_customizer_preview = is_customize_preview();
 		// Used by the Newspack Plugin's Campaigns Wizard.
 		$is_view_as_preview = false != Newspack_Popups_View_As::viewing_as_spec();
-		return ! empty( self::previewed_popup_id() ) || $is_view_as_preview || $is_customizer_preview;
+		return ! empty( self::previewed_popup_id() ) || $is_view_as_preview || $is_customizer_preview || self::is_user_admin();
 	}
 
 	/**
--- a/includes/class-newspack-popups.php
+++ b/includes/class-newspack-popups.php
@@ -739,7 +739,7 @@ final class Newspack_Popups {
 		$is_customizer_preview = is_customize_preview();
 		// Used by the Newspack Plugin's Campaigns Wizard.
 		$is_view_as_preview = false != Newspack_Popups_View_As::viewing_as_spec();
-		return ! empty( self::previewed_popup_id() ) || $is_view_as_preview || $is_customizer_preview || self::is_user_admin();
+		return ! empty( self::previewed_popup_id() ) || $is_view_as_preview || $is_customizer_preview;
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


includes/class-newspack-popups.php
```diff
@@ -739,7 +739,7 @@ public static function is_preview_request() {
 		$is_customizer_preview = is_customize_preview();
 		// Used by the Newspack Plugin's Campaigns Wizard.
 		$is_view_as_preview = false != Newspack_Popups_View_As::viewing_as_spec();
-		return ! empty( self::previewed_popup_id() ) || $is_view_as_preview || $is_customizer_preview;
+		return ! empty( self::previewed_popup_id() ) || $is_view_as_preview || $is_customizer_preview || self::is_user_admin();
```
miguelpeixe, line 742: This change is also making `Newspack_Popups::show_admin_bar()` return false. Is this intended?

miguelpeixe (changes requested):
The original intended segment/prompts behavior is working as expected in incognito, but I could not replicate the described behavior for a logged-in admin user.

Although the events appear to be stored correctly, I can't seem to get any prompt to appear for the admin user. I have tried for non-registered, non-subscriber, and also for everyone.

image

image

dkoo:
> I could not replicate the described behavior for a logged-in admin user.

Hmm, that's odd. Are you seeing an AMP Access authorization timeout error? That sometimes happens on local environments (and I've never been able to figure out why), so if that's preventing you from seeing any prompts please try on a JN site instead.

PR #980, opened 2022-09-29
fix: handle purged cache

### LINKED ISSUE #981: "Load" events fire only intermittently in AMP Plus using non-AMP ads
On sites using AMP Plus + non-AMP ads, "Load" events firing on the `ini-load` AMP Analytics trigger only happen intermittently, not on every pageview as would be expected.

Leaving the original description for the record: 

~We have a conflict between the AMP Analytics trigger `ini-load` and AMP Plus behavior, specifically when JS causes the document layout to shift (a big AMP no-no, but allowed in AMP Plus). If JS causes a content layout shift after the document loads, `ini-load` event triggers will never happen. Our "Load" events are supposed to fire on this `ini-load` trigger when a prompt is loaded and ready to display.~

To test:

1. On a test site with Google Analytics connected, enable AMP Plus and an ad in the above-header placement.
2. On the front-end, open the dev tools > network tab and filter by "Newspack Announcement" (which will display event requests to GA).
3. Observe that "Seen" events fire as expected when a prompt is shown in the viewport, but that "Load" events fire only intermittently, not on every pageview.
4. Turn off non-AMP ads by adding the `NEWSPACK_AMP_PLUS_ADS_DISABLED` environment constant to `wp-config.php`. Repeat step 2 and observe that "Load" events fire on every pageview as expected.
5. Turn off AMP Plus and repeat step 2. Observe  that "Load" events still fire on every pageview as expected.

dkoo:
### All Submissions:

Fixes a bug where reader data is not properly fetched from the DB if any reader data was previously cached but then purged. At the moment if a persistent cache is enabled, the plugin assumes that any data that exists in the cache is all that needs to be retrieved, but data in the cache is not guaranteed to persist indefinitely despite having no expiration.

Also implements a "cache version" constant so we can bust the cache for all readers by bumping the version number.

~Also adds a fix that closes #981 by polyfilling AMP Analytics whenever AMP Plus is enabled, not just if AMP Analytics isn't available. This should fix any conflicts between AMP and non-AMP JS in the AMP Plus context.~

1. On `master` using a site with a persistent cache configured, visit the front-end in a new session. Navigate around the site and take some reader actions like subscribing to a newsletter, dismissing overlay prompts, and making a donation.
2. Flush the cache using WP CLI: `wp cache flush`
3. In the same session from step 1, navigate through the site again. Turn on debug mode (`?newspack-campaigns-debug`) and observe that the reader events array only shows events that occurred after the cache flush, and that you're segmented only according to those newly cached events, rather than all past events.
4. Check out this branch and repeat step 3 in the same session. Confirm that segmentation takes into account all activity before and after the cache flush.
5. Refresh or navigate to a new page, and confirm that pre-cache-flush events* are logged to the debug object.
6. ~Also test that #981 can't be reproduced any longer ("Load" events in AMP Plus for each prompt should fire on page load if the prompt will be displayed, even if the layout shifts).~

* Note: `view` events will not persist after a cache flush, as these are temporary events. But the view counts in the `reader` object should be incremented accordingly.

### Other information:

### DIFF
--- a/src/view/utils/index.js
+++ b/src/view/utils/index.js
@@ -172,8 +172,11 @@ export const waitUntil = ( condition, callback, maxTries = 10 ) => {
  * A common scenario is amp-analytics being blocked by ad blockers or privacy browser extensions like uBlock or Ghostery.
  * In this case the custom element will exist but the service's main class will not exist on the global __AMP_SERVICES object.
  * Extensions in this object may or may not be prefixed with `amp-`: e.g. `__AMP_SERVICES.access` vs. `__AMP_SERVICES.amp-analytics`.
+ *
+ * AMP Analytics should always be polyfilled, as non-AMP JS running in AMP Plus can break AMP Analytics features.
  */
 export const shouldPolyfillAMPModule = name =>
+	'analytics' === name ||
 	undefined === customElements.get( `amp-${ name }` ) ||
 	( ! window.__AMP_SERVICES?.hasOwnProperty( `amp-${ name }` ) &&
 		! window.__AMP_SERVICES?.hasOwnProperty( name ) );
--- a/src/view/utils/index.js
+++ b/src/view/utils/index.js
@@ -172,11 +172,8 @@ export const waitUntil = ( condition, callback, maxTries = 10 ) => {
  * A common scenario is amp-analytics being blocked by ad blockers or privacy browser extensions like uBlock or Ghostery.
  * In this case the custom element will exist but the service's main class will not exist on the global __AMP_SERVICES object.
  * Extensions in this object may or may not be prefixed with `amp-`: e.g. `__AMP_SERVICES.access` vs. `__AMP_SERVICES.amp-analytics`.
- *
- * AMP Analytics should always be polyfilled, as non-AMP JS running in AMP Plus can break AMP Analytics features.
  */
 export const shouldPolyfillAMPModule = name =>
-	'analytics' === name ||
 	undefined === customElements.get( `amp-${ name }` ) ||
 	( ! window.__AMP_SERVICES?.hasOwnProperty( `amp-${ name }` ) &&
 		! window.__AMP_SERVICES?.hasOwnProperty( name ) );
--- a/api/campaigns/class-maybe-show-campaign.php
+++ b/api/campaigns/class-maybe-show-campaign.php
@@ -87,7 +87,10 @@ class Maybe_Show_Campaign extends Lightweight_API {
 
 		if ( $settings ) {
 			$settings->best_priority_segment_id = $this->get_best_priority_segment_id( $all_segments, $client_id, $referer_url, $page_referer_url, $view_as_spec );
-			$this->debug['matching_segment']    = $settings->best_priority_segment_id;
+
+			if ( $this->is_debug_enabled() ) {
+				$this->debug['matching_segment'] = $settings->best_priority_segment_id;
+			}
 		}
 
 		// Check each matching popup against other global factors.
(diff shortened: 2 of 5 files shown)


dkoo:
Note: 89f3081 removes a failing test that was incorrectly written. I'm not sure why it was passing previously, but it tests that only certain keys are allowed in the `reader_data` column of the readers table, where as that's not the case. The keys in this column are not restricted because they're keyed by post type and taxonomy type of articles viewed, and these can vary across sites.

src/view/utils/index.js
```diff
@@ -172,8 +172,11 @@ export const waitUntil = ( condition, callback, maxTries = 10 ) => {
  * A common scenario is amp-analytics being blocked by ad blockers or privacy browser extensions like uBlock or Ghostery.
  * In this case the custom element will exist but the service's main class will not exist on the global __AMP_SERVICES object.
  * Extensions in this object may or may not be prefixed with `amp-`: e.g. `__AMP_SERVICES.access` vs. `__AMP_SERVICES.amp-analytics`.
+ *
+ * AMP Analytics should always be polyfilled, as non-AMP JS running in AMP Plus can break AMP Analytics features.
```
adekbadek, line 176: Polyfilling AMP Analytics while some other plugin enqueues AMP Analytics would result in doubled pageviews & other events. And I [can't reproduce the issue](https://github.com/Automattic/newspack-popups/issues/981#issuecomment-1265372140) it's supposed to fix.

dkoo: Ah, that's a good point. We'll have to do some work to ensure that when AMP Plus is on, the Campaigns AMP Analytics JSON is not picked up by the real AMP Analytics if it also happens to be active.

dkoo: 5c3b80b reverts the polyfill—we can tackle that in a separate PR.

adekbadek (changes requested):
On a site with persistent cache, I was not able to confirm the fix. After flushing the cache and continuing browsing in the same session, the reader events were still missing in the debug data.

dkoo:
> After flushing the cache and continuing browsing in the same session, the reader events were still missing in the debug data.

Are all events flushed, or just the `view` ones? The `view` events will be purged due to our memory management policy, but the the view counts on the `reader` object should be incremented for long-term storage.

dkoo:
> Before flushing the cache, I've had a few events types in the debug response: `view`, `prompt_seen`, `user_account`, `subscription`, `donation`. After `wp cache flush`ing, I see the view counts on the reader object are correct, but in the events list there's only `user_account`. I'd expect only `view` events to be purged. 

Hmm, I see the same but I wonder if it's just a debugging issue. Can you confirm the following?

* After purging cache and refreshing, do you see the prompts you'd expect to see based on past activity (e.g. subscriber or donor prompts)?
* After refreshing again, does the debug response contain the old events? Perhaps the debug response is getting returned before some of those events are re-fetched.

adekbadek:
Ok, I confirm the events re-building works, but only for requests after the one following the cache flushing:

1. On a site with a newsletter subscription prompt, segmented for subscribers, I sign up for the newsletter. 
2. On subsequent visits I don't see the prompt
3. After flushing the cache and navigating to another post, **I do see the prompt**
4. On navigating further, I don't see the prompt

So there seems to be an order issue – the segmentation data is rebuild **after** segmentation criteria are used to determine if a prompt should be seen. 

I admit it's and improvement over not rebuilding the data at all (as it was before), but can this be fixed?

dkoo:
> So there seems to be an order issue – the segmentation data is rebuild **after** segmentation criteria are used to determine if a prompt should be seen.

Thanks for clarifying! I'll see if I can improve this. Can you clarify one more thing?

> On a site with a newsletter subscription prompt, segmented for subscribers, I sign up for the newsletter.

Do you mean this prompt should be seen by non-subscribers, or subscribers?

dkoo:
@adekbadek 30f094 should fix the cache validation so that the cache is rebuilt sooner.

adekbadek:
> > On a site with a newsletter subscription prompt, segmented for subscribers, I sign up for the newsletter.

> Do you mean this prompt should be seen by non-subscribers, or subscribers?

Sorry, error on my part – I meant "segmented for non-subscribers".

adekbadek (approved):
With the latest changes, segmentation withstands cache flushing 👌 .

PR #995, opened 2022-10-19
fix: several fixes related to AMP Plus and non-AMP

dkoo:
### All Submissions:

This includes fixes from and supersedes #991 and #994. 

Addresses several issues we've been seeing related to AMP Plus and non-AMP environments:

1. In AMP Plus sites, a race condition that causes overlay prompts to not render on the first pageview of a new session, especially on slower connections. This is fixed by enqueuing AMP scripts.
2. Seemingly in AMP Transitional sites only, the `newspack-cid` cookie is never set, resulting in segmentation behavior being totally broken as the session does not get assigned a client ID. This is also fixed by enqueuing AMP scripts.

#### Race condition

This is kind of hard to test on a local site, so a JN site is recommended.

1. Set up your JN site with Campaigns and an overlay prompt that is meant to be displayed on the very first pageview of a new session.
2. Enable AMP Plus.
3. In a new incognito session, while on `master` or `release` (e.g. without this patch), throttle your network connection to a very slow connection using dev tools.
4. Visit the site front-end and observe that the prompt is displayed inconsistently. The slower the connection, the more likely it is that the prompt is not displayed on the first pageview. (It is often shown properly after that due to AMP scripts being cached by the browser.)
5. Apply this patch and repeat in several new incognito sessions. Confirm that the prompt is always displayed.

#### AMP Transitional fix

I've found this really difficult to replicate on a local site, but it's consistently replicable and fixed by this patch on two live publisher sites. Ping me if you need more details.

1. Turn on AMP and turn off AMP Plus. In AMP > Settings, turn on Transitional mode.
2. Publish a prompt that's set to display only once per month.
3. In a new incognito session, visit the front-end.
4. In the Storage tab of dev tools, observe that no `newspack-cid` cookie is set. Also observe that the prompt appears on every pageview, as the lack of cookie means no reader events are logged.
5. Apply this patch and repeat steps 3-4. Confirm that the `newspack-cid` cookie is properly set and that the prompt appears only once per session.

### Other information:


leogermani:
We still need to confim it, but from what I saw it looks like we don't need https://github.com/Automattic/newspack-popups/pull/991. It is actually preventing page views in GA4.

My guess is that we are not firing our own page view for GA4, only for GA3... So the cause for duplicates was something else...

And one question: Do we have to worry about scenarios where the site has only GA3 or only GA4 enabled?

miguelpeixe:
Being a hotfix, shouldn't this be rebased to `release`?

dkoo:
> Being a hotfix, shouldn't this be rebased to `release`?

Yes, you're right. The branch was created off `release`, I just forgot to update the PR base.

dkoo:
> Do we have to worry about scenarios where the site has only GA3 or only GA4 enabled?

Thanks for confirming—I reverted that commit and removed from the PR description.

> We still need to confim it, but from what I saw it looks like we don't need https://github.com/Automattic/newspack-popups/pull/991. It is actually preventing page views in GA4.

This is a question for @kmwilkerson.

dkoo:
Update: a6d0409 reverts #961, which we determined was causing double pageviews to be reported on certain non-AMP sites.

miguelpeixe (approved):
🙌

PR #997, opened 2022-10-24
Invoke WP core polyfills for WP 6.1 compatibility

leogermani:
### All Submissions:

A recent change to WordPress core added a call to `str_contains` in the `wpdb` class.

See https://github.com/WordPress/wordpress-develop/commit/fed98bd9ef9a232d102c41e74944d3c21cd6183e

This function is only available in PHP 8, and WordPress has a polyfill for it in `wp-includes/compat.php`.

Since our endpoint for segmentation loads `wpdb`, it now will fail when `wpdb` tries to call `str_contains`.

In order to fix it, we are now also including the core's polyfills.

Invoke WP core polyfills for WP 6.1 compatibility

1. Upgrade your WP to the latest WP 6.1 RC release. You can use the Beta tester plugin for that
2. Set up some segmentations and prompts
3. Visit your site and inspect the response to the ajax call to `api/campaigns/index.php` or `api/segmentation/index.php`.
4. Confirm you see an error in the response `Call to undefined function str_contains`.
5. Checkout this branch and confirm the error is gone
6. Confirm Campaigns work as expected.

### Other information:

### DIFF
--- a/api/setup.php
+++ b/api/setup.php
@@ -49,6 +49,7 @@ function trailingslashit( $string ) {
 }
 
 if ( file_exists( ABSPATH . WPINC . '/wp-db.php' ) ) {
+	require_once ABSPATH . WPINC . '/compat.php';
 	require_once ABSPATH . WPINC . '/wp-db.php';
 	require_once ABSPATH . WPINC . '/functions.php';
 } else {


adekbadek (approved):
Can't reproduce the issue locally, but including the polyfills for older PHP version makes sense in principle.

dkoo (approved):
Confirmed on a staging site 👍

PR #325, opened 2020-11-27
Report segmentation-related data as custom dimension to GA

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Allows for reporting reader frequency, donor status, subscriber status, as custom dimensions to GA. 

Note that with current implementation these custom dimensions will be reported only for pageviews (not for custom events).

Closes #259 .

### How to test the changes in this Pull Request:

1. Switch Newspack Plugin to branch `add/external-custom-dimensions` (https://github.com/Automattic/newspack-plugin/pull/685)
2. Visit the Analytics Wizard, Custom Dimensions tab
3. Observe new options in the "Role" select:

image

4. Assign some custom dimensions to the Campagins segmentation related roles (the three latter ones)
5. Visit the homepage, then a post, observe the `pageview` is reported to GA with the custom dimensions, with appropriate values
6. Repeat for both AMP and non-AMP versions of the site, the custom dimensions sent with `pageview` should be the same

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [X] Have you successfully ran tests with your changes locally?

PR #1014, opened 2022-11-22
fix(ga): add custom dimensions via Newspack instead of remote config …

adekbadek:
### All Submissions:

Changes how segmentation-derived custom dimensions are supplied to GA (introduced in https://github.com/Automattic/newspack-popups/pull/325). 

The previous solution was to disable Site Kit's Analytics module's initial pageview reporting, in order to deliver the pageview to GA with this plugin, which appended segmentation-derived custom dimensions. This was error-prone, because if the site used something else on top of SiteKit to report to GA, this plugin would duplicate the pageview. 

The proposed solution uses Newspack's custom dimensions' values filtering to append the values before the page is sent to the client.  

1. Switch to `feat/custom-dimensions-filter` branch of Newspack Plugin (https://github.com/Automattic/newspack-plugin/pull/2148)
2. In Analytics wizard -> Custom Dimensions, configure two dimensions (create and assign a role in the table): one for Word count, one for Reader frequency (this one is segmentation-derived)
3. Visit a page, observe* that both custom dimensions are sent to GA** 

\* in network tab of devtools, filter by `collect` and look for a `pageview`
\** these are the `cd<index>` params in the request URL  

### Other information:


includes/class-newspack-popups-segmentation.php
```diff
@@ -151,47 +123,59 @@ public static function register_custom_dimensions( $default_dimensions ) {
...
+		$donation_events     = $api->get_reader_events( $client_id, 'donation' );
+
+		foreach ( $custom_dimensions as $custom_dimension ) {
+			// Strip the `ga:` prefix from gaID.
+			$dimension_id = substr( $custom_dimension['gaID'], 3 ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
+			switch ( $custom_dimension['role'] ) {
+				case Segmentation::CUSTOM_DIMENSIONS_OPTION_NAME_READER_FREQUENCY:
+					$read_count = Campaign_Data_Utils::get_post_view_count( $api->get_reader( $client_id ) );
+					// Tiers mimick NCI's – https://news-consumer-insights.appspot.com.
+					$read_count_tier = 'casual';
```
dkoo, line 152: This value doesn't seem to change in my testing—it always says `casual` no matter how many posts I view. I verified that the reader row in the custom table contains multiple post views:

```
+----------------------------+---------------------+---------------------+---------------------------------------------------------------------+------------+
| client_id                  | date_created        | date_modified       | reader_data                                                         | is_preview |
+----------------------------+---------------------+---------------------+---------------------------------------------------------------------+------------+
| amp-O3PbkPCQrtThfyrvy9xhqg | 2022-11-22 22:22:22 | 2022-11-22 22:25:56 | {"views":{"page":2,"post":8},"category":{"4":4,"3":4,"2":4,"12":4}} |       NULL |
+----------------------------+---------------------+---------------------+---------------------------------------------------------------------+------------+
```

dkoo (changes requested):
This is working well—however, I can't seem to get the "frequency" dimension to report more than a `casual` tier. See below:

includes/class-newspack-popups-segmentation.php
adekbadek: Thanks for catching that! The recent changes to client reconciliation changed the signature of `Campaign_Data_Utils::get_post_view_count`. Fixed in b59ab125003d39f6988548e91cc37de2e8f56367

dkoo (approved):

PR #1015, opened 2022-11-23
feat: support custom class names on inline prompt blocks

dkoo:
### All Submissions:

Adds support for the "Additional CSS Class(es)" field on Single Prompt and Custom Placement blocks.

Closes /0/1200550061930446/1203415265039373.

1. Add a Single Prompt and Custom Placement block to a post. Ensure that the latter has at least one prompt assigned to the selected placement.
2. Add class names to the Additional CSS Class(es) field:

Screen Shot 2022-11-22 at 5 28 56 PM

3. View on the front-end and inspect each element. Confirm in Dev Tools > Inspector that your class names are appended to the `aside` wrapper element for each prompt. Note that all prompts within the Custom Placement block will inherit the same class names.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -589,9 +589,15 @@ final class Newspack_Popups_Inserter {
 			return;
 		}
 
+		$class_names = '';
+
+		if ( isset( $atts['class'] ) && $atts['class'] ) {
+			$class_names .= ' class="' . $atts['class'] . '"';
+		}
+
 		// Wrapping the inline popup in an aside element prevents the markup from being mangled
 		// if the shortcode is the first block.
-		return '<aside>' . Newspack_Popups_Model::generate_popup( $found_popup ) . '</aside>';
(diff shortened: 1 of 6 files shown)


includes/class-newspack-popups-inserter.php
```diff
@@ -589,9 +589,15 @@ public static function popup_shortcode( $atts = array() ) {
 			return;
 		}
 
+		$class_names = '';
+
+		if ( isset( $atts['class'] ) && $atts['class'] ) {
```
leogermani, line 594: nit: ` ! empty( $atts['class'] ) ` does the same thing. It checks for isset and checks if the value is not empty. Just be careful that it will consider `0` as empty

leogermani (approved):
Works as described

includes/class-newspack-popups-inserter.php
dkoo: I usually use `! empty` a lot, not sure why I didn't here. Updated in badb030!

PR #1019, opened 2022-11-29
Create Json schemas

leogermani:
### All Submissions:

This PR creates a few JSON Schemas to describe the json format of the Prompts, Segments and Campaigns we will import and export.

This is a proposition based on what I could find, as the format of each entity is not formally described. The Schemas created here could also be used in the `schema` parameter of the REST api endpoints for documentation and validation if we want.

How I plan to use this:
* The importer/exporter will use these classes to validate the input json file
* Sanitize the input (with `rest_sanitize_value_from_schema`) and fill in default values when they are blank
* Santize the output json when exporting

Other notes:

* Note that the IDs (campaigns and segments IDs) that are used in the prompts will be replaced on importing. They will only have to be consistent within the json file being imported
* Not sure yet if we will import fields such as `favorite_categories` and `excluded_tags`, since they depend on terms to exist in the destination blog. We can discuss if we want to create them on the fly later...
* Question: Prompts can be assigned to Categories and Tags. Is this used? Does it define in which category/tag the prompt will be displayed? Should we add this to the schema? (considering that there is the issue of these terms having to exist in the destination blog)
* As much as possible, I made properties not required. I imagine a situation where we introduce a new property to one entity and if they were all required an old json would no longer be compatible.
* I added `additionalProperties` as `false` in all schemas. Is there any plugin adding additional properties we should consider? Is there a filter for that I should look into?

Nothing to test. Look at the Schemas and see if it has all we need, if the defaults are right, etc.

### Other information:

### DIFF
--- a/includes/schemas/class-prompts.php
+++ b/includes/schemas/class-prompts.php
@@ -250,7 +250,16 @@ class Prompts extends Schema {
 							'required' => false,
 							'default'  => [],
 							'items'    => [
-								'type' => 'integer',
+								'type'                 => 'object',
+								'additionalProperties' => false,
+								'properties'           => [
+									'id'   => [
+										'type' => 'integer',
+									],
+									'name' => [
+										'type' => 'string',
+									],
+								],
 							],
 						],
 						'excluded_tags'                  => [
@@ -259,7 +268,52 @@ class Prompts extends Schema {
 							'required' => false,
 							'default'  => [],
 							'items'    => [
-								'type' => 'integer',
+								'type'                 => 'object',
+								'additionalProperties' => false,
+								'properties'           => [
+									'id'   => [
+										'type' => 'integer',
+									],
+									'name' => [
+										'type' => 'string',
+									],
+								],
+							],
+						],
+						'categories'                     => [
+							'name'     => 'categories',
+							'type'     => 'array',
+							'required' => false,
+							'default'  => [],
+							'items'    => [
+								'type'                 => 'object',
+								'additionalProperties' => false,
+								'properties'           => [
+									'id'   => [
+										'type' => 'integer',
+									],
+									'name' => [
+										'type' => 'string',
+									],
+								],
+							],
+						],
+						'tags'                           => [
+							'name'     => 'tags',
+							'type'     => 'array',
+							'required' => false,
+							'default'  => [],
+							'items'    => [
+								'type'                 => 'object',
+								'additionalProperties' => false,
+								'properties'           => [
+									'id'   => [
+										'type' => 'integer',
+									],
+									'name' => [
+										'type' => 'string',
+									],
+								],
 							],
 						],
 						'duplicate_of'                   => [
--- a/includes/schemas/class-prompts.php
+++ b/includes/schemas/class-prompts.php
@@ -1,0 +1,283 @@
+<?php
+/**
+ * The Prompts Schema
+ *
+ * @package Newspack
+ */
+
+namespace Newspack\Campaigns\Schemas;
+
+use \Newspack\Campaigns\Schema;
+use Newspack_Popups_Model;
+
+/**
+ * The Prompts Schema
+ */
+class Prompts extends Schema {
+
+	/**
+	 * Get the schema.
+	 *
+	 * @return array The schema.
+	 */
+	public function get_schema() {
+		return [
+			'type'                 => 'object',
+			'additionalProperties' => false,
+			'properties'           => [
+				'title'           => [
+					'name'     => 'title',
+					'type'     => 'string',
+					'required' => true,
+				],
+				'content'         => [
+					'name'     => 'content',
+					'type'     => 'string',
+					'required' => true,
+				],
+				'campaign_groups' => [
+					'name'     => 'campaign_groups',
+					'type'     => 'array',
+					'required' => false,
+					'items'    => [
+						'type' => 'integer',
+					],
+					'default'  => [],
+				],
+				'meta'            => [
+					'type'                 => 'object',
+					'required'             => true,
+					'additionalProperties' => false,
+					'properties'           => [
+						'background_color'               => [
+							'name'     => 'background_color',
+							'type'     => 'string',
+							'format'   => 'hex-color',
+							'required' => false,
+							'default'  => '#FFFFFF',
+						],
+						'display_title'                  => [
+							'name'     => 'display_title',
+							'type'     => 'boolean',
+							'required' => false,
+							'default'  => false,
+						],
+						'hide_border'                    => [
+							'name'     => 'hide_border',
+							'type'     => 'boolean',
+							'required' => false,
+							'default'  => false,
+						],
+						'large_border'                   => [
+							'name'     => 'large_border',
+							'type'     => 'boolean',
+							'required' => false,
+							'default'  => false,
+						],
+						'frequency'                      => [
+							'name'     => 'frequency',
+							'type'     => 'string',
+							'required' => true,
+							'enum'     => [
+								'once',
+								'weekly',
+								'daily',
+								'always',
+								'custom',
+							],
+
+						],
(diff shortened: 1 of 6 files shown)


includes/schemas/class-prompts.php
```diff
@@ -0,0 +1,283 @@
...
+							'name'     => 'no_overlay_background',
+							'type'     => 'boolean',
+							'required' => false,
+							'default'  => false,
+						],
+						'placement'                      => [
+							'name'     => 'placement',
+							'type'     => 'string',
+							'required' => true,
+							'enum'     => [
```
dkoo, line 156: Non-blocking: There are some other properties which interact with the various placement options. For example, if the prompt's placement is an overlay placement, then the `trigger_type` property cannot be `blocks_count`, which applies to inline prompts only. These conflicts should be handled by the [`Newspack_Popups_Model::generate_popup()`](https://github.com/Automattic/newspack-popups/blob/master/includes/class-newspack-popups-model.php#L1125) and [`generate_inline_popup()`](https://github.com/Automattic/newspack-popups/blob/master/includes/class-newspack-popups-model.php#L1062) methods (which will set reasonable default behaviors if they encounter unexpected property values), but wanted to state here that the schema makes it possible to create impossible configs, so we'll want to make sure we handle any possible conflicts upon prompt creation.

dkoo (changes requested):
I love the approach to create the schemas first! Some nitpicks below, some of which are non-blocking.

To answer your questions:

> Not sure yet if we will import fields such as `favorite_categories` and `excluded_tags`, since they depend on terms to exist in the destination blog. We can discuss if we want to create them on the fly later...

This can probably be left out for now. One thing we could consider is having the schema accept term config arrays containing a slug and a taxonomy, and then trying to match those up to an existing term upon import (and creating the term if none is found). But that seems like it could get messy if the slugs don't match exactly.

> Question: Prompts can be assigned to Categories and Tags. Is this used? Does it define in which category/tag the prompt will be displayed? Should we add this to the schema? (considering that there is the issue of these terms having to exist in the destination blog)

We do use this, exactly as you describe: if a prompt is assigned categories and/or tags, it will only be rendered on posts/archives that match one of those assigned terms. Again, I don't think we need to include it in the schema now until we resolve the question about term slugs/IDs.

> As much as possible, I made properties not required. I imagine a situation where we introduce a new property to one entity and if they were all required an old json would no longer be compatible.

This sounds like the right approach to me. Note that the [`Newspack_Popups_Model::get_popup_options()`](https://github.com/Automattic/newspack-popups/blob/master/includes/class-newspack-popups-model.php#L349) method already handles setting default/missing values. I don't think there's any harm in duplicating the defaults in the schema or even replacing the default values in the method with the schema defaults.

> I added `additionalProperties` as `false` in all schemas. Is there any plugin adding additional properties we should consider? Is there a filter for that I should look into?

You're just missing one filter that I know of, called out below.

includes/schemas/class-schema.php
```diff
@@ -0,0 +1,73 @@
...
+	 * Constructor.
+	 *
+	 * @param mixed $value The value to be validated.
+	 */
+	public function __construct( $value ) {
+		$this->value  = $value;
+		$this->errors = array();
+		$this->schema = array(
+			'type' => 'string',
+			'enum' => array( 'foo', 'bar' ),
```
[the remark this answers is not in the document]
leogermani: This shouldn't be there... it's a left over from a previous attempt

includes/schemas/class-prompts.php
leogermani: Yes, I saw that.

The methods you link are called when the popup is being rendered, right? So they could not be used for validation...

In any case, I think we can handle that in a `sanitize` method in the Schema class. Before saving or exporting, we check all these internal dependencies and fix them... we could add these checks to the `is_valid` method as well...

leogermani:
> I love the approach to create the schemas first! Some nitpicks below, some of which are non-blocking.
> 
> To answer your questions:
> 
> > Not sure yet if we will import fields such as `favorite_categories` and `excluded_tags`, since they depend on terms to exist in the destination blog. We can discuss if we want to create them on the fly later...
> 
> This can probably be left out for now. One thing we could consider is having the schema accept term config arrays containing a slug and a taxonomy, and then trying to match those up to an existing term upon import (and creating the term if none is found). But that seems like it could get messy if the slugs don't match exactly.

I thought about that, but considered that in a real life scenario I would not want to create categories on my site based on the prompts... The prompts should adapt to the categories I use in my site, and not the other way around.

Maybe we could add an option in the importer to map categories to existing categories - like the WP Importer does with authors... 

Anyway, for now, I think it's a good idea to add name and slug to the json.

> 
> > Question: Prompts can be assigned to Categories and Tags. Is this used? Does it define in which category/tag the prompt will be displayed? Should we add this to the schema? (considering that there is the issue of these terms having to exist in the destination blog)
> 
> We do use this, exactly as you describe: if a prompt is assigned categories and/or tags, it will only be rendered on posts/archives that match one of those assigned terms. Again, I don't think we need to include it in the schema now until we resolve the question about term slugs/IDs.

ok, thanks

> 
> > As much as possible, I made properties not required. I imagine a situation where we introduce a new property to one entity and if they were all required an old json would no longer be compatible.
> 
> This sounds like the right approach to me. Note that the [`Newspack_Popups_Model::get_popup_options()`](https://github.com/Automattic/newspack-popups/blob/master/includes/class-newspack-popups-model.php#L349) method already handles setting default/missing values. I don't think there's any harm in duplicating the defaults in the schema or even replacing the default values in the method with the schema defaults.
> 

Yes, that's where I got the default values from.

I tried to reuse existing methods at first to have a picture of the schema... or to use them to generate the schema dynamically, but it was very hard...

I thought we could reuse this json schema not only for the importer/exporter, but also for the REST API validation (I noticed that there's no validation in the value the api endpoints receive)... And maybe with time we can centralize all this validation/sanitization in these classes...

But please feel free to point a different direction if you think there's a better/easier solution

includes/schemas/class-schema.php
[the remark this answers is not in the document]
leogermani: https://github.com/Automattic/newspack-popups/pull/1019/commits/74a2a9c0f50e7e2b7ebbab4af259d877c63c0030

includes/schemas/class-prompts.php
dkoo: Yes, those methods will render the actual markup. I think it's okay to proceed for now as any "impossible" configs imported via JSON will be fixed by these methods when the prompts are rendered, but I agree that abstracting them to a `sanitize` or `is_valid` method is probably the better way to go, long-term.

dkoo:
> I tried to reuse existing methods at first to have a picture of the schema... or to use them to generate the schema dynamically, but it was very hard...

> I thought we could reuse this json schema not only for the importer/exporter, but also for the REST API validation (I noticed that there's no validation in the value the api endpoints receive)... And maybe with time we can centralize all this validation/sanitization in these classes...

I think this is a sound approach for another PR. For now, it's enough to ensure that the defaults are recorded in the schemas, so that we can refer to them from there in the future.

dkoo (approved):
This is looking good to me. As discussed throughout, I think in general this PR does a great job of establishing the schemas for campaigns, segments, and prompts and matching their default values to the various methods sprinkled throughout. In the future we can rely on the schemas more for validation/sanitization.

leogermani:
@dkoo thanks

in https://github.com/Automattic/newspack-popups/pull/1019/commits/95258de29029a8d0ed37757ba9ade04f2cc2f4db I just updated the tags and categories schema to include ID and name, and then we can decide how to treat them in the future

dkoo:
@leogermani thanks for the heads up. In the future we might want to consider updating the actual prompt meta that currently stores only term IDs to match the schema, but this looks good for now.

PR #1021, opened 2022-12-02
Add Exporter class

leogermani:
### All Submissions:

1. Make sure to run `composer install` to update the autoloader
2. Tests are passing?
3. Test it on a site with existing Campaigns, prompts and segments
4. Run `wp newspack-popups export`
5. Check the output the the json file created
6. It should contain all your data
7. Check if the `totals` have the correct information with the total number of exported items
8. Modify your database to create an invalid data

* For example, set the `overlay_size` to something that is not a valid value.
* Set the 'options->excluded_tags` with an ID of a tag that does not exist.

9. Export again
10. Check that the invalid items are not exported
11. Check the errors array containing the respective errors

### Other information:


includes/class-newspack-popups-exporter.php
```diff
@@ -0,0 +1,263 @@
...
+	 *
+	 * Transforms an array of terms in an array in the format expected by the Exporter, containing only the term ID and name.
+	 *
+	 * @param int[]|WP_Term[] $terms Array of terms. If term must be either an integer with the term ID, or a WP_Term object.
+	 * @param string          $taxonomy Taxonomy slug. Used to fetch the term name if only an ID is informed.
+	 * @return array
+	 */
+	private function sanitize_terms( $terms, $taxonomy ) {
+		$sanitized_terms = [];
+		foreach ( $terms as $term ) {
```
adekbadek, line 243: For this line, the following warning is logged: `PHP Warning:  foreach() argument must be of type array|object, bool given`

includes/class-newspack-popups-exporter.php
```diff
@@ -0,0 +1,263 @@
...
+	private function get_segments() {
+		$segments        = [];
+		$stored_segments = Newspack_Popups_Segmentation::get_segments();
+		foreach ( $stored_segments as $stored_segment ) {
+			$val = new Newspack\Campaigns\Schemas\Segments( $stored_segment );
+			if ( $val->is_valid() ) {
+				$segments[] = $this->prepare_segment_for_export( $stored_segment );
+				$this->totals['segments']++;
+			} else {
+				$this->errors['segments'][ $stored_segment->term_id ] = $val->get_errors();
```
adekbadek, line 139: For this line, the following warning is logged: `PHP Warning:  Attempt to read property "term_id" on array`

adekbadek (changes requested):
1. I've four prompts on a site, the output of the exporter lists the same error for each: `[options][utm_suppression] is not of type string.`. I've not edited this prompt option on any of these. 
2. I've one segment on a site. The export contains the following error for it: `created_at is not a valid property of Object.`
3. This feature is tricky to test without any interface. How about adding a simple WP CLI command, which would output a JSON-formatted string?

leogermani:
Thanks for having a look! I'll check the points you raised

>     3. This feature is tricky to test without any user interface. How about adding a simple WP CLI command, which would output a JSON-formatted string?

Yes, that's the idea, I was planning to do this as the next PR, so they didn't get too big. But if you think it's a good idea we could it all in the same PR

leogermani:
@adekbadek I have added a simple CLI command and fixed the valdiation issues

leogermani:
I will look at the PHP warnings first thing tomorrow

adekbadek (changes requested):
1. When running the command after pulling the branch, I got `Error: Callable "Newspack\\Campaigns\\CLI\\Export" does not exist, and cannot be registered as wp newspack-popups export` error, so I've ran `composer install`
2. Now it errors out with `Error: 'newspack-campaigns' is not a registered wp command. See 'wp help' for available commands.` 
3. The PR description should be updated with the testing instructions :)

leogermani:
@adekbadek Also please note that I just added a quick and simple command for not. Strings are not i18ned, we might think of a better name, example, parameters... this was just to help with testing

includes/cli/class-export.php
```diff
@@ -0,0 +1,63 @@
...
+	 * Export Campaigns, Segments and Prompts.
+	 *
+	 * ## OPTIONS
+	 *
+	 * [--file=<file>]
+	 * : The name of the output file.
+	 *
+	 * ## EXAMPLES
+	 *
+	 *     wp newspack-campaigns export --file=newspack-campaigns.json
```
adekbadek, line 29: > The command is newspack-popups and not newspack-campaigns

I was following the code comment, which should be updated: 

```suggestion
	 *     wp newspack-popups export --file=newspack-campaigns.json
```

adekbadek (approved):
> I did [update PR description] yesterday! And included the composer install step.

Sorry, I see it now – I've got only my dysfunctional brain to blame! 

There's just one small change needed in the CLI command code comment.

PR #1029, opened 2022-12-14
Update/refine cli commands

leogermani:
### All Submissions:

Refines the CLI commands giving more options, localizing it and also fixing some issues in the importer / exporter

Please use this Pull request to test the full extent of the importer / exporter feature

1. Create a (or using an existing site) with many prompts, campaigns and segments
2. Make sure you have prompts in different states (publish, draft, etc)
3. Make sure to use tags and categories in all possible fields (tags and categories in prompts, excluded_tags and excluded_categories in prompts and favorite_categories in segments)
4. Run the exporter CLI command
5. does the doc and the messaging look good?
6. Now fire a brand new site
7. Create at least one of the categories that existed in the site you did the export (just one!)
8. Create a new category with a different name - that didn't exist in the site you did the export
9. Look at the importer docs `wp newspack-popups export --help`
10. can you understand it? can you improve it? any typos?
11. Run the importer
12. You are going to be prompted to what to do about the missing categories and tags
13. Try different options for each one. See how the interaction feels
14. Choose at least one dfifferent option once
15. When you choose to map to an existing category, choose the category you created in step 8
16. You should NOT be asked about the category you created in step 7
17. After the import is complete, go to the admin and check things were created correctly
18. Check all tags and categories in all fields
19. on a new site, again, run the importer and test the `--skip-terms` and `--create-terms` flags.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -259,10 +259,9 @@ final class Newspack_Popups_Model {
 			'post_type'      => Newspack_Popups::NEWSPACK_POPUPS_CPT,
 			'posts_per_page' => 1,
 			'p'              => $post_id,
+			'post_status'    => $include_drafts ? [ 'draft', 'pending', 'future', 'publish' ] : 'publish',
+
 		];
-		if ( false === $include_drafts ) {
-			$args['post_status'] = 'publish';
-		}
 
 		$popups = self::retrieve_popups_with_query( new WP_Query( $args ) );
 		return count( $popups ) > 0 ? $popups[0] : null;
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -251,7 +251,7 @@ final class Newspack_Popups_Model {
 	 * Retrieve popup CPT post by ID.
 	 *
 	 * The query for prompts relies on the dynamic default value of the post_status parameter.
-	 * In admin context, it will include the drafts, in non-admin context is will be just published posts.
+	 * In admin context, it will include the drafts, and in non-admin context it will only return published posts.
 	 *
 	 * @param string $post_id Post id.
 	 * @param bool   $use_default_status_query Whether to rely on the default behavior of the post_status parameter. If false, only published posts will be returned.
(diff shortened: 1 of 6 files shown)


includes/class-newspack-popups-model.php
```diff
@@ -259,10 +259,9 @@ public static function retrieve_popup_by_id( $post_id, $include_drafts = false )
 			'post_type'      => Newspack_Popups::NEWSPACK_POPUPS_CPT,
 			'posts_per_page' => 1,
 			'p'              => $post_id,
+			'post_status'    => $include_drafts ? [ 'draft', 'pending', 'future', 'publish' ] : 'publish',
```
leogermani, line 262: @adekbadek Can you help me having a look here?

This change breaks one test, but I wonder if the problem is in the test or in the logic here.

The problem I was trying to solve is that if I call `retrieve_popup_by_id` with `$include_drafts` as `true`, drafts are not being returned!

I added a `test_retrieve_popup_by_id` in the `test-model.php` file that passes with my change and fails if you try to revert the changes in this method.

adekbadek: The code on `master` relies on the [dynamic default set for the query](https://developer.wordpress.org/reference/classes/wp_query/#status-parameters). In admin context, it will include the drafts, in non-admin context is will be just published posts. The argument name is misleading, should rather be called `$use_default_post_status` or something. For _actual_ draft-inclusion, there could be another argument, or a larger refactor that would make the code behave as suggested by the argument name.

leogermani: Gotcha, thanks for looking and clarifying, I was very confused to see it behaving differently.

I'll think of a solution

leogermani: And tests for the win!

leogermani: I have submitted a fix. Let me know what you think @adekbadek

adekbadek: Looks good!

includes/cli/class-import.php
```diff
@@ -32,26 +43,85 @@ class Import extends WP_CLI_Command {
...
+							WP_CLI::success( __( 'Ignoring', 'newspack-popups' ) );
+							$done = true;
+
+						}
+					}
+				}
+			}
+		}
+
+		WP_CLI::success( __( 'Running the importer', 'newspack-popups' ) );
```
adekbadek, line 116: Shouldn't this just be a message?

```suggestion
		WP_CLI::log( __( 'Running the importer', 'newspack-popups' ) );
```

adekbadek (approved):
Left one non-blocking comment. This works really nice!

PR #1032, opened 2022-12-16
feat: add ras defaults preset

leogermani:
### All Submissions:

Closes # .

1. Checkout this branch, make sure to run `composer install`
2. Run `wp newspack-popups import --ras-defaults`
3. Confirm that a Campaign named "Reader Activation Defaults" was created with a set of prompts and Segments
4. Check pemhSX-2m-p2 and verify everything was created as expected

### Other information:

### DIFF
--- a/includes/cli/class-import.php
+++ b/includes/cli/class-import.php
@@ -51,12 +51,12 @@ class Import extends WP_CLI_Command {
 		$ras_defaults = WP_CLI\Utils\get_flag_value( $assoc_args, 'ras-defaults', false );
 		$file         = $args[0] ?? null;
 
-		if ( ! $file && ! $ras_defaults ) {
-			WP_CLI::error( __( 'You must either supply a json file path or use the --ras-defaults flag.', 'newspack-popups' ) );
-		}
-
 		if ( $ras_defaults ) {
 			$file = dirname( NEWSPACK_POPUPS_PLUGIN_FILE ) . '/presets/ras-defaults.json';
+		}
+
+		if ( ! $file && ! $ras_defaults ) {
+			WP_CLI::error( __( 'You must either supply a json file path or use the --ras-defaults flag.', 'newspack-popups' ) );
 		}
 
 		if ( ! is_readable( $file ) ) {
--- a/includes/cli/class-import.php
+++ b/includes/cli/class-import.php
@@ -55,7 +55,7 @@ class Import extends WP_CLI_Command {
 			$file = dirname( NEWSPACK_POPUPS_PLUGIN_FILE ) . '/presets/ras-defaults.json';
 		}
 
-		if ( ! $file && ! $ras_defaults ) {
+		if ( ! $file ) {
 			WP_CLI::error( __( 'You must either supply a json file path or use the --ras-defaults flag.', 'newspack-popups' ) );
 		}
 
--- a/includes/cli/class-import.php
+++ b/includes/cli/class-import.php
@@ -26,7 +26,7 @@ class Import extends WP_CLI_Command {
 	 *
 	 * ## OPTIONS
 	 *
-	 * <file>
+	 * [<file>]
 	 * : The name of the input file.
 	 *
 	 * [--ignore-terms]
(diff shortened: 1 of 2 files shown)


includes/cli/class-import.php
```diff
@@ -45,7 +48,16 @@ class Import extends WP_CLI_Command {
 	public function __invoke( $args, $assoc_args ) {
 		$ignore_terms = WP_CLI\Utils\get_flag_value( $assoc_args, 'ignore-terms', false );
 		$create_terms = WP_CLI\Utils\get_flag_value( $assoc_args, 'create-terms', false );
-		$file         = $args[0];
+		$ras_defaults = WP_CLI\Utils\get_flag_value( $assoc_args, 'ras-defaults', false );
+		$file         = $args[0] ?? null;
+
+		if ( ! $file && ! $ras_defaults ) {
+			WP_CLI::error( __( 'You must either inform a json file name or use the --ras-defaults flag.', 'newspack-popups' ) );
```
adekbadek, line 55: Seems more appropriate:

```suggestion
			WP_CLI::error( __( 'You must either supply a json file path or use the --ras-defaults flag.', 'newspack-popups' ) );
```

includes/cli/class-import.php
```diff
@@ -45,7 +48,16 @@ class Import extends WP_CLI_Command {
 	public function __invoke( $args, $assoc_args ) {
 		$ignore_terms = WP_CLI\Utils\get_flag_value( $assoc_args, 'ignore-terms', false );
 		$create_terms = WP_CLI\Utils\get_flag_value( $assoc_args, 'create-terms', false );
-		$file         = $args[0];
+		$ras_defaults = WP_CLI\Utils\get_flag_value( $assoc_args, 'ras-defaults', false );
+		$file         = $args[0] ?? null;
+
+		if ( ! $file && ! $ras_defaults ) {
```
adekbadek, line 54: Suggestion – moving this check below the `$file` assignment below would make this `$file` check more universal. `$file` would be either user-supplied JSON or a preset, in case more presets are added in the future.

presets/ras-defaults.json
```diff
@@ -0,0 +1,334 @@
...
+                "referrers": "",
+                "referrers_not": ""
+            },
+            "id": "639b337fdbf33",
+            "priority": 2
+        }
+    ],
+    "campaigns": [
+        {
+            "id": 30,
```
adekbadek, line 330: Will the importer handle a possible campaign ID conflict?

presets/ras-defaults.json
```diff
@@ -0,0 +1,334 @@
...
+                "is_subscribed": false,
+                "is_not_subscribed": true,
+                "is_donor": false,
+                "is_not_donor": false,
+                "is_former_donor": false,
+                "is_logged_in": false,
+                "is_not_logged_in": true,
+                "favorite_categories": [],
+                "referrers": "",
+                "referrers_not": ""
```
adekbadek, line 322: Question indirectly related to this PR: shouldn't the importer fill in the default values? That'd keep the JSONs  more readable.

adekbadek (commented):
Works well, left some minor comments & questions. As for the configuration itself is seems all according to the plan, but I think @dkoo could weigh in just to double-check.

presets/ras-defaults.json
leogermani: yes, the id only serves the purpose of creating the relation between prompts and campaigns, it's not used when creating the term in the destination site

presets/ras-defaults.json
leogermani: yes, good point. That's an improvement I want to do to the schema classes

includes/cli/class-import.php
leogermani: done

adekbadek: What I meant was something different – since `$file` may be set on line 55, later you can just do:

```
	if ( ! $file ) {
			WP_CLI::error( __( 'You must either supply a json file path or use a preset flag.', 'newspack-popups' ) );
		}
```

But it's a nitpick!

leogermani: ah of course

dkoo (approved):
Works well, and the segments + prompts have the correct configuration! 👍

PR #1040, opened 2023-01-11
feat: new option to show close button only on form submit

dkoo:
### All Submissions:

This implements a highly-requested feature for some advanced publishers who want to start soft-gating content behind email collection or registration actions. It introduces a new option in the Advanced Settings sidebar for overlay prompts which lets publishers specify that the prompt can't be dismissed until a form inside the prompt is submitted.

Screen Shot 2023-01-11 at 11 51 01 AM

The intended use case is for an overlay prompt targeted to unregistered or unsubscribed readers who have read a minimum number of articles. Once they've viewed the requisite number of articles, we can show an "undismissible" prompt to encourage registration or newsletter signup in order to continue viewing article content. Once the reader takes the required action, the prompt becomes dismissible again.

This is a fairly rudimentary and soft registration wall, as it's intentionally easy to defeat. If an overlay prompt contains a form element and is set to "Dismiss Action: Form Submission", its close button will be hidden upon initial render until the reader submits the form. Because we don't know what form could be inside the prompt or the behavior of the form after initial submit, we only listen for the `submit` action, not the successful async response to the submission, which would greatly increase the complexity of the feature. Tech-savvy readers can also use dev tools to defeat the prompt, as the article content is still rendered in the HTML underneath the overlay.

Other caveats:

* The feature is somewhat hidden in the Advanced Settings sidebar as it's an advanced feature that should be used with caution.
* Form submission is the only alternative dismiss action supported at this time. We may choose to add others in the future.
* The close button can only be hidden if the prompt contains a form. If a prompt is set to "Dismiss Action: Form Submission" but doesn't contain a form, the close button behavior will revert to the regular behavior (close button is shown on initial render).
* The feature depends on JavaScript and is therefore only available on for AMP Plus and non-AMP pages. On pages rendered with full AMP standard mode, the close button behavior will revert to regular behavior.

1. On an AMP Plus or non-AMP site, create an overlay prompt and observe a new "Dismiss Action" field under Advanced Settings, set to "Close button" by default (status quo).
2. Set the Dismiss Action to "Form Submission" and add a block that will render a `form` element in the prompt content (such as a Registration or Newsletter Subscription Form block). Publish the prompt.
3. View the prompt on the front-end and confirm that the prompt renders with its close button hidden.
4. Submit the form and confirm that the close button becomes available upon form submission.
5. Remove the form block from the prompt content and repeat step 3, but this time confirm that the close button is shown on initial render (because there's no form to submit, which would make the prompt impossible to dismiss)
6. Add the form block again but configure your site to use AMP (not AMP Plus). Repeat step 3 again, and confirm that the close button is shown on initial render (front-end JS isn't supported in AMP).

### Other information:

### DIFF
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -1191,9 +1191,10 @@ final class Newspack_Popups_Model {
 
 		// If the dismiss action is by form submission, dismiss the prompt when the form is submitted.
 		if ( 'form_submission' === $dismiss_action ) {
+			$site_is_amp         = Newspack_Popups_Inserter::is_amp() && ! Newspack_Popups_Inserter::is_amp_plus();
 			$newspack_form_class = apply_filters( 'newspack_campaigns_form_class', 'newspack-subscribe-form' );
 			$has_form            = preg_match( '/<form\s|mc4wp-form|\[gravityforms\s|' . $newspack_form_class . '/', $body ) !== 0;
-			if ( $has_form ) {
+			if ( $has_form && ! $site_is_amp ) {
 				$body = str_replace(
 					'<form',
 					'<form data-submit-success on="submit-success:' . esc_attr( $element_id ) . '-close.show"',
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -380,6 +380,7 @@ final class Newspack_Popups_Model {
 			'additional_classes'             => get_post_meta( $id, 'additional_classes', true ),
 			'excluded_categories'            => get_post_meta( $id, 'excluded_categories', true ),
 			'excluded_tags'                  => get_post_meta( $id, 'excluded_tags', true ),
+			'dismiss_action'                 => get_post_meta( $id, 'dismiss_action', true ),
 		];
 
 		// Remove empty options, except for those whose value might actually be 0.
@@ -425,6 +426,7 @@ final class Newspack_Popups_Model {
 				'additional_classes'             => '',
 				'excluded_categories'            => [],
 				'excluded_tags'                  => [],
+				'dismiss_action'                 => 'close_button',
 			]
 		);
 	}
@@ -1176,6 +1178,7 @@ final class Newspack_Popups_Model {
 		$wrapper_classes       = [ 'newspack-popup-wrapper' ];
 		$wrapper_classes[]     = 'publish' !== $popup['status'] ? 'newspack-inactive-popup-status' : null;
 		$is_scroll_triggered   = 'scroll' === $popup['options']['trigger_type'];
+		$dismiss_action        = $popup['options']['dismiss_action'];
 
 		add_filter(
 			'newspack_analytics_events',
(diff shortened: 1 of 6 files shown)


includes/class-newspack-popups-model.php
```diff
@@ -1222,7 +1250,26 @@ class="<?php echo esc_attr( implode( ' ', $classes ) ); ?>"
 			</div>
 			<?php if ( ! $no_overlay_background ) : ?>
 				<div style="opacity: <?php echo floatval( $overlay_opacity ); ?>;background-color:<?php echo esc_attr( $overlay_color ); ?>;" class="newspack-lightbox-shim"></div>
-				</form>
+			<?php endif; ?>
+			<?php if ( 'form_submission' === $dismiss_action ) : ?>
+				<script data-amp-plus-allowed type="text/javascript">
```
adekbadek, line 1255: There's the `src/view` directory for front-end JS. Why should this feature rely on an inline script?

includes/class-newspack-popups-model.php
```diff
@@ -1186,6 +1189,27 @@ function ( $evts ) use ( $popup, $body, $element_id ) {
 
 		$animation_id = 'a_' . $element_id;
 
+		// If the dismiss action is by form submission, dismiss the prompt when the form is submitted.
+		if ( 'form_submission' === $dismiss_action ) {
+			$site_is_amp         = Newspack_Popups_Inserter::is_amp() && ! Newspack_Popups_Inserter::is_amp_plus();
+			$newspack_form_class = apply_filters( 'newspack_campaigns_form_class', 'newspack-subscribe-form' );
+			$has_form            = preg_match( '/<form\s|mc4wp-form|\[gravityforms\s|' . $newspack_form_class . '/', $body ) !== 0;
```
adekbadek, line 1196: Both MC4WP and GF plugins' output an HTML form. Isn't `|mc4wp-form|\[gravityforms\s` in the RegExp redundant?

includes/class-newspack-popups-model.php
```diff
@@ -1222,7 +1250,26 @@ class="<?php echo esc_attr( implode( ' ', $classes ) ); ?>"
 			</div>
 			<?php if ( ! $no_overlay_background ) : ?>
 				<div style="opacity: <?php echo floatval( $overlay_opacity ); ?>;background-color:<?php echo esc_attr( $overlay_color ); ?>;" class="newspack-lightbox-shim"></div>
-				</form>
+			<?php endif; ?>
+			<?php if ( 'form_submission' === $dismiss_action ) : ?>
+				<script data-amp-plus-allowed type="text/javascript">
+					var s = document.querySelector('form[data-submit-success]');
```
adekbadek, line 1256: This selector will select the first matching form on the page – if there are multiple prompts with a form, the listener might be attached to a different form.

includes/class-newspack-popups-model.php
```diff
@@ -1222,7 +1250,26 @@ class="<?php echo esc_attr( implode( ' ', $classes ) ); ?>"
...
+					var d = document.getElementById('<?php echo esc_attr( $element_id . '-close' ); ?>');
+					if ( s && d ) {
+						s.addEventListener('submit',function(e){
+							var observer = new MutationObserver(function(){
```
adekbadek, line 1260: Why is a `MutationObserver` needed here? Shouldn't the `hidden` attribute be removed simply on submission?

adekbadek (changes requested):
I'm not sure about the UX here. As a reader, I'd assume a registration-wall to disappear right after registering. Or that at least a prominent button would be displayed in place of the registration form. The subtle appearance of the close icon is too little as a means of removing the wall, I think.

dkoo:
I think I pushed this prematurely—I had done most of it before the break and it was working locally, but I think I missed a step where I wanted to vet the approach with the rest of the team first. Let me take this back to draft mode for now so we can have that discussion.

PR #1049, opened 2023-02-02
feat: add feature to disable reader data caching via flag

dkoo:
### All Submissions:

Adds a feature to bypass reader and reader event data caching via a URL param or environment flag. Cached data can sometimes become unreliable as a reader amasses many client IDs, and this offers us a way to bypass the cache.

1. On a site with memcache setup, check out this branch and browse around the site as a reader.
2. Observe that after the first pageview, on subsequent pageviews your reader data and events are fetched from and written to the persistent object cache instead of directly to the DB.
3. Add a `?nocache` param to the URL and refresh. Confirm that reader data and events are fetched directly from the DB when this param is prseent.
4. Add `define( 'NEWSPACK_POPUPS_NOCACHE', true );` to your `newspack-popups-config.php` file and repeat step 3 without the param. Confirm that the cache is bypassed in the same way for every pageview when this environment variable is set.

### Other information:

### DIFF
--- a/api/classes/class-lightweight-api.php
+++ b/api/classes/class-lightweight-api.php
@@ -105,6 +105,11 @@ class Lightweight_API {
 		if ( ! file_exists( WP_CONTENT_DIR . '/object-cache.php' ) || ( defined( 'IS_TEST_ENV' ) && IS_TEST_ENV ) ) {
 			$this->ignore_cache = true;
 		}
+
+		if ( $this->is_cache_disabled() ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+			$this->ignore_cache     = true;
+			$this->debug['nocache'] = true;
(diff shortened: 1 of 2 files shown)


dkoo:
> I can't confirm that. Any more details on how I could test?

With memcache enabled, when you visit the site as a reader your reader events should be written to a flat file at `/tmp/newspack-popups-events.log` instead of being immediately written to the DB. This flat file is parsed to the `wp_newspack_campaigns_reader_events` table every 15 minutes. Some reader data in the `wp_newspack_campaigns_readers` table may be written immediately upon the pageview request, but the flat file has been sufficient to prevent too many concurrent DB writes from happening at once.

leogermani (approved):
Sorry it took me so long! I confirm everything working as expected now!

PR #1055, opened 2023-02-13
fix: avoid notice on php 8

leogermani:
### All Submissions:

Fixes the method signature so it won't throw notices on PHP 8+

1. Spin up a site with php 8 and Campaigns
2. Visit the site and inspect the ajax requests to `.../wp-content/plugins/newspack-popups/api/campaigns/index.php?cid=...`
3. Check the response and confirm you see the following warning:
```
<br />
<b>Deprecated</b>:  Required parameter $client_id follows optional parameter $all_segments in <b>/newspack-repos/newspack-popups/api/campaigns/class-maybe-show-campaign.php</b> on line <b>185</b><br />
```
4. Checkout this branch and confirm the warning is gone.

### Other information:

### DIFF
--- a/api/campaigns/class-maybe-show-campaign.php
+++ b/api/campaigns/class-maybe-show-campaign.php
@@ -182,7 +182,7 @@ class Maybe_Show_Campaign extends Lightweight_API {
 	 *
 	 * @return string|null ID of the best matching segment, or null if the client matches no segment.
 	 */
-	public function get_best_priority_segment_id( $all_segments = [], $client_id, $referer_url = '', $page_referer_url = '', $view_as_spec = false ) {
+	public function get_best_priority_segment_id( $all_segments, $client_id, $referer_url = '', $page_referer_url = '', $view_as_spec = false ) {
 		// If using "view as" feature, automatically make that the matching segment. Otherwise, find the matching segment with the best priority.
 		if ( $view_as_spec && isset( $view_as_spec['segment'] ) ) {
 			return $view_as_spec['segment'];
--- a/api/campaigns/class-maybe-show-campaign.php
+++ b/api/campaigns/class-maybe-show-campaign.php
@@ -272,7 +272,7 @@ class Maybe_Show_Campaign extends Lightweight_API {
 					$should_display = true;
 				}
 			}
-		} elseif ( $should_display && ! empty( $popup_segment_ids ) ) {
+		} elseif ( $should_display && ! empty( $popup_segment_ids ) && ! empty( $settings ) ) {
 			// $settings->best_priority_segment_id should always be present, but in case it's not (e.g. in a unit test), we can fetch it here.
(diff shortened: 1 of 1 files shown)


adekbadek (commented):
Is it certain that this argument will always be an array?

leogermani:
> Is it certain that this argument will always be an array?

There is no check for that. But that issue is beyond this PR. We know for sure that this argument is always passed, so the default vaule of `[]` is never used (because the second argument is required).

This method could certainly have more checks in place, but that should be treated in a dedicated effort.

adekbadek:
If the argument is passed, but has a `null` value, won't that result in a regression since it won't be coerced to and array anymore?

leogermani:
> If the argument is passed, but has a null value, won't that result in a regression since it won't be coerced to and array anymore?

afaik it  won't be coerced to an array. If you pass a NULL value, you passed something, so it will still be NULL. The only case in which the parameter is filled with the default value is when it is not informed

adekbadek:
What I mean is that this argument may not always be provided, that's why the default value is in place. When removing that default value, we should ensure that the argument is always provided. 

If you're previewing a prompt, the `$settings` supplied to the `should_popup_be_shown` method, which calls `get_best_priority_segment_id`, [will be empty](https://github.com/Automattic/newspack-popups/blob/a443ecaf84d3464c3c9f44e519aef83b0b0dfea5/includes/class-newspack-popups-inserter.php#L808-L816). This actually results in a PHP warning on `master`, because the empty settings are read by the "lightweight API"! 

5a859396e6399575b90b9b6b7550ddb2c6d89e9a ensures that the empty `$settings` (when previewing) won't be read, and thus the empty `$all_segments` argument will not be passed to `get_best_priority_segment_id`.

adekbadek (approved):
Assuming @leogermani 's merge will be an approval of the change I've added (5a859396e6399575b90b9b6b7550ddb2c6d89e9a) :)

leogermani:
> What I mean is that this argument may not always be provided, that's why the default value is in place. When removing that default value, we should ensure that the argument is always provided.

I don't get it. The reason of the warning is because the second argument is required. So it's currently impossible to call this method without providing the first argument, because you have to inform the second - and it's impossible to inform the second without informing the first. :rofl: 

If there's a case throwing warning because of an empty value (which is different from not passing the value) it is not related to the change in this PR. Nothing would change. 

But if we can fix that, let's do it

adekbadek:
Ah, got you now. Yes, the argument will always be **provided** – even if the value is not an array, as expected. So my initial comment was misguided because I was not clear on whether the argument would be coerced into the default value (it won't). 

Anyway, by looking into this I've found an issue which would cause a PHP warning when previewing prompts – fixed in 5a859396e6399575b90b9b6b7550ddb2c6d89e9a. So this PR is now fixing **two** PHP warnings :)

PR #1078, opened 2023-03-28
feat: add user_input fields to prompt schema

dkoo:
### All Submissions:

Adds an optional `user_inputs` array to the prompts schema. If these are populated in a prompt's config JSON, they can be used to populate placeholders by field name in the prompt's content. In the future, the array of fields can be used to create actual input fields in the default prompts UI, the values of which can be used to create prompt config JSON to generate the prompts with user input.

Also adds `featured_image_id` to the schema so that the importer can generate a prompt with a given attachment ID, which will be another feature of the default prompts UI.

1. On a fresh site, run `wp newspack-popups import --ras-defaults`.
2. Compare the generated prompts with the config JSON in `./presets/ras-defaults.json`. The `{{placeholder}}` values in the prompts' content should be populated by the default values of the corresponding user input fields. In the future, an optional `value` property can store the user's real input.
3. You can play around with the config or create your own JSON config with other placeholders and user inputs, and different `default` or `value` properties, to ensure that the placeholders are always populated by the corresponding `user_input` field values.
4. Also test the featured image functionality by hard-coding a `featured_image_id` property for a prompt to import, and confirm that the prompt is created with the given attachment ID as featured image.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-importer.php
+++ b/includes/class-newspack-popups-importer.php
@@ -228,9 +228,17 @@ class Newspack_Popups_Importer {
 		$prompts = $this->pre_process_prompts_terms( $prompts );
 
 		foreach ( $prompts as $prompt ) {
+			$prompt_content = $prompt['content'];
+			$user_inputs    = isset( $prompt['user_inputs'] ) ? $prompt['user_inputs'] : [];
+
+			foreach ( $user_inputs as $user_input ) {
+				$value          = ! empty( $user_input['value'] ) ? $user_input['value'] : $user_input['default'];
+				$prompt_content = str_replace( '{{' . $user_input['name'] . '}}', $value, $prompt_content );
+			}
+
 			$post_data = [
 				'post_title'   => $prompt['title'],
-				'post_content' => $prompt['content'],
+				'post_content' => $prompt_content,
 				'post_status'  => $prompt['status'],
 				'post_type'    => Newspack_Popups::NEWSPACK_POPUPS_CPT,
 			];
@@ -252,6 +260,11 @@ class Newspack_Popups_Importer {
 			}
 			if ( ! empty( $prompt['tags'] ) ) {
 				Newspack_Popups_Model::set_popup_terms( $new_post, $prompt['tags'], 'post_tag' );
+			}
+
+			// If there's a featured image.
+			if ( ! empty( $prompt['featured_image_id'] ) && false !== wp_get_attachment_url( (int) $prompt['featured_image_id'] ) ) {
+				set_post_thumbnail( $new_post, (int) $prompt['featured_image_id'] );
 			}
 		}
 	}
--- a/includes/schemas/class-prompts.php
+++ b/includes/schemas/class-prompts.php
@@ -25,17 +25,22 @@ class Prompts extends Schema {
 			'type'                 => 'object',
 			'additionalProperties' => false,
 			'properties'           => [
-				'title'           => [
+				'title'             => [
 					'name'     => 'title',
 					'type'     => 'string',
 					'required' => true,
 				],
-				'content'         => [
+				'content'           => [
 					'name'     => 'content',
 					'type'     => 'string',
 					'required' => true,
 				],
-				'campaign_groups' => [
+				'featured_image_id' => [
+					'name'     => 'featured_image_id',
+					'type'     => 'integer', // Attachment ID to be used as the featured image.
+					'required' => false,
+				],
+				'campaign_groups'   => [
 					'name'     => 'campaign_groups',
 					'type'     => 'array',
 					'required' => false,
@@ -53,7 +58,7 @@ class Prompts extends Schema {
 						],
 					],
 				],
-				'status'          => [
+				'status'            => [
 					'name'     => 'status',
 					'type'     => 'string',
 					'required' => true,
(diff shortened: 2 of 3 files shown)


includes/class-newspack-popups-importer.php
```diff
@@ -228,9 +228,17 @@ private function process_prompts() {
 		$prompts = $this->pre_process_prompts_terms( $prompts );
 
 		foreach ( $prompts as $prompt ) {
+			$prompt_content = $prompt['content'];
+			$user_inputs    = isset( $prompt['user_inputs'] ) ? $prompt['user_inputs'] : [];
+
+			foreach ( $user_inputs as $user_input ) {
+				$value          = ! empty( $user_input['value'] ) ? $user_input['value'] : $user_input['default'];
+				$prompt_content = str_replace( '{{' . $user_input['name'] . '}}', $value, $prompt_content );
```
leogermani, line 236: shouldn't we crop the string if manx_length is set?

includes/schemas/class-prompts.php
```diff
@@ -355,6 +360,61 @@ function( $item ) {
...
+							'description' => [
+								'type' => 'string',
+							],
+							// Whether the input is required to generate the prompt or if it can be generated with an empty value. If true, the default value will be used when the input is empty.
+							'required'    => [
+								'type' => 'boolean',
+							],
+							// Default value of the input. This will be populated by default in the UI.
+							'default'     => [
+								'type' => [
```
leogermani, line 395: From what I see in the doc, and in the defaults below, this is always a string, isn't it?

Up there we are using the value to replace placeholders in the content, so we expect a string.

In what cases `default` and `value` are not strings?

leogermani (approved):
Added two non-blocking comments, as they can be addressed in follow up PRs.

Maybe it's still too early to know the best approach, but the actual values of the user input could be informed nas an additional parameter to the Popups_Importer constructor.

If we'll be storing the progress of the user inputs in the database somewhere as an option or user_meta, when they click to generate the prompt we could then invoke the importer and pass the data.

The way it's now we would have to parse the json and edit the object to add the values. If we do that, we could just edit the `defaults` as well... At a first sight, it looks to me that it makes more sense to send the user input in another way, and fallback to the default in the json if they are not informed...

But again, no need to decide it now, I think when we start to develop the UI and the data handling this will become more clear

includes/class-newspack-popups-importer.php
dkoo: Yes, but I was planning to leave this up to the React (UI) part of things.

includes/schemas/class-prompts.php
dkoo: Yeah, I wasn't sure if there would ever be a case in which we'd want to add a user input that's a checkbox or other type of field, so I left the types as mixed. However, we could decide now that this feature will only be used for string-based placeholder replacements, and lock it down to `string` only. What do you think?

includes/class-newspack-popups-importer.php
leogermani: Makes sense. But since you can also run the importer via CLI, I think we should make sure the back-end also handles it

includes/schemas/class-prompts.php
leogermani: Let's start with what we have and evolve as needed.

If we need other type of inputs, we will also need to treat them differently when running the importer, so I guess we can leave it for later

includes/class-newspack-popups-importer.php
dkoo: Perhaps better to do it in both places, though. Added in 269e0fc

includes/schemas/class-prompts.php
dkoo: 269e0fc sets the `type` of `default` to `string` only.

dkoo:
>The way it's now we would have to parse the json and edit the object to add the values. If we do that, we could just edit the defaults as well... At a first sight, it looks to me that it makes more sense to send the user input in another way, and fallback to the default in the json if they are not informed...

That's kind of what I was planning to do, but I felt the defaults would be useful to retain in case the user input value was empty or otherwise invalid. But I can see it being cleaner to pass the user input values as an additional parameter in the Importer class constructor. I think the inputs should be stored in `wp_options` until the prompts are created, so this would work well. 269e0fc scaffolds this so that the importer class constructor can receive a `user_inputs` param that contains an object containing each field's value keyed by field name. To avoid confusion I've renamed the schema param to `user_input_fields` so that it's clearer that these config object describe the fields and not the values.

leogermani:
> but I felt the defaults would be useful to retain in case the user input value was empty or otherwise invalid.

Yes, the defaults still make sense to live in the json. I think it's looking great!

dkoo:
Great! I think we'll need to figure out a better structure for the user inputs that come into the Importer constructor so that we can tie input values to individual prompts, but that can come in a later PR. Do think it makes sense to add an optional unique slug to the prompt schema so that we can identify the prompt and its expected user inputs by purpose (e.g. `ras_registration_overlay`)?

leogermani:
> Do think it makes sense to add an optional unique slug to the prompt schema so that we can identify the prompt and its expected user inputs by purpos?

Yes, definitely! Good call

PR #1079, opened 2023-03-29
fix(memberships): restriction check for inline prompt insertion

miguelpeixe:
### All Submissions:

The paywalled post check is using the `wc_memberships_is_post_content_restricted()` function incorrectly. This will always be true if the post is restricted to any user and does not take the current user into account. 

This bug was found while working [the same fix for another feature](https://github.com/Automattic/newspack-plugin/pull/2366#issuecomment-1489119379).

This PR improves the check with the `wc_memberships_view_restricted_post_content` capability check.

1. Install and configure Woo Memberships while in the master branch
2. Setup an always visible inline prompt set to position 0
3. Login with an account that has a plan (can be the admin account)
4. Confirm the inline prompt does not render
5. Checkout this branch and confirm the prompt renders
6. Log out, visit the page and confirm the prompt does not render

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -418,6 +418,24 @@ final class Newspack_Popups_Inserter {
 	}
 
 	/**
+	 * Whether the given post is being restricted by Woo Memberships.
+	 *
+	 * @param int $post_id The post ID.
+	 */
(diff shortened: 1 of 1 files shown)


leogermani (approved):
it works!

PR #1087, opened 2023-04-05
feat: start proof-of-concept for localStorage API

dkoo:
### All Submissions:

Proof of concept of a `localStorage`-based reader data model + segmentation API.

This proof of concept only explores a potential data model for reader data cached in `localStorage`, and APIs to read/write reader data and match it to prompt segments. It handles reader data storage and retrieval for anonymous readers only, and segmentation criteria for newsletter signup, donation, and article views. Things it doesn't handle (not exhaustive):

* User login and storage to user meta
* Segmentation by user account, favorite categories, or referral URL
* Deprecation of server-based API or AMP elements
* Analytics
* Frequency-based prompt display ("display prompt every _n_ pageviews")
* Scroll-based overlays
* Favorite category updates
* Prompt interaction events (loaded/seen/dismissed)
* Newsletter signup or donation events (upon submitting signup/donation forms)
* Newsletter signup or donation lookup (for existing subscribers/donors)
* Previewing

1. Create some segments and prompts based on article views, or "not a subscriber"/"not a donor"/"no user account" criteria.
2. Confirm that the prompts display as expected according to the highest-priority matching segment.

### Other information:


src/view/segmentation.js
```diff
@@ -0,0 +1,248 @@
...
+	) {
+		matches = false;
+	}
+	if (
+		configuration.max_posts &&
+		configuration.max_posts > 0 &&
+		articleViewCount > configuration.max_posts
+	) {
+		matches = false;
+	}
```
miguelpeixe, line 92: I know it's a draft but I wanted to register that we might wish to use a more abstract logic for implementing the matching logic for segmentation criteria.

In the spirit of having criteria as an extensible aspect of the plugin, we can think of them as something you configure and register, along with its matching logic:

```js
// Article view count criteria registration
const criteria = {
	name: 'Article view count',
	// ... other necessary metadata
	matchCallback: () => {
		const articleViewCount = readerData.store.get( 'articleViewCount' ) || 0;
		return true;
	}
}
registerCriteria( 'articleViewCount', criteria );
```

The `registerCriteria()` function would store a map of criteria. The `doesReaderMatchSegment()` function would read from that map to filter the relevant criteria for the given segment and execute each criteria callback.

miguelpeixe: The criteria could even be the one creating any necessary data, either from an `init` callback in its registration schema or as part of its overall logic, imagining it as something located in a `criteria/articleViewCount.js` file. 🤔

miguelpeixe: In addition to that, in the same discussion but for a separate PR, we should also consider a criteria registration for the backend in the same fashion

dkoo: This is a great idea. I think I grasp what you're suggesting and will give it a go, but I might huddle with you if I run into any blockers since you seem to have a strong idea of how we could set this up.

dkoo:
Closing in favor of #1169

PR #1090, opened 2023-04-12
feat: support RAS default prompts wizard UI

dkoo:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

Supports new RAS-related setup UI in https://github.com/Automattic/newspack-plugin/pull/2399. The main changes:

- Updates the prompt schema to support additional user input field data types (array and integer in addition to string)
- Implements logic to store user inputs by field name and populate preset config placeholders with matching stored inputs when fetching the preset configs
- Implements logic to build dynamic prompt previews from preset config + user input, without having to create prompt posts

### How to test the changes in this Pull Request:

Follow testing instructions in https://github.com/Automattic/newspack-plugin/pull/2399. The plugin should function as usual and all tests should continue passing even without the changes in https://github.com/Automattic/newspack-plugin/pull/2399.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [ ] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1096, opened 2023-04-19
feat: add help info for preset prompts

dkoo:
### All Submissions:

Adds static help copy and screenshot assets for prompt presets and schema. See https://github.com/Automattic/newspack-plugin/pull/2415.

Also fixes a regression introduced in [49cdc86](https://github.com/Automattic/newspack-popups/pull/1090/commits/49cdc86dab259324a5c72a74f492da9c86ffef90) in which featured images are not included in dynamic preset previews.

Also tweaks the order of RAS presets to match the UI designs.

See instructions in https://github.com/Automattic/newspack-plugin/pull/2415.

### Other information:


presets/ras-defaults.json
```diff
@@ -105,7 +105,18 @@
                     "required": true,
                     "default": []
                 }
-            ]
+            ],
+            "help_info": {
+                "screenshot": "reg-overlay.png",
```
miguelpeixe, line 110: I think it's best if this is left empty, if the image is not available it'll cause a 404. The prompt UI shouldn't render anything in that case.

miguelpeixe (approved):
🙌 

Non-blocking suggestion:

presets/ras-defaults.json
dkoo: Ah, good call! d92742 removes the `screenshot` property if it can't find the referenced filename in the `src/assets` folder, which should avoid rendering the `img` tag and making a 404 request.

PR #1121, opened 2023-05-24
fix: attempt to create config file on every host

miguelpeixe:
### All Submissions:

Tweaks the lightweight API config file creation to attempt on every host. The logic to skip if set as env var is preserved but all config variables will be checked, instead of just `DB_CHARSET`, `WP_CACHE_KEY_SALT`, and `DB_PREFIX`.

1. On a fresh instance, install the plugin and confirm the file is created
2. Remove the created file and set your database credentials as env vars, if not possible through runtime configuration, [`putenv()`](https://www.php.net/manual/en/function.putenv.php) is convenient
3. Refresh wp-admin and confirm the file is created but the env vars were skipped

### Other information:

### DIFF
--- a/includes/class-newspack-popups.php
+++ b/includes/class-newspack-popups.php
@@ -874,23 +874,54 @@ final class Newspack_Popups {
 	 * Create the config file for the API, unless it exists.
 	 */
 	public static function create_lightweight_api_config() {
-		// Don't create a config file if not on Newspack's Atomic platform, or if there is a file already.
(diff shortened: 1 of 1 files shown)


leogermani (approved):
Works fine

I tried to find some context on why we  were limiting this file creation only to atomic sites but didn't find anything -> https://github.com/Automattic/newspack-popups/pull/228

I think this should be fine

PR #1127, opened 2023-06-05
fix: don’t include inactive segments in API request payload

dkoo:
### All Submissions:

Fixes a bug in which inactive segments are included in the payload for the segmentation API request. This doesn't resolve, but helps avoid, an issue where including too many segments in the payload results in a "Request URI too long" error response.

This is fairly unlikely to happen in a production environment, so it's an edge case.

1. Generate a lot of segments, but make most of them inactive. The easiest way to do this is to repeatedly navigate to `domain/wp-admin/admin.php?page=newspack-engagement-wizard#/reader-activation/complete` and click the button to "Enable Reader Activation". Each time you do this it will generate a new set of default segments + prompts, and deactivate the old ones. Do this at least 4 times to replicate the bug.
2. View the front-end and observe that no prompts are shown. If you open Dev Tools > Network and inspect the response from the `newspack-popups/api/campaigns/index.php` GET request, you should see the "Request URI too long" error response.
3. Check out this branch and repeat step 2. Confirm that the prompts matching your reader segment are shown as expected, and that the payload for the GET request is now much shorter.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-segmentation.php
+++ b/includes/class-newspack-popups-segmentation.php
@@ -357,9 +357,11 @@ final class Newspack_Popups_Segmentation {
 	/**
 	 * Get all configured segments.
 	 *
+	 * @param boolean $include_inactive If true, fetch both inactive and active segments. If false, only fetch active segments.
+	 *
 	 * @return array Array of segments.
 	 */
-	public static function get_segments() {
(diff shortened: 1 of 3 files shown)


miguelpeixe (approved):
Working as described! 🙌

PR #1128, opened 2023-06-06
feat: scaffold GA4 modules and implement `loaded` event

dkoo:
### All Submissions:

Scaffolds JS modules for setting up and firing GA4 custom events for prompt interactions. Also scaffolds a `loaded` event, which fires as soon as a prompt is loaded in the DOM (so no event handler needed for this trigger).

Because UA/GA3 events are being deprecated in July, this PR implements the GA4 events separately and in parallel with the existing analytics events, which can be removed after the official UA deprecation.

Closes `1204616440912018/1204751246643877`.

1. Ensure that you have Google Analytics and a GA4 property set up in Site Kit (otherwise the JS won't init).
2. Check out this branch.
3. Visit a post or page that has prompts (doesn't matter whether or not they're displayed for the `loaded` event).
4. In your JS console, confirm that each prompt logs a `prompt_interaction` event name and an accompanying payload with info about the specific prompt for the `loaded` action.

### Other information:

### DIFF
--- a/src/view/analytics/load.js
+++ b/src/view/analytics/load.js
@@ -1,0 +1,15 @@
+import { getEventPayload } from '../utils';
+
+/**
+ * Event fired as soon as a prompt is loaded in the DOM.
+ *
+ * @param {Array} prompts Array of prompts loaded in the DOM.
+ */
+export const manageLoadEvents = prompts => {
+	prompts.forEach( prompt => {
+		const payload = getEventPayload( 'loaded', 'passive', prompt );
+
+		console.log( 'event name', 'prompt_interaction' );
+		console.log( 'event payload', payload );
+	} );
+};
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -1145,6 +1145,35 @@ final class Newspack_Popups_Model {
 	}
 
 	/**
+	 * Get a description of a prompt's custom frequency settings, for analytics purposes.
+	 *
+	 * @param array $popup The popup object for the prompt.
+	 *
+	 * @return string Frequency summary.
+	 */
+	private static function get_custom_frequency_summary( $popup ) {
+		$custom_settings = [];
+
+		if ( 0 < $popup['options']['frequency_between'] ) {
+			// Translators: %d is the number of pageviews in between prompt displays, if greater than 0 (every pageview).
+			$custom_settings[] = sprintf( __( 'every %d pageviews', 'newspack-popups' ), $popup['options']['frequency_between'] + 1 );
+		}
+		if ( 0 < $popup['options']['frequency_start'] ) {
+			// Translators: %d is the pageview when the prompt starts to be displayed, if greater than 0 (first pageview).
+			$custom_settings[] = sprintf( __( 'starting on pageview %d', 'newspack-popups' ), $popup['options']['frequency_start'] + 1 );
+		}
+		if ( 0 < $popup['options']['frequency_max'] ) {
+			// Translators: %d is the max number number of displays for the prompt, if greater than 0 (no max).
+			$custom_settings[] = sprintf( __( 'max %d times', 'newspack-popups' ), $popup['options']['frequency_max'] );
+
+			// Translators: %s is the time period for when the prompt can be displayed again after the max number of displays.
+			$custom_settings[] = sprintf( __( 'resetting every %s', 'newspack-popups' ), $popup['options']['frequency_reset'] );
+		}
+
+		return implode( ',', $custom_settings );
+	}
+
+	/**
 	 * Generate markup for an inline popup.
 	 *
 	 * @param string $popup The popup object.
@@ -1163,7 +1192,7 @@ final class Newspack_Popups_Model {
 		self::remove_form_hooks( $popup );
 		do_action( 'newspack_campaigns_after_campaign_render', $popup );
 
-		$element_id           = self::get_uniqid();
+		$element_id           = self::canonize_popup_id( $popup['id'] );
 		$endpoint             = self::get_reader_endpoint();
 		$hide_border          = $popup['options']['hide_border'];
 		$large_border         = $popup['options']['large_border'];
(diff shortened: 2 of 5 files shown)


src/view/analytics/load.js
```diff
@@ -0,0 +1,15 @@
+import { getEventPayload } from '../utils';
+
+/**
+ * Event fired as soon as a prompt is loaded in the DOM.
+ *
+ * @param {Array} prompts Array of prompts loaded in the DOM.
+ */
+export const manageLoadEvents = prompts => {
+	prompts.forEach( prompt => {
+		const payload = getEventPayload( 'loaded', 'passive', prompt );
```
dkoo, line 10: @leogermani I wasn't sure what to put for the `action_type` event param. Does `passive` make sense, since this event will fire without any user interaction?

leogermani: Ideally this should be one of the values described in the docs here -> https://github.com/Automattic/newspack-plugin/tree/master/includes/data-events#prompt_interaction

leogermani:
If you agree I can wrap up that PR with some test instructions and you can continue with this one using that as a basis.

And then you can add other parameters, like you are doing with frequency, in a way it will be applied to both front-end and back-end events

src/view/analytics/load.js
dkoo: Ah, so it should be `donation`, `registration` or `newsletters_subscription` if the loaded prompt contains one of the corresponding block types, or empty if it doesn't?

leogermani:
> @leogermani That makes sense to me—thanks for pointing me in the right direction. Okay if I pull that draft into this branch and continue working on it here?

yes, sure!

And that should handle the action_type already

dkoo:
@leogermani Thanks again for the data normalizing class—I've integrated it into this PR in 550a023. Is it closer to what you had in mind?

leogermani:
Yes! I'm gonna test this asap

includes/class-newspack-popups-data-api.php
```diff
@@ -0,0 +1,176 @@
...
+
+		unset( $sanitized['interaction_data'] );
+		$sanitized = array_merge( $sanitized, $popup_params['interaction_data'] );
+
+		unset( $sanitized['prompt_blocks'] );
+		foreach ( $popup_params['prompt_blocks'] as $block ) {
+			$sanitized[ 'prompt_has_' . $block ] = 1;
+		}
+
+		// @TODO: How to handle prompts with more than one block?
```
dkoo, line 148: @leogermani If there's more than one block, what about making this a comma-delimited list? e.g. `donation, registration, newsletters_subscription`

This probably won't happen much in practice as our recommended/default prompts only contain a single type at a time.

dkoo: I'm not sure what data types GA supports for these values—if they can accept an array then that would probably be more useful.

leogermani: it only supports strings. up to 100 characters.

Yeah, we could concatenate them in alphabetical order I guess. But I wanted us to take a moment and see what are the cases this will happen to make sure this is the better way to go.

The only place I see this happening is in the new Memberships/content gating features - and we are not tracking membership purchases at all.. so I'd say that this is part of the task to add the Content gating data to the whole Data events api and ga4

leogermani:
Apparently it works well for me

image

My only comment is that I saw the 3 loaded events for 3 prompts that were never displayed in the pageivew.

I'm wondering about the value of tracking the `load` event if it is triggered every time even when the prompt will not be displayed...

leogermani (approved):

includes/class-newspack-popups-data-api.php
dkoo: True, I guess I'm not seeing the utility of the `action_type` param right now.

> I wanted us to take a moment and see what are the cases this will happen to make sure this is the better way to go.

Membership-focused prompts are a good case. But then we may want to add additional `action_type` values to cover more membership-specific purposes, right? Custom publisher-authored prompts might also be another case, if they include both a donation and newsletter signup form in the same prompt (unlikely but could happen). I'm okay leaving this as-is for now until we have some concrete use case scenarios we need to support.

dkoo:
> My only comment is that I saw the 3 loaded events for 3 prompts that were never displayed in the pageivew.

> I'm wondering about the value of tracking the load event if it is triggered every time even when the prompt will not be displayed...

The `load` event in GA3 tracks all prompts that are inserted in the DOM, regardless of whether they're displayed to the reader. This can be measured against the `seen` event (which fires only once a prompt becomes visible in the viewport) to create a sort of view rate metric per prompt. Although this does create an ambiguity: was a prompt not seen because the reader didn't match its segment, or was it not seen because they didn't scroll down far enough to see it in the post content?

I'm open to changing the behavior of this to whatever is most useful to us—the way it's implemented now simply matches what we've been tracking in GA3.

includes/class-newspack-popups-data-api.php
leogermani: > True, I guess I'm not seeing the utility of the action_type param right now.

Yeah, maybe it's redundand with the other we have.

I think we could discuss this and the `load` event subject with Katie to make a decision

dkoo:
@leogermani turns out you're right! I was incorrect about the `load` event firing as soon as the element is available in the DOM. [This JS function](https://github.com/Automattic/newspack-plugin/blob/master/includes/analytics/class-analytics-events.php#L708) is what tracks those events for GA3 (based on the `ini-load` browser event from AMP). I've duplicated this `MutationObserver` logic here so now the event will only fire when a prompt is unhidden, save as with GA3.

src/view/analytics/load.js
dkoo: Resolved by later discussions/changes

includes/class-newspack-popups-data-api.php
dkoo: Since I was wrong about this, and the `load` GA3 event does only fire for prompts that will be displayed, I think this clarifies the usefulness of the `loaded` vs. `seen` events.

PR #1132, opened 2023-06-07
feat: add tracking for click events

laurelfulford:
### All Submissions:

This PR adds tracking for links inside of prompts.

1. Ensure that you have Google Analytics and a GA4 property set up in Site Kit (otherwise the JS won't init).
2. Check out this branch.
3. Set up a few different prompts with and without links; confirm that clicking those links is getting tracked correctly in GA. 

### Other information:

### DIFF
--- a/src/view/analytics/clicked.js
+++ b/src/view/analytics/clicked.js
@@ -1,0 +1,23 @@
+import { getEventPayload, getRawId, sendEvent } from '../utils';
+
+/**
+ * Execute a callback function to send a GA event when a prompt is dismissed.
+ *
+ * @param {Array} prompts Array of prompts loaded in the DOM.
+ */
+
+export const manageClickedEvents = prompts => {
+	prompts.forEach( prompt => {
+		const anchorLinks = [
+			...prompt.querySelectorAll( '.newspack-inline-popup a, .newspack-popup__content a' ),
+		];
+		const handleEvent = () => {
+			const payload = getEventPayload( 'clicked', getRawId( prompt.getAttribute( 'id' ) ) );
+			sendEvent( payload );
+		};
+
+		anchorLinks.forEach( function ( link ) {
+			link.addEventListener( 'click', handleEvent );
+		} );
+	} );
+};
--- a/src/view/analytics/clicked.js
+++ b/src/view/analytics/clicked.js
@@ -11,8 +11,18 @@ export const manageClickedEvents = prompts => {
 		const anchorLinks = [
 			...prompt.querySelectorAll( '.newspack-inline-popup a, .newspack-popup__content a' ),
 		];
-		const handleEvent = () => {
-			const payload = getEventPayload( 'clicked', getRawId( prompt.getAttribute( 'id' ) ) );
+		const handleEvent = e => {
+			const extraParams = {};
+
+			if ( e.currentTarget?.href && '#' !== e.currentTarget?.href ) {
+				extraParams.action_value = e.currentTarget.getAttribute( 'href' );
+			}
+
+			const payload = getEventPayload(
+				'clicked',
+				getRawId( prompt.getAttribute( 'id' ) ),
+				extraParams
+			);
 			sendEvent( payload );
 		};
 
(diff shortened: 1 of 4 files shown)


dkoo:
> Non-blocking: should we have more information about the click? The link URL is not included in the payload so I don't have a reference of what was clicked.

@leogermani should we consider adding a new param to hold the URL? I was thinking about this too, but we didn't have any predetermined params that fit.

dkoo:
89d5a3c implements an `action_value` custom param that can be appended to an event's payload to send browser interaction data. This data is contextual based on the `action` value and will only be sent if there's any data to send for the event.

miguelpeixe (approved):
Working well! Good call to have a more abstract `action_value` to cover other cases.

Some non-blocking comments:

There's a chance this event never reaches GA, since a link click may leave the page before it had time to push the request. The only way around this is hijacking the click event, which I'm not a big fan of and creates very bad UX (you have to account for scroll wheel clicks opening in a new tab, etc)

Around the schema:
image

The naming `action` and `action_type` is getting semantically mixed with `action_type`. Not sure if that's on purpose, but my click action is not a registration, although it was made in a prompt that has a registration block. I find it a bit confusing, maybe we can remove the `action_type` from this prompt interaction?

dkoo:
> There's a chance this event never reaches GA, since a link click may leave the page before it had time to push the request. The only way around this is hijacking the click event, which I'm not a big fan of and creates very bad UX (you have to account for scroll wheel clicks opening in a new tab, etc)

I was worried about this, too. GA3 has a [mechanism](https://github.com/Automattic/newspack-plugin/pull/1726) to handle this scenario, but GA4 does not seem to. I couldn't find much info on why, except for [this support page](https://support.google.com/analytics/answer/9964640?hl=en#event-batching&zippy=%2Cin-this-article) which suggests that batched GA4 events are sent by default via the [sendBeacon](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon) method. This would suggest that GA4's default behavior should account for this. In my testing, I saw `clicked` events get tracked at or close to 100% of the time (noting that there's a slight delay of up to 10 seconds or so when reporting events, even in the RealTime dashboard—I think this is because of the event batching), but this is a non-scientific test. Are you able to positively confirm that some `clicked` events are not getting reported at all?

> The naming action and action_type is getting semantically mixed with action_type. Not sure if that's on purpose, but my click action is not a registration, although it was made in a prompt that has a registration block. I find it a bit confusing, maybe we can remove the action_type from this prompt interaction?

@leogermani can probably answer this better, but I believe the `action_type` is supposed to be consistent for all events related to the prompt so that we can track readers' interactions with that prompt all the way through the funnel. So if you think about the link click in the context of clicking a link from a prompt with a registration block, for example, it kind of makes sense to have the `action_type` be `registration`. The reader's actual registration is tracked by a [separate back-end event](https://github.com/Automattic/newspack-plugin/tree/master/includes/data-events/connectors/ga4#reader_registered), so this lets us view all of the events that led up to that end-of-funnel registration.

miguelpeixe:
> Are you able to positively confirm that some clicked events are not getting reported at all?

I was not able to confirm events not being sent, it was my assumption because I wasn't aware that `Navigator.sendBeacon()` was a thing 😄 . That's pretty awesome!

From the article you linked, I'm under the impression that `sendBeacon` is the standard strategy.

>  So if you think about the link click in the context of clicking a link from a prompt with a registration block, for example, it kind of makes sense to have the action_type be registration.

Got it! It feels a bit fragile because technically, prompts can have any number of blocks, but that's probably a separate discussion.

dkoo:
> Got it! It feels a bit fragile because technically, prompts can have any number of blocks, but that's probably a separate discussion.

Very true! The `action_type` for the prompt will only be `registration` or `donation` if the prompt contains one of those blocks, so hopefully that will accurately describe the intent of the prompt. We'll have to see how creative publishers get with building their own prompt content. 😅

leogermani:
Maybe it should be called 'prompt_action` and `prompt_action_type`...

PR #1134, opened 2023-06-07
feat: GA4 `form_submission` events for prompts

dkoo:
### All Submissions:

Implements `form_submission` GA4 events when a form inside prompt content is submitted.

Also implements an `action_value` for all donation-related events, which lists the suggested donations for that form (if manually configured in the block, from the block's attributes, if not, from the sitewide donation settings).

#### `form_submission` events

1. On a site connected to a GA4 property via Site Kit, publish prompts containing registration, newsletter signup, and donation forms.
2. On the front-end, submit each form. Confirm in your GA dashboard > Reports > Realtime that a `form_submission` event with relevant event params gets fired upon form submission for each form.

#### `clicked` event `action_value`

1. Publish a prompt containing an link pointing to a URL anywhere in the content.
2. View the prompt on the front-end and click on the prompt.
3. In your GA dashboard, confirm that the corresponding `clicked` event has an `action_value` param with the link's URL. Note that we won't pass `#` values if the link triggers HTML, but we will pass `#id` values if it links to another element on the same page.

#### Suggested donation `action_value`

1. Publish several prompts containing Donation blocks. Make sure you have at least the following variations (in all cases make sure the suggested donation values for manually configured blocks differ from default Donation settings in Reader Revenue):
  - Not manually configured, standard layout
  - Manually configured, standard layout, tiered
  - Manually configured, standard layout, not tiered
  - Manually configured with at least one disabled frequency (different from the default Donation settings in Reader Revenue, if any are disabled there)
  - Not manually configured, tiers layout
  - Manually configured, tiers layout
4. View the prompts on the front-end. In your GA dashboard, confirm that any event fired by a prompt containing a donation block has an `action_value` param with a value showing the donate block layout plus the suggested donations for every enabled tier, like so:
  - `standard layout | once: 9,20,90,20 | month: 7,15,30,15 | year: 84,180,360,180`
  - `tiers layout | month: 20,40,60 | year: 84,180,360`
5. Confirm that each `action_value` matches the suggested donation amounts displayed in each block variation. You can also more easily debug this by changing the `gtag` to `console.log` at [this line](https://github.com/Automattic/newspack-popups/blob/epic/ga4/src/view/utils/index.js#L233).

Notes on the format of the summary:
  - If a block is set to "manually configured", the values come from block attributes. If not, they come from Reader Revenue > Donation settings.
  - Disabled frequencies won't be shown.
  - If standard layout + not tiered, the Donate block only shows the "other" input field, so each frequency should show only a single suggested value.
  - If tiered layout, the Donate block doesn't allow for one-time frequency or "other" inputs, so these should be omitted from the summary.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-data-api.php
+++ b/includes/class-newspack-popups-data-api.php
@@ -98,6 +98,42 @@ final class Newspack_Popups_Data_Api {
 		foreach ( $watched_blocks as $key => $block_name ) {
 			if ( \has_block( $block_name, $popup['content'] ) ) {
 				$data['prompt_blocks'][] = $key;
+
+				// Get the suggested donation values for the donation block.
+				if ( 'donation' === $key ) {
+					$prompt_blocks = \parse_blocks( $popup['content'] );
+					$donate_block  = array_shift(
+						array_filter(
+							$prompt_blocks,
+							function( $block ) use ( $block_name ) {
+								return $block_name === $block['blockName'];
+							}
+						)
+					);
+
+					if ( $donate_block && method_exists( '\Newspack\Donations', 'get_donation_settings' ) ) {
+						$is_manual         = $donate_block['attrs']['manual'] ?? false;
+						$default_settings  = \Newspack\Donations::get_donation_settings();
+						$donation_settings = $is_manual ? \wp_parse_args( $donate_block['attrs'], $default_settings ) : $default_settings;
+						$is_tiered         = $donation_settings['tiered'] ?? false;
+						$suggested_amounts = $donation_settings['amounts'] ?? [];
+						$disabled_tiers    = $donation_settings['disabledFrequencies'] ?? [];
+						$suggested_summary = [];
+
+						foreach ( $suggested_amounts as $frequency => $amounts ) {
+							if ( empty( $disabled_tiers[ $frequency ] ) ) {
+								if ( ! $is_tiered ) {
+									$amounts = [ end( $amounts ) ];
+								}
+								$suggested_summary[] = $frequency . ': ' . implode( ',', $amounts );
+							}
+						}
+
+						if ( ! empty( $suggested_summary ) ) {
+							$data['action_value'] = implode( ' | ', $suggested_summary );
+						}
+					}
+				}
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


miguelpeixe (changes requested):
Working as described! The `clicked` event is done in #1132, so it doesn't need to be in this PR.

A few questions:

For donation prompts, the `form_submission` is triggered when choosing a donation amount and clicking to checkout, because it's a form that generates the checkout page. This is a few steps before the donation intent is actually made. Even though we shouldn't infer a donation from a front-end donation intent, firing this early can be misleading when analyzing.

Unless I'm missing something, with `standard layout | once: 9,20,90,20 | month: 7,15,30,15 | year: 84,180,360,180` we're creating a new format that could be replaced with plain JSON, which is safer to encode/decode:
```json
{ "layout": "standard", "once": [9,20,90,20], "month": [7,15,30,15], "year": [84,180,360,180] }
```

dkoo:
`{"layout":"standard","once":[9,20,90,20],"month":[7,15,30,15],"year":[84,180,360,180]}` (with whitespace stripped) is only 87 characters, and this may be the longest this value gets, so maybe JSON is the way to go, though? I love the idea of sending JSON in `action_value` for values that need to be parsed like this.

dkoo:
@miguelpeixe f0b9ed3 encodes the `action_value` string as JSON. It gets us close to the 100 character limit, but I think for purposes of this particular value, it shouldn't ever exceed 100 since the longest possible string is likely something like:

```json
"{"layout":"standard","once":[30,40,100,100],"month":[10,25,50,100],"year":[250,500,1000,1000]}"
```

which is 97 characters.

miguelpeixe:
We could also go for:

```json
"{"layout":"standard",[[30,100,200,100],[10,25,100,100],[250,500,1000,1000]]}"
```

dkoo:
But then we lose the info on which frequencies are enabled/disabled. Maybe this?

```json
{"layout":"standard","o":[9,20,90,20],"m":[7,15,30,15],"y":[84,180,360,180]}
```

miguelpeixe:
Could be represented by the same index but `null` 🤔 

```json
{"layout":"standard",[null,[10,25,100,100],[250,500,1000,1000]]}
```

miguelpeixe:
Just an update if you go for the matrix option, that's no valid json. Should be something like this:

```json
{"layout":"standard","config":[null,[10,25,100,100],[250,500,1000,1000]]}
```

dkoo:
Ok, I went with the shorter key names in 0c363a0. Thanks for the discussion!

miguelpeixe:
Perfect! Thank you!

miguelpeixe (approved):
Changes look good and test well! 🎉  🙌

PR #1137, opened 2023-06-09
fix: ensure inline posts are inserted into main post content only

dkoo:
### All Submissions:

There's a weakness in the [`the_content`](https://developer.wordpress.org/reference/hooks/the_content/) filter hook in that it doesn't tell you what post object the filtered content is coming from. For this reason, the callback to insert inline prompts in article content sometimes gets called on the wrong content, as was discovered in https://github.com/Automattic/newspack-theme/pull/2127.

This PR attempts to mitigate that situation by comparing the `post_content` property on the post in the global `$wp_query` and comparing it to the `$content` string being filtered, before executing the prompt insertion logic. Since this filter callback is being called with priority `1`, before most or all other content filters would be run to transform this content string, the strings should match. I haven't done an exhaustive test of this, though.

There shouldn't be any major changes in functionality. Inline prompts should be inserted in article post content only.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -461,6 +461,8 @@ final class Newspack_Popups_Inserter {
 			// Don't inject inline popups on paywalled posts.
 			// It doesn't make sense with a paywall message and also causes an infinite loop.
 			|| self::is_memberships_restricted()
+			// At filter priority 1, $content should be the same as the unfiltered post_content. This guards against inserting in other content such as featured image captions/descriptions.
+			|| get_post()->post_content !== $content
 		) {
 			return $content;
 		}


miguelpeixe (changes requested):
I'm not sure this is the safest strategy to mitigate the problem. If anything filters the content at priority 1 we will not have prompts rendering.

WDYT of making the filter run only once so any other run inside the content or afterward should be skipped. Does that make sense? Or does an inner run execute before the post run? 🤔

dkoo:
> I'm not sure this is the safest strategy to mitigate the problem. If anything filters the content at priority 1 we will not have prompts rendering.

> WDYT of making the filter run only once so any other run inside the content or afterward should be skipped. Does that make sense? Or does an inner run execute before the post run? 🤔

We already run the filter only once, but unfortunately in the bug found in https://github.com/Automattic/newspack-theme/pull/2127 the featured image caption filter runs before the post content. 😞

miguelpeixe (approved):
Ah, I didn't realize that :(

Well, it tests well!

dkoo:
I always wish the `the_content` filter hook would provide a post ID as a second arg, but alas.

If this causes future issues we could try a less-precise comparison, such as stripping all tags and comparing the first 50 characters, or something similar...

PR #1155, opened 2023-06-23
feat: segmentation criteria system

miguelpeixe:
### All Submissions:

pemhSX-jU-p2
pemhSX-km-p2

Implements a criteria system so it can be registered with a simple API and common matching functions integrated with the [reader data library](https://github.com/Automattic/newspack-plugin/pull/2451).

#### Registration

The criteria are first registered with `Newspack_Popups_Criteria::register_criteria( $id, $config );`. See [its definition](https://github.com/Automattic/newspack-popups/blob/6ee1154aafbc7e4539cb7b1aec953ef54e7566f2/includes/class-newspack-popups-criteria.php#L117-L149), and [`src/criteria/default/index.php`](https://github.com/Automattic/newspack-popups/blob/bfa2a75cb431c604c99ad33f7759d9e6c61b94e7/src/criteria/default/index.php) and [`tests/test-criteria.php`](https://github.com/Automattic/newspack-popups/blob/1470b5411fac7f24a9a25ae38e29264fb59f0d23/tests/test-criteria.php) for usage examples.

Once registered, their ID, matching function, and matching attribute are sent to the front-end through `wp_localize_script()`. In the front-end they can optionally be enhanced with `setMatchingAttribute( id, callback )` ([example](https://github.com/Automattic/newspack-popups/blob/1792f19e398eb2b9e3ca0545360f1d0236019910/src/criteria/default/articles-read.js)) or `setMatchingFunction( id, callback )` ([example](https://github.com/Automattic/newspack-popups/blob/7f53a9ba8d80e64bc402fcf09d6f6ae0c7b9a85f/src/criteria/default/donation.js)) to have a dynamic matching strategy.

If the matching function or attribute are strings, the criteria matching will look for [common matching functions](https://github.com/Automattic/newspack-popups/blob/611765da8a46cc21cb8d63dcba9864e4b3253e30/src/criteria/matching-functions.js) or store items by key respectively, in order to fetch the criteria value and match against segment configuration.

#### Matching

Matching occurs entirely in the front-end through `criteria.matches( config )`, e.g.:

```js
import { getCriteria } from './criteria/utils';
const segmentConfig = { value: 'foo' };
const criteria = getCriteria( criteriaId );
if ( criteria.matches( segmentConfig ) ) {
	console.log( 'Reader matches' );
}
```

Written tests should pass.

#### Registering 

1. Check out this branch and https://github.com/Automattic/newspack-plugin/pull/2537 to see how criteria can render in the UI
2. Add the following at the end of `src/criteria/default/index.php` so we can see new criteria in action:
```php
\Newspack_Popups_Criteria::register_criteria( 'favorite_color' );
```
3. In the dashboard, visit Campaigns -> Segments -> Add New Segment
4. Confirm you see the new input under the "Reader Activity" section

image

5. Change the config in the code:

```php
$config = [
	'name'        => __( 'Favorite Color', 'newspack-popups' ),
	'description' => __( 'Favorite color of reader.', 'newspack-popups' ),
	'category'    => 'reader_engagement',
	'options'     => [
		[
			'label' => __( 'Any color', 'newspack-popups' ),
			'value' => '',
		],
		[
			'label' => __( 'Red', 'newspack-popups' ),
			'value' => 'red',
		],
		[
			'label' => __( 'Green', 'newspack-popups' ),
			'value' => 'green',
		],
		[
			'label' => __( 'Blue', 'newspack-popups' ),
			'value' => 'blue',
		],
	],
];
\Newspack_Popups_Criteria::register_criteria( 'favorite_color', $config );
```

6. Refresh the page and confirm the criteria moved to the "Reader Engagement" category and render options:

image

#### Criteria matching 

1. Visit the site in a fresh session and open the browser logs
2. Confirm you see 2 logged entries with `{ segmentId: '1', config: Array(), matched: false }`
3. Edit `src/criteria/segments-example.js` to match against the newly created criteria
4. Add the following in line 23:
```js
	3: [
		{
			criteria_id: 'favorite_color',
			value: 'blue',
		},
	],
```
5. Refresh the page and confirm the new log entry with `matched: false`
6. In DevTools set the store value to match the segment: `newspackReaderActivation.store.set( 'favorite_color', 'blue' );`
7. Refresh the page and confirm it now logs `matched: true` for this segment

The more complete integration with segments is a WIP in this PR: https://github.com/Automattic/newspack-popups/pull/1159

### Other information:


src/criteria/default/referrer.js
```diff
@@ -0,0 +1,15 @@
+import { setMatchingAttribute } from '../utils';
+
+const matchingAttribute = ( { store } ) => {
+	const value = document.referrer
+		? ( new URL( document.referrer ).hostname.replace( 'www.', '' ) || '' ).toLowerCase()
+		: '';
+	// Persist the referrer in the store.
+	if ( value ) {
+		store.set( 'referrer', value );
```
leogermani, line 9: I think we should avoid writing to the store as much as possible in the matching functions.

Isn't `referrer` something the store could provide by default, just like it does with `article_view` and other info? It looks like something generic enough to live there.

includes/class-newspack-popups-criteria.php
```diff
@@ -0,0 +1,237 @@
...
+	/**
+	 * Get registered criteria config to be used in the front-end.
+	 *
+	 * @return array
+	 */
+	public static function get_criteria_config() {
+		$config = [];
+		foreach ( self::$registered_criteria as $id => $criteria ) {
+			$config[ $id ] = [
+				'matchingFunction'  => $criteria['matching_function'],
```
leogermani, line 208: doesn't this throw a warning for the criteria without these attributes?

miguelpeixe: Not by design. The default matching attribute is the criteria ID and the default matching function is `default`.

src/criteria/default/referrer.js
miguelpeixe: Why avoid? It's embedded support to a powerful API. More advanced criteria will very likely benefit from this.

includes/class-newspack-popups-criteria.php
leogermani: I see it now!

src/criteria/default/referrer.js
leogermani: For me it's a matter of separation of concerns.

On one side, the criteria checks if the current reader matches a specific situation. In order to do that, it READS info from the store and whatever other information needed. But it only reads.

On the other side the store is updated based on user interaction and other events that happen. This happens somewhere else.

I feel that if we start to write data to the store as we are checking for criteria things will get really confusing, with one criteria potentially stepping in the toes of the others... And it will also become hard to track what's making updates to the store.

Also, our Reader Data Library could be as complete as possible, offering global and valuable information about the current reader to all its consumers. Referrers feels like one of those things that could be part of the store by default...

(If you ask me, I still think that the whole criteria system should belong to the Data Library as well)

miguelpeixe: Mind that this is the matching attribute, not the matching function.

I agree that the matching function is only reading data to determine a bool result.

The matching attribute is generating the data, which can either be a string, which runs `store.get( string )`, or a function, which will create the data. The data creation can be just stored data processing with a dynamic result, but that might not always be the case, and it's in the context of data creation to also allow storing it.

miguelpeixe: > (If you ask me, I still think that the whole criteria system should belong to the Data Library as well)

I don't know what that means. Criteria have requirements to integrate with our UI, and segments and be performant when calculating matching. It leverages the store library as much as possible while still being able to dynamically process data when it needs.

miguelpeixe: I moved the referrer store item logic to this PR: https://github.com/Automattic/newspack-plugin/pull/2541

src/criteria/default/index.php
```diff
@@ -0,0 +1,121 @@
...
+ * Register default segmentation criteria.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack_Popups;
+
+defined( 'ABSPATH' ) || exit;
+
+$criteria = [
```
dkoo, line 12: Nit: in the future, when we want to consider using this system register additional criteria from other plugins of ours, wouldn't we want to apply a filter to this `$criteria` object so other plugins could add their own criteria by simply appending new keys and config to the object, and/or remove default criteria by unsetting default keys? I see the `newspack_popups_registered_criteria` filter [here](https://github.com/Automattic/newspack-popups/pull/1155/files#diff-8da51b859e12deff15c7b71debad6c7e6859326edc7d11c8c56630d57215b816R98), but being able to hook into the config before the criteria are registered seems more useful to me.

dkoo (approved):
This is a beautiful API that describes what segments are and how readers match their criteria in a more abstract way. It's working exactly as described, and my comments/questions are non-blocking, so approving. Also appreciate how much less verbose the unit tests are in both PHP and JS, now.

Noting to myself that segments-example.js will be merged into the epic branch and will need to replaced in a separate PR before the epic is ready for deployment.

src/criteria/utils.js
```diff
@@ -0,0 +1,146 @@
...
+	// Check if the criteria matches the segment config.
+	criteria._matched = {}; // Cache results.
+	criteria.matches = segmentConfig => {
+		const configString = JSON.stringify( segmentConfig );
+		if ( criteria._matched[ configString ] !== undefined ) {
+			return criteria._matched[ configString ];
+		}
+		const ras = window.newspackReaderActivation;
+		if ( ! ras ) {
+			console.warn( 'Reader activation script not loaded.' ); // eslint-disable-line no-console
```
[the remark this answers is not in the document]
miguelpeixe: It shouldn't error but if the matching attribute is a string, which fetches a store item from the reader data library, it'll always return `null`:

https://github.com/Automattic/newspack-popups/blob/a31365c6ab0153984d4af7ffe7f55190adbbb8dd/src/criteria/utils.js#L70-L72

Theoretically, you could have criteria independent of the data library, but the goal right now is to fail gracefully without throwing errors.

src/criteria/default/favorite-categories.js
```diff
@@ -0,0 +1,19 @@
...
+		}
+		return c;
+	}, [] );
+	const counts = categories.reduce( ( number, category ) => {
+		number[ category ] = ( number[ category ] || 0 ) + 1;
+		return number;
+	}, {} );
+	const countsArray = Object.entries( counts );
+	countsArray.sort( ( a, b ) => b[ 1 ] - a[ 1 ] );
+	/* TODO: Decide how to rank categories. */
```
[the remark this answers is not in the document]
miguelpeixe: We could also merge as is and have a separate PR for this.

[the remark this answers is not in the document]
dkoo: Agreed that we should probably handle this outside of this PR. I do think we should better define what "favorite categories" means (even if it differs from the current behavior) and make that logic clearer in the code.

includes/class-newspack-popups-criteria.php
```diff
@@ -0,0 +1,150 @@
...
+		wp_localize_script(
+			self::SCRIPT_HANDLE,
+			'newspackPopupsCriteria',
+			[
+				'config' => self::get_criteria_config(),
+			]
+		);
+		wp_enqueue_script(
+			'newspack-popups-default-criteria',
+			plugins_url( '../dist/defaultCriteria.js', __FILE__ ),
```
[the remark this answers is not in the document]
dkoo: Got it—the ease of the dependency management trumps the micro-optimization we'd get from bundling the JS files and having to handle the async data through code.

src/criteria/default/articles-read.js
```diff
@@ -0,0 +1,29 @@
...
+} );
+
+setMatchingAttribute( 'articles_read_in_session', ras => {
+	const views = ras.getUniqueActivitiesBy( 'article_view', 'post_id' );
+	if ( ! views.length ) {
+		return 0;
+	}
+	// Sort by descending timestamp.
+	views.sort( ( a, b ) => b.timestamp - a.timestamp );
+	// Bail if the most recent view is older than 30 days.
```
dkoo, line 15: Wouldn't we want to bail if the most recent view is older than 30 _minutes_? If that's the case, then any views couldn't be considered the current session, right?

dkoo (changes requested):
One question about the new logic for articles read in session. Otherwise the logic looks 👍  to me!

src/criteria/default/articles-read.js
miguelpeixe: Whoops, that was a typo in the inline comment. The code is 30 minutes. Fixed in 3bee99371212cca106d3555778087aa4569829ab

dkoo (approved):
I should have done the math! Thanks for fixing the comment. :)

PR #1159, opened 2023-07-03
feat: integrate criteria with segments

miguelpeixe:
### All Submissions:

Changes the Segments schema and store an array of criteria.

Also drafts the migration method, changing the segment structure on the fly.

1. Checkout and build https://github.com/Automattic/newspack-plugin/pull/2537
2. Create and edit segments. Confirm everything still work as usual
3. Confirm tests are passing

### Other information:


leogermani:
@miguelpeixe Only thing I want to do here is move the migration code and make it work like the other migrations (and also add a failsafe for that cache bug we had). I think it's better to loop through all segments and run the migration once for all.

But we can do this in a separate PR.

I think we can move on with this PR

dkoo (approved):
I ran into an issue when editing an existing segment. Updating the select dropdown values doesn't always update the `isDirty` state, so the "Save" button remains disabled until you edit another field. Editing other types of fields seems to work as expected, and the issue is intermittent with dropdowns. I'm approving this PR because I think the issue is coming from https://github.com/Automattic/newspack-plugin/pull/2537, not this one.

miguelpeixe:
Thanks, @dkoo! Copied your comment to https://github.com/Automattic/newspack-plugin/pull/2537 so it doesn't get lost while I wrap up the draft for review.

PR #1166, opened 2023-07-13
chore: remove Lightweight API and client ID handling

dkoo:
### All Submissions:

Removes the lightweight PHP API and associated unit tests. Also removes other major features that will no longer be needed after the rearchitecture:

- AMP client ID handling (and all functionality related to setting and getting the AMP client ID via cookies): This is no longer needed in this plugin, since all reader data and segmentation features will be tied to either browser session via `localStorage`, or to a WP user
- AMP Analytics markup and GA3/UA analytics tracking via AMP Analytics: [GA3 has been deprecated as of July 1](https://support.google.com/analytics/answer/11583528?hl=en), so there's no need to retain these methods, which send events only via GA3
- AMP Access config and handling: This will be replaced by a pure JS solution in a future PR
- Setting of reader donation/subscription/registration events: This is now handled by the Reader Data library in the main plugin
- The data pruning cron job that cleans up temp reader events from the legacy tables: No longer needed since we won't be writing to these tables anymore.

Note that this PR does extensive cleanup that will intentionally result in broken functionality. The following features are broken and will need to be reimplemented in separate PRs using the criteria system introduced in #1155:

- Segmentation: In previews and on the front-end, all prompts will be visible at all times. The prompts will need to be hooked up with the new criteria system.
- Previewing: The new criteria system does not implement per-prompt or per-segment previewing. This will need to be  reintroduced using the new system. I left the view-as parameter setting and parsing intact in case we want to reuse those in the new system.
- Segment reach estimates: This feature in the Campaigns dashboard UI relies on reader data existing in the SQL database, and will no longer work. A separate PR in `newspack-plugin` will need to deprecate this feature in that plugin.

This PR also doesn't do some further cleanup which will be done after we have a means to migrate legacy reader data to the new system. Notably, we're not cleaning up any reader data in the custom tables (although we're no longer writing any new data to those tables). There may also be methods related to AMP cookies and client IDs across our other repos that we can delete, too, but I'm choosing to leave the client ID handling in other repos intact until we're sure they're no longer needed in those plugins.

Note also that some changes are required in other Newspack plugins to remove references to methods that will no longer exist after this changeset. https://github.com/Automattic/newspack-plugin/pull/2558 and https://github.com/Automattic/newspack-blocks/pull/1498 are works-in-progress which we can add more changes to if necessary for the rearchitecture.

Closes `1204751246381647/1205010980866823`.

1. Confirm that the plugin functions in the admin, and that you can still create/edit segments and prompts as expected.
2. Confirm that the plugin functions on the front-end. Note that segmentation will be broken as noted above, but that you should see the new criteria-based segment matching tests being logged in the JS console as implemented in #1155.
3. Confirm there are no new PHP fatals or warnings, and no new JS errors resulting from the deleted files.

### Other information:


includes/class-newspack-popups-newsletters.php
```diff
@@ -153,60 +102,47 @@ private static function fetch_reader_data_from_esp( $email_address ) {
...
-							'amount' => $donation_amount,
-						],
-					];
-				}
-			}
-		}
 
-		if ( ! empty( $events_to_add ) ) {
-			$api->save_reader_events( $client_id, $events_to_add );
+			// TODO: set donation status in the Reader Data store.
```
miguelpeixe, line 145: That's been done in https://github.com/Automattic/newspack-plugin/pull/2539

includes/class-newspack-popups-newsletters.php
```diff
@@ -153,60 +102,47 @@ private static function fetch_reader_data_from_esp( $email_address ) {
...
+			$events_to_add[] = [
+				'type'    => 'subscription',
+				'context' => $email_address,
+				'value'   => [
+					'esp'   => $newsletter_provider->service,
+					'lists' => $subscribed_lists,
+				],
+			];
+
+			// TODO: set subscription status in the Reader Data store.
```
miguelpeixe, line 122: Also resolved in https://github.com/Automattic/newspack-plugin/pull/2539

miguelpeixe (changes requested):
What a beautiful changeset ❤️ 

I think we can also remove `includes/class-newspack-popups-donations.php` and `includes/class-newspack-popups-newsletters.php`, wdyt?

includes/class-newspack-popups-newsletters.php
dkoo: Ah, you're right—this file only checks subscription/donation status upon user login, but that's already handled by https://github.com/Automattic/newspack-plugin/pull/2539. 🙌

includes/class-newspack-popups-newsletters.php
dkoo: You're right, thanks for pointing it out!

dkoo: Removed this file in 6c97c16

includes/class-newspack-popups-newsletters.php
dkoo: Removed this file in 6c97c16

dkoo:
> I think we can also remove `includes/class-newspack-popups-donations.php` and `includes/class-newspack-popups-newsletters.php`, wdyt?

Removed those files in 6c97c16. You're right that they're no longer needed since the new reader data library API handles the newsletter/donation status updates on reader login.

dkoo:
Realized there were still a couple of failing PHPunit tests that will no longer be needed (since the decision on which segments match the reader will no longer be executed in PHP). Removed in c08b6bc.

miguelpeixe (approved):
Nice! LGTM!

PR #1168, opened 2023-07-14
Refactor/segments relationship

leogermani:
### All Submissions:

Refactors how the Segments and Prompts relationship is stored.

Removes the `selected_segment_id` option and uses the native post - taxonomy relationship

Make sure you have https://github.com/Automattic/newspack-plugin/pull/2583  and https://github.com/Automattic/newspack-blocks/pull/1498 checked out and built.

We need to test the whole Popups and segmentation feature, from all its angles.

Here is a non-exhaustive list of things to test:

* On a site with many existing prompts and segments, confirm the data migration works fine and they all still work the same way

* on a fresh site, test creating segments and prompts
* test assigning prompts to Segments via the default taxonomy UI in the prompt Editor
* try editing segments, enabling and disabling them, both from the segment edit page and from the toggle in the segments list
* try editing prompts via the Settings Modal in the Campaigns wizard. Try changing the prompts segments
* Confirm that the Conflict warning for uncategorized prompts in the same segment still work as expected
* test the export CLI command
* test the import CLI command
* on a fresh site, test the RAS wizard and confirm the default prompts and segments are created well

Confirm that the `data-segments` html attributes of the popups are correctly filled

Confirm that the Event names for GA tracking gets the correct segment names for the prompt (Search for "Newspack Announcement" in the page source)

* Important: This is basically all I tested. Please think of things I didn't test. We really need to stress it.

### Other information:


miguelpeixe (changes requested):
The relationship appears to be working well.

A few issues I'm running into:

1. Switching to this branch and https://github.com/Automattic/newspack-plugin/pull/2583 is throwing a warning, which I'm not getting on the epic branches:
```
PHP Warning:  Undefined array key "type" in wp-includes/rest-api.php on line 3075
```
2. The segment selector includes deactivated selectors, which can create a bad UI. I'd expect deactivated segments to not be available:
image
3. Every term selector (even outside of prompts editing) got a weird spacing/height on 6.3 RC4:
image

The issue is different on 6.2. There's a weird spacing but only on the segments selector:

image

leogermani:
Thanks for testing @miguelpeixe 

> Switching to this branch and https://github.com/Automattic/newspack-plugin/pull/2583 is throwing a warning, which I'm not getting on the epic branches:

Fixed in https://github.com/Automattic/newspack-popups/pull/1168/commits/44bdab9a5423a1a1262745241af6c6f566d3f6f5. Actually related to a bug in core -> https://core.trac.wordpress.org/ticket/56494, so I did a workaround.

> The segment selector includes deactivated selectors, which can create a bad UI. I'd expect deactivated segments to not be available:

Good call. I addressed it in https://github.com/Automattic/newspack-popups/pull/1168/commits/54df2995316664e25bf8d7e4e54842d5e6f6658a . `is_disabled` should be refactored and live as a meta of its own, but I think there's enough refactor for one PR.

> Every term selector (even outside of prompts editing) got a weird spacing/height on 6.3 RC4:

I can't reproduce this. Running 6.3 on Firefox and Chrome and things look normal. Can you double check or see if you find any other info?

miguelpeixe (approved):
Thank you for revising!

I can no longer reproduce the style issues I was having before.

It's working great now! 🎉

PR #1169, opened 2023-07-14
feat: segmentation match logic

dkoo:
### All Submissions:

Implements a JS-based segmentation API using the criteria system. Removes AMP markup and config JSON. Replaces AMP polyfill JS with vanilla DOM-based JS.

Not working right now (will be resolved in future PRs):

- Previewing single prompts or segments from the Campaigns dashboard (interestingly, previewing a single prompt from the editor works)
- Donor landing page setting
- Mailchimp donor merge fields setting
- Non-interactive mode setting (this may be removed since the segmentation API is no longer server-based)

Closes `1204189396255901/1204261432690236`.

1. Create a variety of segments and both overlay/inline prompts targeting different criteria. Ensure the prompts have a variety of settings, especially when it comes to frequency, and trigger (for overlays) settings.
2. View on the front-end and confirm that all prompts display as expected according to frequency settings, scroll/delay triggers, and segment criteria.
3. Also test Above Header, Custom Placement, Manual Only placements, and shortcodes in post/page content and in widgets.

### Other information:


includes/class-newspack-popups-inserter.php
```diff
@@ -62,13 +69,46 @@ public static function popups_for_post() {
...
+				'wp_footer',
+				function() {
+					?>
+					<script>
+						var newspackPopupsSegments = <?php echo \wp_json_encode( self::$segments ); ?>;
+					</script>
+					<?php
+				}
+			);
+		}
```
miguelpeixe, line 109: Any reason to favor inline JS instead of `wp_localize_script()`?

src/criteria/segments-example.js
```diff
@@ -1,10 +1,10 @@
-/* globals newspackPopupsSegmentsExample */
```
miguelpeixe, line 1: WDYT of moving this bit to a `src/prompts` and `src/segments` directory? Should be helpful for debugging

src/criteria/utils.js
```diff
@@ -144,3 +144,18 @@ export function setMatchingFunction( id, matchingFunction ) {
...
+ * @param {any}    data Data to log.
+ */
+export function debug( key, data ) {
+	if ( ! newspack_popups_view.debug ) {
+		return;
+	}
+
+	window.newspack_popups_debug = window.newspack_popups_debug || {};
+	window.newspack_popups_debug[ key ] = data;
+}
```
miguelpeixe, line 161: This seems like something that could live in a different utility js, like `src/prompts`. wdyt?

miguelpeixe (commented):
Thank you for putting this together so quickly! I left some comments on file structure, but I am not sure how we'd want all of that to work in the final code.

includes/class-newspack-popups-inserter.php
dkoo: Not at all—when I started I wasn't sure if we needed something other than JSON config for this, but I don't think we will.

src/criteria/utils.js
dkoo: Great idea. I restructured the files to something closer to this in e79cba4.

src/criteria/segments-example.js
dkoo: How about this? e79cba4

includes/class-newspack-popups-inserter.php
dkoo: Changed to a localized value in e79cba4.

dkoo:
@miguelpeixe I'm open to suggestions on how to make this more unit-testable. Right now it's very DOM-heavy.

dkoo:
I suppose we could output prompt settings config as JSON instead of in `data-` attributes. 🤔

dkoo:
> I suppose we could output prompt settings config as JSON instead of in data- attributes. 🤔

On third thought, bundling the required config with each prompt's markup is actually the easier way to handle things, to avoid having to run multiple queries for the markup to be inserted plus the configs for each. That's a situation that requires us to use two separate queries in the lightweight API: one for prompt markup and one for AMP Access config. Because prompts can be inserted programmatically via a `the_content` filter, or via custom/manual placements, which may also appear anywhere on the page, the config query had to query ALL published prompts in order to ensure that all configs would be available for segmentation. By bundling the config with each prompt's markup, we no longer need to worry about having to fetch configs separately from the prompt markup.

src/view/utils/segments.js
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Check the reader's activity against a given prompt's assigned segments.
+ *
+ * @param {HTMLElement} prompt          HTML element of the prompt being checked.
+ * @param {string}      matchingSegment ID of the reader's highest-priority matching segment.
+ * @param {Object}      ras             Reader Data Library object.
+ * @return {boolean} True if the prompt should be displayed, false if not.
+ */
+export const shouldPromptBeDisplayed = ( prompt, matchingSegment, ras ) => {
```
miguelpeixe, line 66: Answering your question on unit testing, I think this is a great candidate. It'll require mocking the `ras` object but otherwise, it's a core function that connects a lot of pieces.

miguelpeixe: This entire file, actually 😄

dkoo: Tests added in 355972c

src/view/analytics/seen.js
```diff
@@ -46,7 +8,8 @@ const getObserver = handleEvent => {
 export const manageSeenEvents = prompts => {
 	prompts.forEach( prompt => {
 		const handleEvent = () => {
-			const payload = getEventPayload( 'seen', getRawId( prompt.getAttribute( 'id' ) ) );
+			const promptId = getRawId( prompt.getAttribute( 'id' ) );
+			const payload = getEventPayload( 'seen', promptId );
 			sendEvent( payload );
 		};
```
miguelpeixe, line 14: Non-blocking nit: dispatching the `prompt_seen` activity triggers an event that can get picked up through `ras.on( 'activity', cb )`. It might be better to use that API for the same goal instead of running another intersection observer.

miguelpeixe (changes requested):
This is looking great!

I added some comments below for minor stuff, but I'm having issues making the overlay prompts work with Perfmatters enabled. I'm not sure what's happening but I think it's related to the plugin's delay script strategy in combination with the observer that handles the unhide(). All the logic is applied correctly but doesn't render/unhide.

src/view/analytics/seen.js
dkoo: The logic is backwards, though—the intersection observer is what triggers the `prompt_seen` activity, not the other way around. We only want to dispatch `prompt_seen` once the prompt is visible in the viewport for a certain period of time.

src/view/utils/segments.js
```diff
@@ -0,0 +1,108 @@
...
+	// By frequency.
+	// eslint-disable-next-line @wordpress/no-unused-vars-before-return
+	const [ start, between, max, reset ] = prompt.getAttribute( 'data-frequency' ).split( ',' );
+	const pageviews = ras.store.get( 'pageviews' );
+	if ( pageviews[ reset ] ) {
+		const views = pageviews[ reset ].count || 0;
+
+		// If reader hasn't viewed enough articles yet.
```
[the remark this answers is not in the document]
dkoo: You're right—fixed in 7e1a92b!

src/view/analytics/seen.js
miguelpeixe: Is it not like that in the code? My suggestion is to make the analytics read from `prompt_seen` event trigger ([which is already behind an intersection observer](https://github.com/Automattic/newspack-popups/blob/7e1a92b0dd26850f92a695da12f6aa540c798aa2/src/view/utils/index.js#L68-L78)) instead of having another intersection observer.

dkoo: Oh! I see what you mean. You're right, we should trigger the analytics event based on the activity dispatch.

src/view/utils/segments.js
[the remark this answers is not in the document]
miguelpeixe: Learned a new verb today 😄

dkoo:
> I'm having issues making the overlay prompts work with Perfmatters enabled. I'm not sure what's happening but I think it's related to the plugin's delay script strategy in combination with the observer that handles the unhide(). All the logic is applied correctly but doesn't render/unhide.

@miguelpeixe c9fab16 fixes this for me—does it for you, too?

src/view/analytics/seen.js
dkoo: 173e03d refactors the `seen` GA4 event so it gets triggered based on `ras.on( 'activity' )` instead of its own intersection observer.

miguelpeixe:
> c9fab1669fab5723b9136255df4fed57de7a9cc8 fixes this for me—does it for you, too?

Not really, it still doesn't render for me and this causes a JS error when Perfmatters is not active. The default criteria setup must load after their registration in `src/criteria/index.js`, Perfmatters async strategy might be allowing for that to work, but it's not failproof.

miguelpeixe:
Wait, resetting my Perfmatters defaults fixed it 🤦 

Not sure what was breaking it, but I suspect it could be the CSS tree-shaking. Since every JS piece was running correctly and the rendering was the only thing missing...

src/view/segmentation.js
```diff
@@ -0,0 +1,82 @@
...
+				closeButton.addEventListener( 'click', closeOverlay );
+			} );
+
+			// Check segmentation.
+			const shouldDisplay = shouldPromptBeDisplayed(
+				prompt,
+				matchingSegment,
+				ras,
+				overlayDisplayed
+			);
```
miguelpeixe, line 45: I'm getting this returning `true` for every overlay prompt on the homepage, which doesn't render all the prompts on the page but dispatches the `prompt_seen` activity. Here's the localStorage on a fresh session with RAS defaults:

image

It's unclear to me why the `overlayDisplayed` is being sent as the `override` argument in the function. This causes every`shouldPromptBeDisplayed` to be forced as `true`.

dkoo: Whoops, that was a late-day logic mistake. 🤦 This should fix it: dd3fbd3

dkoo:
> Still, this doesn't resolve the rendering on Perfmatters. The matching logic is working as expected, as it was before, and reaches the `unhide()` function [here](https://github.com/Automattic/newspack-popups/blob/f567e1afca4208a3215d56cc2868d6f99ff5e170/src/view/segmentation.js#L54-L59). For some reason it just doesn't render 😕 

Wait, I'm actually seeing the same thing now when I deactivate Perfmatters. Will test with #1175...

dkoo:
Ah, based on https://github.com/Automattic/newspack-popups/pull/1169#issuecomment-1649966827 I had thought it may not have been an issue. I can confirm that merging #1175 does fix the issue with and without Perfmatters for me.

miguelpeixe:
Reverting c9fab1669fab5723b9136255df4fed57de7a9cc8 and merging #1175 should do the trick!

miguelpeixe:
Can you paste the errors?

dkoo:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'store')
    at Object.eval [as matchingFunction] (donation.js:5:20)
    at criteria.matches (VM2054 utils.js:110:48)
    at match (segments.js:42:21)
    at getBestPrioritySegment (segments.js:67:9)
    at handleSegmentation (segmentation.js:19:87)
    at eval (index.js:20:70)
    at domReady (index.js:76:17)
    at eval (index.js:17:51)
    at ./src/view/index.js (view.js?ver=1690299117:119:1)
    at __webpack_require__ (view.js?ver=1690299117:323:41)
```

miguelpeixe:
Looks like RAS is undefined there, right?

dkoo:
Maybe we can emit a custom event when the criteria are finished registering, and only run the segmentation logic once that event fires?

miguelpeixe:
I'll debug a bit more to get the root cause.

dkoo:
> Looks like RAS is undefined there, right?

Ah yes, it's the `window.newspackReaderActivation` object that's missing there.

miguelpeixe:
Found the trace. [This](https://github.com/Automattic/newspack-popups/blob/19e30ef792e8652b0309a8c2dd845b372bee1337/src/view/index.js#L15) is called without ensuring that RAS has loaded, which call [this](https://github.com/Automattic/newspack-popups/blob/19e30ef792e8652b0309a8c2dd845b372bee1337/src/view/segmentation.js#L18) then [this](https://github.com/Automattic/newspack-popups/blob/19e30ef792e8652b0309a8c2dd845b372bee1337/src/view/utils/segments.js#L41), and breaks.

dkoo:
I had thought that pushing to the `window.newspackRAS` array [here](https://github.com/Automattic/newspack-popups/pull/1169/files#diff-5c4afc658ed53af3e9430b59f7c9221a0092d428f676b9183c06a293e36906b8R22) would ensure that the callback only runs after the RAS library has loaded?

miguelpeixe:
Yes, but there's a `getBestPrioritySegment( segments )` a few lines up outside the callback push. This function executes criteria matching, which will fail without the store.

dkoo:
🤦 Sorry about that! Does 7eafd06 fix it?

dkoo:
Not seeing any race conditions with or without Perfmatters now! 🥳

miguelpeixe (approved):
All is working as expected! 🎉

dkoo:
Thank you for your help in resolving the async and Perfmatters-related issues!

PR #1178, opened 2023-08-07
fix: previews

dkoo:
### All Submissions:

Fixes two preview-related issues resulting from the rearchitected segmentation API:

- When loading a single-prompt preview from the Newspack admin UI, manual and custom placement prompts are replaced by the prompt being previewed.
- When previewing a segment from the Newspack Campaigns wizard, mocking the segment matching was broken by the front-end implementation of segmentation matching.

#### Manual and custom placement prompts in single-prompt previews

1. On `epic/rearchitecture`, publish a manual-only or custom placement prompt and place it via a block in your most recent post (this will be used by preview windows created by this plugin).
2. Preview a different inline prompt and observe that the the manually placed block renders the wrong prompt (the prompt being previewed instead of the one specified in the block's attributes).
3. Check out this branch and repeat step 2. This time confirm that the manually placed block renders the expected prompt.

#### Previewing as a segment

1. Check out https://github.com/Automattic/newspack-plugin/pull/2607.
2. On this branch, visit **Newspack > Campaigns** and test previewing segments.
3. Confirm that only the prompts within the previewed segment are displayed in the preview iframe.
4. Confirm that when navigating the site within each new preview window, the prompts within the previewed segments continue to display and respect frequency settings (so if a prompt is set to display every other pageview, it will do so within the preview window).

#### Previewing a single prompt

1. With this branch and https://github.com/Automattic/newspack-plugin/pull/2607 checked out, visit **Newspack > Campaigns** and test previewing a singular prompts from both this screen and from the post editor while editing a single prompt.
2. Confirm that the specific prompt (and only that prompt) is always shown in the preview window.

### Other information:


dkoo:
> Note that a consequence of moving reader-related data to `localStorage`, reader data is now shared between preview iframes and other non-preview windows in the same browser session. To mitigate confusion, prompts are always displayed on every pageview in preview iframes, otherwise frequency-based prompts won't always display when opening a new preview iframe.

@miguelpeixe there may be a way to isolate reader data set and read in preview sessions using the `newspack_reader_data_store_prefix` filter when a `view_as` spec is appended to the URL, but I couldn't get this to work because the store `get` methods always seem to get reader data associated with the default store prefix. I'm thinking this behavior shouldn't be a blocker to shipping the rearchitecture since most publishers are previewing segments on a single page to check prompt design.

miguelpeixe (commented):
> I couldn't get this to work because the store get methods always seem to get reader data associated with the default store prefix

I'm not sure I undestand. The front-end `store.get()` method is not using the filtered `newspack_reader_data_store_prefix`? If that's the case, it's a bug. The store prefix should be localized with the value configured on the backend:

https://github.com/Automattic/newspack-plugin/blob/fbf2ab542a525bfaf92661033ecafca67f13baa5/assets/reader-activation/store.js#L17

https://github.com/Automattic/newspack-plugin/blob/fbf2ab542a525bfaf92661033ecafca67f13baa5/includes/reader-activation/class-reader-data.php#L42-L68

I like the idea of mocking a reader data store for campaign preview. The browser's `sessionStorage` implements the same [WebStorage API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API).

If we are able to programmatically have the reader data store load from `sessionStorage`, it should be a nice disposable place to work with. Currently, it's configured [here](https://github.com/Automattic/newspack-plugin/blob/fbf2ab542a525bfaf92661033ecafca67f13baa5/assets/reader-activation/store.js#L18), which we can change to something like this:

```js
storage: newspack_reader_data?.use_session ? window.sessionStorage : window.localStorage,
```

dkoo:
> I'm not sure I undestand. The front-end store.get() method is not using the filtered newspack_reader_data_store_prefix? If that's the case, it's a bug. The store prefix should be localized with the value configured on the backend

I tried this locally and it wasn't working as expected in the iframed window, but it's possible I wasn't doing it correctly. I was simply filtering the `newspack_reader_data_store_prefix` value, which caused the `set` methods to set with the filtered prefix, but the `get` methods did not.

That said, I really like your suggestion of using `sessionStorage` as a temporary store for this purpose and will revisit this using session storage.

dkoo:
@miguelpeixe was in the process of reviewing this, but @leogermani if you have time today it could use your eyes as well. Otherwise it can wait til Miguel is back

dkoo:
> > I'm not sure I undestand. The front-end store.get() method is not using the filtered newspack_reader_data_store_prefix? If that's the case, it's a bug. The store prefix should be localized with the value configured on the backend
> 
> I tried this locally and it wasn't working as expected in the iframed window, but it's possible I wasn't doing it correctly. I was simply filtering the `newspack_reader_data_store_prefix` value, which caused the `set` methods to set with the filtered prefix, but the `get` methods did not.
> 
> That said, I really like your suggestion of using `sessionStorage` as a temporary store for this purpose and will revisit this using session storage.

e3ef33d and https://github.com/Automattic/newspack-plugin/pull/2607 implement the temporary sessions via `sessionStorage`.

327eb0f fixes single-prompt previewing from the Campaigns wizard.

includes/class-newspack-popups.php
```diff
@@ -1211,5 +1212,33 @@ public static function disable_prompts_for_protected_pages() {
...
+
+		// Tell the reader store to use sessionStorage instead of localStorage, and tie the reader prefix to the session.
+		\add_filter( 'newspack_reader_data_store_is_temp_session', '__return_true' );
+		\add_filter(
+			'newspack_reader_data_store_prefix',
+			function() use ( $session_id ) {
+				return sprintf( 'np_temp_session_%d_', $session_id );
+			}
+		);
+	}
```
miguelpeixe, line 1242: I wonder if this bit should live under `newspack-plugin` as a core feature of the reader data store. We can use [`is_preview()`](https://developer.wordpress.org/reference/functions/is_preview/) and [`is_customize_preview()`](https://developer.wordpress.org/reference/functions/is_customize_preview/) to always use temporary sessions.

Separate PR, though. We don't have to do this now.

src/view/segmentation.js
```diff
@@ -27,22 +29,23 @@ export const handleSegmentation = prompts => {
 		prompts.forEach( prompt => {
 			const promptId = prompt.getAttribute( 'id' );
 			const isOverlay = prompt.classList.contains( 'newspack-lightbox' );
+			const override =
+				// eslint-disable-next-line no-nested-ternary
+				getRawId( promptId ) === getPreviewedPromptId() // If previewing a single prompt, it should always be displayed.
+					? true
+					: isOverlay && overlayDisplayed // If an overlay and another overlay has already been displayed, it should not be displaeyd.
+					? false
+					: null; // Default behavior lets frequency/segmentation determine whether it should be dipslayed.
```
miguelpeixe, line 38: This logic is getting a bit more difficult to read. Perhaps it can move to a separate function with guard clauses?

miguelpeixe (approved):
Awesome! Tests well 🎉

includes/class-newspack-popups.php
dkoo: I can see that argument! My reasoning for where it is:

- Using the two filters in the main plugin (`newspack_reader_data_store_is_temp_session` and optionally `newspack_reader_data_store_prefix` to create isolated reader data sets) temp sessions can be created for any purpose from any other plugin.
- This particular temp session is created for the specific purpose of previewing Campaigns prompts and segments, so it makes sense in my mind to live in Campaigns.

miguelpeixe (approved):
The latest changes test well! 🙌

includes/class-newspack-popups.php
dkoo: I'm definitely open to revisiting this decision in a future PR, though

PR #1179, opened 2023-08-08
fix: fallback if no newspack plugin

dkoo:
### All Submissions:

Segmentation and frequency-based features in the rearchitected plugin rely on APIs from the main Newspack Plugin. This PR ensures that if those APIs are not available, prompts will still be displayed, albeit without any regard to segmentation or frequency options.

Closes `1204189396255901/1204850211981984`.

1. Check out this branch and disable the main plugin: `wp plugin deactivate newspack-plugin`
2. In WP admin, confirm that the Prompts menu item is available and that it lists all prompts in a standard CPT UI.
3. Edit a prompt and confirm that the "Frequency" and "Segments" sidebar panels don't appear, as those options won't work without the main Newspack plugin.
4. View the front-end and confirm that all published prompts are displayed (although only one overlay prompt will be displayed).

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -658,22 +658,28 @@ final class Newspack_Popups_Inserter {
 				true
 			);
 
-			$segments = Newspack_Popups_Segmentation::get_segments( false );
-
-			// Gather segments for all prompts to be displayed.
-			foreach ( $segments as $segment ) {
-				if ( ! empty( $segment ) && ! empty( $segment['criteria'] ) && ! isset( self::$segments[ $segment['id'] ] ) ) {
-					self::$segments[ $segment['id'] ] = [
-						'criteria' => $segment['criteria'],
-						'priority' => $segment['priority'],
-					];
-				}
-			}
-
 			$script_data = [
-				'debug'    => defined( 'WP_DEBUG' ) && WP_DEBUG,
-				'segments' => self::$segments,
+				'debug' => defined( 'WP_DEBUG' ) && WP_DEBUG,
 			];
+
+			if ( class_exists( '\Newspack\Reader_Data' ) ) {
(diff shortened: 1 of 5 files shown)


includes/class-newspack-popups-inserter.php
```diff
@@ -658,23 +658,29 @@ public static function enqueue_scripts() {
...
+							'criteria' => $segment['criteria'],
+							'priority' => $segment['priority'],
+						];
+					}
+				}
+
+				$script_data['segments'] = self::$segments;
+
+			} else {
+				$script_data['segmentation_disabled'] = true;
```
leogermani, line 681: nit: can we avoid the negative name here? It can become very confusing (as we saw yesterday)

dkoo: Good call—d4b9a25 removes it, as there's really no need for a separate variable at all. More simply, if there are no `segments` to handle, then there's no need to use the reader data library at all.

dkoo: Whoops, ef8ac67 fixes a bug in the prior commit which assumed that `newspack_popups_view.segments` would be an array and therefore have a `length` property, when in fact it's a keyed object.

leogermani:
hey @dkoo , I fixed the conflict after my PR was merged to the epic branch.

Also, now that Segments use the core taxonomy UI in the Editor, we needed a different approach to hide it. 

It tests well for me but can you double-check it?

PS - I tried to not even register the taxonomy if segmentation is disabled but then many things that rely on it started to break, so I figured it would be simpler to just hide it

dkoo (commented):
@leogermani Changes look good to me, but I can't currently test. If they're working for you with and without Newspack Plugin active, that's good enough for me. Feel free to approve or request another PR from @miguelpeixe or @laurelfulford for a second set of eyes.

PR #1188, opened 2023-08-15
Fix: auto select segment when coming from the Campaigns wizard

leogermani:
### All Submissions:

Automatically selects the segment when coming from the Campaigns wizard "Add new prompt" button in each segment group:

image

1. Go to the Campaigns wizard
2. Click on the "Add new prompt" button on any segment
3. Select any placement
4. In the editor, confirm that the segment you clicked on is automatically selected in the "Segments" sidebar control
5. Save and confirm the value is stored
6. Repeat the process and make changes to the selected segments, also hide and show the Segments form, and confirm your changes are kept

NOTE: Toggling the form visibility was forcing a component re-render and thus the `useEffect` call, resetting the selected segments. The workaround I found was to push the browser history and remove the `segment` query var from the URL once it's read for the first time. Maybe there's a better solution for this?

### Other information:

### DIFF
--- a/src/editor/index.js
+++ b/src/editor/index.js
@@ -9,11 +9,13 @@
  */
 import { __ } from '@wordpress/i18n';
 import { compose } from '@wordpress/compose';
-import { withSelect, withDispatch, useSelect } from '@wordpress/data';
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Works for me

PR #1211, opened 2023-08-29
fix: avoid SQL error if legacy table doesn't exist

dkoo:
### All Submissions:

Avoids running a SQL query on a non-existent table when attempting to migrate reader data from legacy tables.

1. On `master`, set up a brand new site with Campaigns.
2. Register a new reader account in a new session.
3. In another new session, log into the reader account.
4. Observe an error in your debug.log:

```
WordPress database error Table 'wordpress.wp_newspack_campaigns_reader_events' doesn't exist for query 
					SELECT DISTINCT client_id
					FROM wp_newspack_campaigns_reader_events
					WHERE type = 'user_account' AND context = '2'
				 made by require('wp-blog-header.php'), wp, WP->main, do_action_ref_array('wp'), WP_Hook->do_action, WP_Hook->apply_filters, Newspack_Segments_Migration::migrate_user_data
```
5. Check out this branch and repeat step 3 in a new session. Confirm that the error doesn't occur and that the `migrate_user_data` function exits silently without attempting to migrate any data.

### Other information:

### DIFF
--- a/includes/class-newspack-segments-migration.php
+++ b/includes/class-newspack-segments-migration.php
@@ -271,6 +271,11 @@ final class Newspack_Segments_Migration {
 
 		global $wpdb;
 
+		// No need to migrate if the old tables don't exist.
+		if ( empty( $wpdb->get_var( $wpdb->prepare( 'SHOW TABLES LIKE %s', $wpdb->prefix . 'newspack_campaigns_reader_events' ) ) ) ) { // phpcs:ignore WordPress.DB.DirectDatabaseQuery.NoCaching, WordPress.DB.DirectDatabaseQuery.DirectQuery
+			return;
+		}
+
 		// Fetch the user's client ids.
 		$client_ids = $wpdb->get_results( // phpcs:ignore WordPress.DB.DirectDatabaseQuery.NoCaching, WordPress.DB.DirectDatabaseQuery.DirectQuery
 			$wpdb->prepare(


includes/class-newspack-segments-migration.php
```diff
@@ -262,7 +282,8 @@ public static function migrate_user_data() {
 		if (
 			! is_user_logged_in() ||
 			get_user_meta( get_current_user_id(), 'newspack_popups_reader_data_migrated', true ) ||
-			! class_exists( 'Newspack\Reader_Data' )
+			! class_exists( 'Newspack\Reader_Data' ) ||
+			! self::$legacy_table_exists
```
miguelpeixe, line 286:
```suggestion
			! self::check_for_legacy_table()
```

To avoid running the `check_for_legacy_table()` method (and a db query) on every page load, the method could cache `self::$legacy_table_exists` and return if the value has been set.

dkoo: 5f2f118 caches the result in a new options row—did you have another method in mind?

miguelpeixe: My original idea was just caching in memory, but that also works 🙂

miguelpeixe: Why run the method on `init` instead of calling the method when needed inside the migration method?

dkoo: > Why run the method on `init` instead of calling the method when needed inside the migration method?

That was an oversight, sorry! 181aff0 removes callback on `init` and only runs the method when needed (when a reader registers or logs in).

> My original idea was just caching in memory, but that also works 🙂

I tried this, but the cached var didn't persist across new pageview requests, so it ended up running the DB query on every request anyway. The other thing I was going to try was a transient, but an option makes the most sense to me because it's unlikely that the existence of the legacy tables will change (until we drop them).

miguelpeixe (approved):

PR #1228, opened 2023-10-02
fix: deduplication strategy for Homepage Posts block

miguelpeixe:
### All Submissions:

https://github.com/Automattic/newspack-blocks/pull/1543 proposes a new strategy for opting out of the deduplication logic in the Homepage Posts block. This requires a change in how it's implemented for campaigns.

This PR implements the necessary changes to the logic initially implemented in #630 so it continues to work as expected.

1. Checkout https://github.com/Automattic/newspack-blocks/pull/1543 and this branch
2. Create a page and a campaign prompt with matching Homepage Posts blocks (so they display the same content)
3. Make sure the prompt is displayed as an overlay and can be rendered on the created page
4. Visit the page and confirm the expected duplicated posts rendered on the page and the prompt

### Other information:

### DIFF
--- a/src/editor/index.js
+++ b/src/editor/index.js
@@ -20,7 +20,7 @@ import { useEffect } from '@wordpress/element';
 /**
  * Internal dependencies
  */
-import { optionsFieldsSelector } from './utils';
+import { optionsFieldsSelector, isOverlayPlacement } from './utils';
 import Sidebar from './Sidebar';
 import StylesSidebar from './StylesSidebar';
 import FrequencySidebar from './FrequencySidebar';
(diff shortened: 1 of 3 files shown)


dkoo (changes requested):
Works as described, but shouldn't the behavior respect the setting of "Use deduplication logic" for the Homepage Posts block(s) inside the prompt instead of forcing non-deduping inside all overlays?

miguelpeixe:
The behavior implemented in #630 and preserved by this PR is to disable deduplication for blocks inside overlay prompts. This exists because if a prompt is generated it'll "rob" posts from the page (see #126).

For overlay prompts to have their own deduplication possible, the deduplication logic should support a separate layer for the overlay. This is far more complicated and is not covered in this PR.

src/editor/index.js
```diff
@@ -145,6 +145,23 @@ registerPlugin( 'newspack-popups-editor', {
 	icon: null,
 } );
 
+// Hide Newspack's Homepage Posts block deduplication toggle when the popup is an overlay.
+registerPlugin( 'newspack-popups-disable-newspack-blocks-deduplication', {
+	render: function HideDeduplicationToggle() {
+		const { isOverlay } = useSelect( select => {
+			const { getEditedPostAttribute } = select( 'core/editor' );
+			return {
+				isOverlay: getEditedPostAttribute( 'meta' )?.placement === 'center',
```
dkoo, line 154: We should check for any of the valid overlay placements as defined here: https://github.com/Automattic/newspack-popups/blob/master/includes/class-newspack-popups-model.php#L19

dkoo (changes requested):
Works beautifully! Just one change required to make this change apply to any overlay placement, not just centered ones (although these are most common for sure).

src/editor/index.js
miguelpeixe: Whoops! Thank you for catching that, fixed in 849bad6215f3002c5cfbc4613d66229f6f06216d

dkoo (approved):

PR #1244, opened 2023-11-21
Fix/ras presets duplication

adekbadek:
### All Submissions:

Adds means of removing duplicate segments (if no active prompts are assigned to them). This does not handle duplicate prompts, since we haven't observed this issue on live sites. Probably duplicate segments have been created accidentally during RAS beta phase. 

See 1200133283036252-as-1205733927102145

1. Switch `newspack-plugin` to `fix/duplicate-segments` branch (https://github.com/Automattic/newspack-plugin/pull/2766)
1. Run `$ wp newspack-popups import --ras-defaults` at least twice, so you end up with some duplicate prompts & segments
2. Observe "DUPLICATE" badges displayed in Campaigns wizard -> Segments tab
3. Run `$ wp eval "\Newspack_Segments_Model::remove_duplicates();"`, observe that the duplicate segments have been removed

### Other information:

### DIFF
--- a/includes/class-newspack-segments-model.php
+++ b/includes/class-newspack-segments-model.php
@@ -415,6 +415,47 @@ final class Newspack_Segments_Model {
 	}
 
 	/**
+	 * Get the ids of prompts assigned to a segment.
+	 *
+	 * @param int $segment_id The segment ID.
+	 */
+	private static function get_segments_prompts_ids( $segment_id ) {
+		return get_posts(
+			[
+				'post_type'      => Newspack_Popups::NEWSPACK_POPUPS_CPT,
+				'fields'         => 'ids',
+				'posts_per_page' => -1,
+				'tax_query'      => [ // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_tax_query
+					[
+						'taxonomy' => self::TAX_SLUG,
+						'field'    => 'term_id',
+						'terms'    => (int) $segment_id,
+					],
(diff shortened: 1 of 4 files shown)


includes/class-newspack-segments-model.php
```diff
@@ -381,9 +381,80 @@ function( $segment ) {
...
+
 		return $segments;
 	}
 
+	/**
+	 * Get the ids of prompts assigned to a segment.
+	 *
+	 * @param int $segment_id The segment ID.
+	 */
+	private static function get_segments_prompts_ids( $segment_id ) {
```
leogermani, line 422: nit: should it be get_segment (singular) _prompts_ids ?

adekbadek: It was meant as "segment's prompts ids", but yeah, that's not super clear. Updated in c9e36e5674cf6d3648836337c699031070f95deb

includes/class-newspack-segments-model.php
```diff
@@ -381,9 +381,80 @@ function( $segment ) {
 			);
 		}
 
+		// Compute a criteria hash, so duplicate criteria sets can be found.
+		$segments = array_map(
+			function( $segment ) {
+				$segment['criteria_hash'] = md5( wp_json_encode( $segment['criteria'] ) );
```
leogermani, line 387: Nice

leogermani (approved):

PR #1258, opened 2024-01-08
fix: dont change global post before header

leogermani:
### All Submissions:

When we loop through prompts before header, we use a standard WP Loop, and use `wp_reset_postdata` to reset the global `$post` object afterwards.

This works in most cases, but we just found a situation where this doesn't work and ends up messing up the global object.

When you are in an archive page that has no posts, `wp_reset_postdata()` will not reset post data, because there is no post. See this check -> https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/class-wp-query.php#L4898

I modified the loop to use a simple foreach and to not touch the global post variable.

1. On `master`, create a new user and make sure there are no posts assigned to them
2. Using Newspack theme, visit this user archive
3. Confirm the user name displayed in the page title is not the user you are seeing the archive for
4. Checkout this branch and reload
5. Confirm the user now matches the user you are looking at
6. Confirm that popups are still working as expected!!!

### Other information:

---
  - https://app.asana.com/0/0/1206059823028916

### DIFF
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -319,9 +319,7 @@ final class Newspack_Popups_Model {
 	protected static function retrieve_popups_with_query( WP_Query $query, $include_taxonomies = false ) {
 		$popups = [];
 		if ( $query->have_posts() ) {
-			while ( $query->have_posts() ) {
-				$query->the_post();
-				$popup_post = get_post( get_the_ID() );
+			foreach ( $query->posts as $popup_post ) {
 				if ( Newspack_Popups::NEWSPACK_POPUPS_CPT === $popup_post->post_type ) {
 					$popup = self::create_popup_object(
 						$popup_post,
@@ -334,7 +332,6 @@ final class Newspack_Popups_Model {
 					}
 				}
 			}
-			wp_reset_postdata();
 		}
 		return $popups;
 	}
--- a/includes/class-newspack-popups.php
+++ b/includes/class-newspack-popups.php
@@ -888,6 +888,9 @@ final class Newspack_Popups {
 	 * @param WP_Post $post The prompt post object.
 	 */
 	public static function is_account_related_post( $post ) {
+		if ( ! $post instanceof WP_Post ) {
+			return false;
+		}
 		return has_shortcode( $post->post_content, 'woocommerce_my_account' );
 	}
 


dkoo (changes requested):
On this branch, I see a warning when I edit a post:

```
PHP Deprecated:  str_contains(): Passing null to parameter #1 ($haystack) of type string is deprecated in /var/www/html/wp-includes/shortcodes.php on line 150
```

I think we can avoid it if we add a check for the results of `get_post()` right after [this line](https://github.com/Automattic/newspack-popups/blob/master/includes/class-newspack-popups-inserter.php#L860), e.g.:

```php
		$post = get_post();
		if ( empty( $post ) ) {
			return false;
		}

		$post_type = $post->post_type;

		// Prompts should be hidden on account related pages (e.g. password reset page).
		if ( Newspack_Popups::is_account_related_post( $post ) ) {
			return false;
		}
```

dkoo:
> Confirm the user name displayed in the page title is not the user you are seeing the archive for

I also couldn't replicate this with a new user. The title shown matches the user.

leogermani:
> I also couldn't replicate this with a new user. The title shown matches the user.

Did you have published prompts on your site? I didn't mention it but you must have some global prompts that will be displayed in the authors archive page

leogermani:
> I think we can avoid it if we add a check for the results of get_post() right after [this line](https://github.com/Automattic/newspack-popups/blob/master/includes/class-newspack-popups-inserter.php#L860), e.g.:

@dkoo on https://github.com/Automattic/newspack-popups/pull/1258/commits/37748cdb83f1334abcd42ff4f9c41e0efc6149cc I did more cirurgical fix for the warnings.

I was afraid that returning false up in the `should_display` method could break things. That method is looking a bit weird.. be cause it assumes there is a global post - or even that we are on a singular page, but it runs on every request.. so it ends up doing the checks using the first post of the archive page, which is not super accurate.

Also, the reason you only see the Warnings on this branch is related to what this branch is fixing. A global post only exists because the Prompts loop sets a global post with `the_post()` and then `reset_post_data` fails to clean it up. So it is actually checking against a prompt post...

I'm not sure what else is needed for you to reproduce it. I could easily reproduce it on my local env using the Newspack theme. An alternative is check on the site mentioned in the Asana task

dkoo (approved):
Confirmed that 37748cd fixes the errors when editing a post. I still can't replicate in my local site, but I'm not seeing any ill effects from these changes.

PR #1259, opened 2024-01-10
feat: use custom capability_type for the CPT

adekbadek:
### All Submissions:

Registers the prompt CPT with a custom `capability_type`, in order to allow granular control over access to this CPT.

1. There should be no functional changes, verify that prompts are editable and visible on the front-end.

### Other information:

### DIFF
--- a/includes/class-newspack-popups.php
+++ b/includes/class-newspack-popups.php
@@ -91,6 +91,7 @@ final class Newspack_Popups {
 		add_action( 'pre_delete_term', [ __CLASS__, 'prevent_default_category_on_term_delete' ], 10, 2 );
 		add_filter( 'show_admin_bar', [ __CLASS__, 'show_admin_bar' ], 10, 2 ); // phpcs:ignore WordPressVIPMinimum.UserExperience.AdminBarRemoval.RemovalDetected
 		add_filter( 'newspack_blocks_should_deduplicate', [ __CLASS__, 'newspack_blocks_should_deduplicate' ], 10, 2 );
+		add_filter( 'cme_plugin_capabilities', [ __CLASS__, 'cme_plugin_capabilities' ] );
 
 		include_once __DIR__ . '/class-newspack-popups-logger.php';
 		include_once __DIR__ . '/class-newspack-popups-model.php';
@@ -1283,6 +1284,16 @@ final class Newspack_Popups {
 				return sprintf( 'np_temp_session_%d_', $session_id );
 			}
 		);
+	}
+
+	/**
+	 * Filter the capabilities list in the capability-manager-enhanced plugin.
+	 *
+	 * @param array $capabilities_list The list of capabilities tabs.
+	 */
+	public static function cme_plugin_capabilities( $capabilities_list ) {
+		$capabilities_list['Newspack Campaigns'] = self::get_capabilities_list();
+		return $capabilities_list;
 	}
 }
 Newspack_Popups::instance();
--- a/includes/class-newspack-popups.php
+++ b/includes/class-newspack-popups.php
@@ -73,13 +73,13 @@ final class Newspack_Popups {
 	 * Constructor.
 	 */
 	public function __construct() {
-
 		// Segmentation requires the main Newspack plugin.
 		self::$segmentation_enabled = class_exists( '\Newspack\Reader_Data' );
 
 		add_action( 'cli_init', [ __CLASS__, 'register_cli_commands' ] );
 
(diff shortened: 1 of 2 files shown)


includes/class-newspack-popups.php
```diff
@@ -156,17 +156,55 @@ public static function register_cpt() {
...
+			}
+		}
+	}
+
+	/**
+	 * Get capabilities necessary to manage this CPT.
+	 *
+	 * See https://developer.wordpress.org/reference/functions/register_post_type/#capabilities.
+	 */
+	public static function get_capabilities_list() {
```
leogermani, line 189: nit, but you don't need to redo this. You can:

```
		$args = [
			'map_meta_cap'    => true,
			'capability_type' => self::NEWSPACK_POPUPS_CPT,
			'capabilities'    => []
		];
		
		return get_post_type_capabilities( (object) $args );
```

Actually `$args` are the same arguments passed to register the post type. So if you had that array available, you could use the same array in both places.

includes/class-newspack-popups.php
```diff
@@ -73,13 +73,13 @@ public static function instance() {
...
 	 */
 	public function __construct() {
-
 		// Segmentation requires the main Newspack plugin.
 		self::$segmentation_enabled = class_exists( '\Newspack\Reader_Data' );
 
 		add_action( 'cli_init', [ __CLASS__, 'register_cli_commands' ] );
 
 		add_action( 'init', [ __CLASS__, 'register_cpt' ] );
+		add_action( 'admin_init', [ __CLASS__, 'add_caps' ] );
```
leogermani, line 82: This only needs to run once, the roles capabilities are stored in the DB. We don't want to run this on every execution

includes/class-newspack-popups.php
adekbadek: Thanks! Added in a55dc11c468c03b23dfa012ee4d3ba2144a7c646

includes/class-newspack-popups.php
adekbadek: Fixed in 39ba1b9765909ced595a0c675adbcb602f00eaa3. The recommended way to set up caps on plugin activation, but the plugin is already active on sites.

adekbadek:
Pinging @leogermani – I'd appreciate a re-review since the corresponding changes in `newspack-plugin` will soon be released 🙏

laurelfulford (changes requested):
Hi @adekbadek! I ran a review on the functionality while waiting for Leo's approval for the code changes -- with this PR and the associated Newspack Plugin one, I:

* Made a role with the capabilities to access `newspack wizard newspack-popups-wizard` plus the post and admin settings you suggested, and all the Newspack Campaign CPTs. Assigned this role, I could successfully edit/add campaigns and prompts, and make changes to the Segments and Settings tabs.
* I made a role that had no access to campaigns, and when assigned this role successfully... didn't see anything to do with campaigns 😆 
* I tested these PR changes with PublishPress Capabilities disabled and repeated the steps in the first bullet point, and it seemed fine.

The only issue I ran into is when I tried mixing up the campaign CPT capabilities -- like when I removed the ability to delete the popup CPTs, it removed the 'Move to trash' button from the prompt editor, but it didn't remove the 'Delete' option for the prompt from the main campaigns screen, or stop me from deleting one:

Settings:
image

Individual prompt post: 
image

Options under campaigns wizard: 
image

Something similar happens when you restrict editing access (you can still pick edit from the dropdown from the wizard), but then you get a WordPress error, so I think it's technically OK:

image

Just let me know if you have any questions about this!

adekbadek:
> like when I removed the ability to delete the popup CPTs, it removed the 'Move to trash' button from the prompt editor, but it didn't remove the 'Delete' option for the prompt from the main campaigns screen, or stop me from deleting one:

This is actually by design. The idea is that the wizard capabilities should override the CPT capabilities – you get access to the whole wizard, not just parts of it. 

I don't think that such fine-grained permissions are justifiable at this point. I imagine a publisher granting an employee capability to manage prompts, period. A scenario where an employee is allowed to e.g. edit prompts but not publish or delete them is _imaginable_, but I don't think it'd be common. We can revisit if we get feedback requesting such fine-grained control over the wizard.

leogermani (approved):

laurelfulford (approved):
> We can revisit if we get feedback requesting such fine-grained control in the wizard.

That all makes sense -- thanks @adekbadek!

laurelfulford:
Going to merge this so I can spin up fresh alphas! 🙂

PR #1260, opened 2024-01-10
fix: prompts and segments schema

leogermani:
### All Submissions:

Fixes the schema for prompts and segments.

Recently `criteria_hash` was added to segments, but we didn't update the schema, and thus we broke the importer/exporter tool. In prompts `duplicate_of` was in the wrong place.

1. on `master` in a site with many promtps and segments recently creted, try exporting it via CLI and confirm you see some error in segments: `wp newspack-popups export`
2. Checkout this branch and confirm export works
3. Also confirm import works

### Other information:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
     "composer/installers": "^2.0",
     "brainmaestro/composer-git-hooks": "^2.8",
     "dealerdirect/phpcodesniffer-composer-installer": "^0.7.0",
-    "yoast/phpunit-polyfills": "^1.0",
+    "yoast/phpunit-polyfills": "^1.1",
     "phpunit/phpunit": "^7.0 || ^9.5",
     "dms/phpunit-arraysubset-asserts": "^0.5.0"
   },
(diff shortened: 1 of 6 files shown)


composer.json
```diff
@@ -9,7 +9,7 @@
     "composer/installers": "^2.0",
     "brainmaestro/composer-git-hooks": "^2.8",
     "dealerdirect/phpcodesniffer-composer-installer": "^0.7.0",
-    "yoast/phpunit-polyfills": "^1.0",
+    "yoast/phpunit-polyfills": "^1.1",
```
dkoo, line 12: Are this and the related changes to `composer.lock` intended changes?

dkoo (approved):
Works! Just one non-blocking question.

composer.json
leogermani: Yes, otherwise phpunit won't run

dkoo: Thanks! Just wanted to make sure because I've sometimes pushed unintended composer updates by accident.

PR #1262, opened 2024-01-16
feat: fetch up to 1000 prompts

dkoo:
### All Submissions:

Bumps the max number of prompts fetched by the API from 100 to 1000.

1. Test with a site that already has more than 100 prompts, or generate more than 100.
2. On `release`, visit Newspack > Campaigns and observe that only 100 prompts are shown.
3. Check out this branch, refresh, and confirm that up to 1000 prompts are now shown.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-importer.php
+++ b/includes/class-newspack-popups-importer.php
@@ -241,7 +241,12 @@ class Newspack_Popups_Importer {
 				'post_status'  => $prompt['status'],
 				'post_type'    => Newspack_Popups::NEWSPACK_POPUPS_CPT,
 			];
-			$new_post  = wp_insert_post( $post_data );
+			if ( $prompt_slug ) {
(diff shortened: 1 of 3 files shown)


adekbadek (approved):

dkoo:
Not sure why the PHP test is failing here. 🤔

dkoo:
Turns out there was a bit of a race condition in the importer test. When creating the posts for each imported prompt, they were getting the same `post_date`, which caused the default sort order (by date) from `WP_Query` to be unpredictable. The test would pass sometimes but not always. Adding a simple `sleep( 1 )` after creating each imported prompt seems to fix it.

I also noticed that the `slug` attribute for prompt imports wasn't being used, so I added it and a test for it as well.

adekbadek (approved):

PR #1277, opened 2024-02-13
fix: ga4 events

adekbadek:
### All Submissions:

Fixes above-header GA4 event handling.

Also simplifies the element selectors – @dkoo or @leogermani I'd be grateful for a review here since you were involved with the original implementation (https://github.com/Automattic/newspack-popups/pull/1136). 

1. On `trunk`, create a new above-header prompt. Insert a link (make it at `#hash` link for convenience)
2. Create an inline prompt, also with a link
3. Ensure you have GA4 configured via Site Kit
4. Use [an extension](https://chromewebstore.google.com/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna) or simply monitor `collect` requests in the network tab
5. Click the links in both prompts, observe that only the inline one triggers a GA4 event to be sent
6. Switch to this branch, observe that both prompts trigger a GA4 event when then link within is clicked
 
### Other information:

### DIFF
--- a/src/view/analytics/clicked.js
+++ b/src/view/analytics/clicked.js
@@ -8,9 +8,7 @@ import { getEventPayload, getRawId, sendEvent } from '../utils';
 
 export const manageClickedEvents = prompts => {
 	prompts.forEach( prompt => {
-		const anchorLinks = [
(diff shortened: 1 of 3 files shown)


chickenn00dle (approved):
Tested and this seems to do the trick. 

I also verified GA4 events are sent with all of our various campaign types.

PR #1282, opened 2024-03-06
fix: update no-padding popup styles

laurelfulford:
### All Submissions:

Right now, the Default and Hide Padding styles for the overlay prompts look the same. This PR makes the 'No Padding' style a true no padding style, but it's worth flagging that prior to this, that style has always had some padding.

I think it's naming/styling came from trying to be a popup variation of the inline prompt's no border style, but unlike removing the border and padding from the inline prompts, an overlay prompt with no padding and no blocks inside with padding looks broken:

image

No padding also adds a lot of flexibility to the designs, but I'd appreciate some feedback around whether this seems like an okay trade off: 

image

If we do opt to go with this, we should probably present it as something of a breaking change, as it'll change the appearance of this style in ways that may not work with live prompts that are using it.

See 1200550061930446-as-1206132801968396.

1. Create a new campaign prompt, and set it to display as an overlay. Give it some regular content.
2. Under the Styles panel on the right, set the prompt to have 'Hide Padding'.
3. Preview and note that the overlay popup has padding:

image

4. Apply the PR and run `npm run build`.
5. Refresh the preview and confirm that the popup now has no padding: 

image

6. Switch out the content blocks to add something a little more logical -- for example, a Cover Block, Media & Text Block, or Group block + padding as a base. Preview again and confirm this works well, and that there truly is no padding: 

image

7. Test on smaller screen sizes, too, and make sure things look okay.
8. As part of this PR, I also added some styles to make sure the overlay always contrasts against the overlay when using the 'No Padding' style. To test this, try to change the overlay colour from a dark to a light colour, and confirm that the close button contrasts against it; test this with a few different colours:

image

9. Test this with the 'No Overlay' option, and confirm that the close button is black.
10. Switch the overlay back on and back to a dark colour, and test the Default and Large Styles for the overlay; confirm that the close button is black for both, even though it would be white for the Hide Padding style.
11. Add a featured image to the prompt; confirm that the button colour remains white, even when the overlay is light:

image

### Other information:

### DIFF
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -1034,8 +1034,8 @@ final class Newspack_Popups_Model {
 		$overlay_color         = $popup['options']['overlay_color'];
 		$overlay_size          = 'full' === $popup['options']['overlay_size'] ? 'full-width' : $popup['options']['overlay_size'];
 		$no_overlay_background = $popup['options']['no_overlay_background'];
-		$close_button_color    = $no_overlay_background ? '#000' : self::foreground_color_for_background( $popup['options']['overlay_color'] );
-		$close_button_shadow   = $no_overlay_background ? 'transparent' : $overlay_color . 'aa';
+		$close_button_color    = ( ! $no_overlay_background && $hide_border ) ? self::foreground_color_for_background( $popup['options']['overlay_color'] ) : '#000';
+		$close_button_shadow   = ( ! $no_overlay_background && $hide_border ) ? $overlay_color . 'aa' : 'transparent';
 		$hidden_fields         = self::get_hidden_fields( $popup );
 		$is_newsletter_prompt  = self::has_newsletter_prompt( $popup );
 		$has_featured_image    = has_post_thumbnail( $popup['id'] ) || ! empty( $popup['options']['featured_image_id'] );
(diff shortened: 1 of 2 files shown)


adekbadek:
I think the offsetting of both the preview label and the close button are an issue now: 

image

image

@thomasguillot / @kevinzweerink – could you weight in here regarding the "no padding" styles? We need to find suitable placement for these two elements with the "no padding" style.

thomasguillot:
In https://github.com/Automattic/newspack-popups/pull/1282/commits/e7f1fa78727db90a7aaeefe183afdf3a536ffc7f I slightly tweaked the position of the "Preview" label

Screenshot 2024-03-07 at 10 32 31@2x

thomasguillot:
For the close button, the position doesn't bother me as much. But I wonder what @kevinzweerink thinks about this.

kevinzweerink:
Close button looks good to me! @thomasguillot 's change to the preview label looks right too.

adekbadek (changes requested):
> If we do opt to go with this, we should probably present it as something of a breaking change, as it'll change the appearance of this style in ways that may not work with live prompts that are using it.

It's a breaking change indeed, even if it fixes the feature. 

I'm in favour of not making it a breaking change – in order to avoid migration and unnecessary support burden.
As suggested on Asana, let's call this "old" option "Small Padding", and add a _real_ "Hide Padding" option.

laurelfulford:
Closing this PR in favour of a reworked version that doesn't include breaking changes 🙂

PR #1305, opened 2024-05-17
feat: prompt expiration date

adekbadek:
### All Submissions:

Adds a prompt expiration date, so prompts can be set to automatically revert to draft state at a certain date.

1. Create or edit a prompt
2. Open the new sidebar panel called "Expiration", toggle the setting on and choose a date 
3. Save the prompt
4. Using WP CLI, set an expiration date in the past (the UI will obviously block such possibility) (`wp post meta set <post-id> expiration_date '2024-03-13T10:43:06'`)
5. Verify that the `newspack_popups_check_expiry` cron event is set to run in under an hour (`wp cron event list`)
6. Either wait for an hour or trigger the cron event (`wp cron event run newspack_popups_check_expiry`)
7. Edit to post – observe that it has been reverted to draft, with a notice saying so at the top of the editor
8. Publish the post, observe that the expiration date has been removed and there's a notice saying so

### Other information:

---
  - https://app.asana.com/0/0/1205701744717479

### DIFF
--- a/src/editor/ExpirationPanel.js
+++ b/src/editor/ExpirationPanel.js
@@ -4,7 +4,7 @@
 import { __, sprintf } from '@wordpress/i18n';
 import { ToggleControl, DatePicker } from '@wordpress/components';
 import { isInTheFuture } from '@wordpress/date';
-import { useEffect, useState } from '@wordpress/element';
+import { useEffect, useState, useMemo } from '@wordpress/element';
 
 const ExpirationPanel = ( {
 	expiration_date = null,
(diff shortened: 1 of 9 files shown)


src/editor/ExpirationPanel.js
```diff
@@ -0,0 +1,81 @@
...
+	return (
+		<>
+			<ToggleControl
+				label={ __( 'Expiration Date', 'newspack-newsletters' ) }
+				checked={ !! expiration_date }
+				onChange={ () => {
+					if ( expiration_date ) {
+						onMetaFieldChange( { expiration_date: null } );
+					} else {
+						onMetaFieldChange( { expiration_date: new Date() } );
```
chickenn00dle: Using the current date causes some strange behavior where toggling the expiration date feature on for the first time will trigger a notice that the prompt has expired and has been reverted to draft while it has not really been reverted to draft:
Screenshot 2024-05-22 at 13 37 33

Maybe we can make the default 24 hours in the future?

src/editor/ExpirationPanel.js
```diff
@@ -0,0 +1,81 @@
...
+			);
+			// This is just for quicked feedback, the actual meta field deletion will
+			// happen on the backend.
+			onMetaFieldChange( { expiration_date: null } );
+		}
+	}, [ postStatus ] );
+
+	return (
+		<>
+			<ToggleControl
```
chickenn00dle, line 65: Nit: The setting seems a bit bare. It might be a good idea to add some helper text here to describe what the setting will do.

Screenshot 2024-05-22 at 13 34 00

Feel free to ignore this one though as I think it might be obvious enough without the helper text.

chickenn00dle (changes requested):
Looks good for the most part! Just one issue with setting an expiration date for a prompt for the first time.

src/editor/ExpirationPanel.js
adekbadek: Good catch! Fixed in 5cf76483d8991112d343b00cb6af66c9913950ec

src/editor/ExpirationPanel.js
adekbadek: Good idea! Added in 5cf76483d8991112d343b00cb6af66c9913950ec

chickenn00dle (approved):

PR #1340, opened 2024-08-28
feat: store prompt activation date

leogermani:
For future analytics purposes

### All Submissions:

Stores the activation date of the prompts so we can later use for analytics purposes

1. create or activate an inactive prompt
2. Check that the `activation_date` metadata was created with the current date
3. Deactivate the prompt
4. Confirm the `dectivation_date` metadata was created with the current date
5. Activate it again
6. Confirm the activation date was updated

### Other information:


includes/class-newspack-popups.php
```diff
@@ -537,6 +538,18 @@ public static function register_meta() {
 			]
 		);
 
+		\register_meta(
```
dkoo, line 541: Do we also need to register `deactivation_date`?

dkoo (approved):
Working as described with a non-blocking question. How do you envision using this data?

includes/class-newspack-popups.php
leogermani: yes, done

leogermani:
> Working as described with a non-blocking question. How do you envision using this data?

It is (or was) part of the Product usage measurement project scope

dkoo (approved):

PR #1341, opened 2024-08-29
fix: prioritize ras overlays for delayed prompts

chickenn00dle:
### All Submissions:

Closes https://app.asana.com/0/1207817176293825/1207922674964414/f

This fixes an issue where it's possible for delayed prompts to overlap with modal checkout during the transition in the auth + checkout flow. This happens because the moment the auth flow ends, the delayed prompt is ready to render. The problem is we don't want this to happen until all ras overlays have been added/removed during the auth + checkout flow.

This PR fixes this by reinitiating the delay whenever a ras overlay is present.

Screenshot 2024-08-29 at 11 03 24

Note: Rather than create a new `epic/ras-acc` branch for this, I'm targeting `trunk` here since the change can be useful for the current implementation of ras. But if we prefer to move this to a new epic branch in popups let me know!

Before testing be sure to checkout `epic/ras-acc` on both newspack plugin and newspack blocks

1. Set up a delayed overlay campaign targeting everyone
2. On a post or page the campaign targets add a checkout button or donation block
3. As a logged out reader, visit the page or post and click donation or checkout button BEFORE the campaign popup renders
4. Complete the auth form
5. On `trunk`, the campaign prompt will render just before the transition to modal checkout as pictured above. Confirm it does not on this branch.
6. Still on this branch, complete modal checkout. Confirm the prompt appears shortly after the modal checkout flow completes.

Bonus: Smoke test the prompt during the auth and modal checkout flows with newspack plugin and newspack blocks on `trunk`.

### Other information:

### DIFF
--- a/src/view/segmentation.js
+++ b/src/view/segmentation.js
@@ -48,7 +48,17 @@ export const handleSegmentation = prompts => {
 
 			// Unhide the prompt.
 			if ( shouldDisplay ) {
+				const delayPrompt = () => {
+					// By delay.
+					const delay = prompt.getAttribute( 'data-delay' ) || 0;
+					setTimeout( unhide, delay );
+				}
 				const unhide = () => {
+					// Prioritize RAS overlays. If there are any reinitiate the delay and return early.
+					if ( ras?.overlays && ras.overlays.get().length ) {
+						delayPrompt();
+						return;
+					}
 					prompt.classList.remove( 'hidden' );
 
 					// Log a "prompt_seen" activity when the prompt becomes visible.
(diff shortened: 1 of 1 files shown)


laurelfulford (approved):
This works super well! I tested it both with the current epic/ras-acc branches of Newspack Blocks and Plugins, and the trunk of each. I think including it in trunk is fine since it helps both cases.

The only issue I spotted was that the "blocking" doesn't work when you open a modal checkout with the variation picker first -- that specific screen doesn't prevent prompts from opening behind it.

I'm marking this as 'Approved' since this is such a nice small change that fixes most cases; I think the variation picker could be handled separately if it's possible to add the same fix to it. Let me know what you think, though!

laurelfulford:
I should have checked the other PRs 🤦‍♀️ I see the variations thing is fixed in https://github.com/Automattic/newspack-blocks/pull/1861. Don't mind me, this one's good to go!

chickenn00dle:
Thanks @laurelfulford!

PR #1347, opened 2024-09-13
fix: check for non-preview logged in user on account criteria

miguelpeixe:
### All Submissions:

1200550061930446-as-1207301960751644/f

The user account criteria only checks for the `reader.email` store item, which is not available for non-reader accounts (admin, editor, author, contributor). This is fine for admins and editors, because we allow prompts to render for preview purposes. For example, the `Newspack_Popups::is_user_admin()` method is used in the reader registration block so it renders a preview of the block.

For authors and contributors it's not the case, so they end up rendering the prompt because they don't match the criteria, but in a success state, because they don't match `Newspack_Popups::is_user_admin()` either.

This PR tweaks the criteria to account for "non-preview users".

1. Make sure you have RAS setup with RAS' default prompts
2. While in `trunk`, create a new author user
3. In a fresh session (fresh localstorage), sign in as the author
4. Confirm you get the reader registration prompt in a success state
5. Also confirm in the console that `newspack_popups_debug.matchingSegment` contains the ID of the "Not registered and not signed up for a newsletter" segment 
6. Check out this branch, clear your localstorage and refresh the page
7. Confirm you don't get prompt
8. Confirm `newspack_popups_debug.matchingSegment` is `null`

### Other information:

### DIFF
--- a/includes/class-newspack-popups-criteria.php
+++ b/includes/class-newspack-popups-criteria.php
@@ -62,7 +62,8 @@ final class Newspack_Popups_Criteria {
 			self::SCRIPT_HANDLE,
 			'newspackPopupsCriteria',
 			[
-				'config' => self::get_criteria_config(),
+				'is_non_preview_user' => is_user_logged_in() && ! Newspack_Popups::is_user_admin(),
+				'config'              => self::get_criteria_config(),
 			]
 		);
 	}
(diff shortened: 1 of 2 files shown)


adekbadek (approved):
Works as described, but the logic is a bit hard to read with the multiple negations.

PR #1371, opened 2024-11-19
fix: ensure prompts with expiration dates expire correctly

dkoo:
### All Submissions:

Expands expiration date logic implemented in #1305 so it can handle dates in various formats. The server-side cron job expected dates to be in standard SQL datetime format (`YYYY-MM-DD HH:MI:SS`), the [`DateTime` React component](https://developer.wordpress.org/block-editor/reference-guides/components/date-time/) parses dates into JS-style [ISO 8601 format](https://en.wikipedia.org/wiki/ISO_8601) (`YYYY-MM-DDTHH:MI:SS`), sometimes with a `Z` timezone suffix as well (`YYYY-MM-DDTHH:MI:SS.000Z` for UTC). The dates are stored as post meta in ISO-style format, causing the meta query comparison to fail, as it compares dates using SQL-style datetime format.

This PR also standardizes the exact time that prompts expire to midnight of the selected date, instead of the time at which the expiration date post meta was saved. This should hopefully make it more predictable exactly when prompts with future dates will expire.

Note: #1305 was working at the time of deployment, so I'm not sure when or why the SQL query using date compare stopped working.

Follow testing instructions in #1305, but repeat with datetime strings in various formats in step 4 (you can use [any format that can be parsed by `strtotime`](https://www.php.net/manual/en/function.strtotime.php)).

### Other information:

---
  - https://app.asana.com/0/0/1208791422053209

### DIFF
--- a/includes/class-newspack-popups-expiry.php
+++ b/includes/class-newspack-popups-expiry.php
@@ -28,37 +28,61 @@ final class Newspack_Popups_Expiry {
 	}
 
 	/**
+	 * Check if a given date should be considered expired.
+	 * A date is expired if it's at or before the next occuring midnight.
+	 *
+	 * @param string $date The date string to check.
+	 *
+	 * @return bool True if the date is expired, false otherwise.
+	 */
+	public static function is_expired( $date ) {
+		$expiration_date = strtotime( $date );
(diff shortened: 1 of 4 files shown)


leogermani (approved):
works as expected with different date formats

PR #1373, opened 2024-11-26
fix: copy original segments when duplicating prompt

dkoo:
### All Submissions:

#1106 refactored the way we store and handle segments from a flat array of segment IDs stored in prompt post meta to a proper custom taxonomy. However, we never added handling for segments-as-taxonomy-terms when duplicating prompts. This fixes that.

Also fixes a random PHP warning I saw when running `phpcs` (`PHP Deprecated:  preg_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated in /newspack-repos/newspack-popups/includes/class-newspack-popups.php on line 337`)

1. On `trunk`, duplicate a prompt that has one or more segments. Observe that the created duplicate prompt has no segments applied.
2. Check out this branch, repeat, and confirm that the duplicated prompt always has the same segments as the original.
3. Confirm that unit tests are passing 👇 

### Other information:

---
  - https://app.asana.com/0/0/1208801695972665

### DIFF
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -91,7 +91,7 @@ final class Newspack_Popups_Model {
 	 * Set terms for a Popup.
 	 *
 	 * @param integer $id ID of Popup.
-	 * @param array   $terms Array of terms to be set.
+	 * @param array   $terms Array of term objects or IDs to be set.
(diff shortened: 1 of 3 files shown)


miguelpeixe (approved):
🎉

PR #1389, opened 2025-01-21
fix: make a real no-padding option for popups

laurelfulford:
### All Submissions:

This PR replaces https://github.com/Automattic/newspack-popups/pull/1282. Instead of changing how the original 'No Padding' option for the overlay prompts worked, it renames it 'Small Padding', and adds an additional 'No Padding' option that has no padding.

I deviated from the approved design in https://github.com/Automattic/newspack-popups/pull/1282 once I realized that when the modal was aligned to the top, it would either need to leave a gap for the close button, or have the close button pushed off screen. The new version overlaps both the close button and 'Preview' label. Both can overlap content if the contents of the overlay don't add padding (like if they're not wrapped in a cover block):

image

@thomasguillot I'd really appreciate your opinion on that placement, if it seems to crowded/weird, or if you have suggestions for a better approach!

This PR also gives the 'Small Padding' option a 16px padding, half of what's used for the default (32px). This does change how the prompt looks, but it's a relatively small change and makes the option make sense by making it actually different.

See 1200550061930446-as-1206943440507768

1. Create a new overlay prompt, and give it the 'Hide Padding' style. Publish.
2. Apply this PR and run `npm run build`. 
3. Edit your prompt, and confirm it now uses the 'Small Padding' styles.
4. Confirm that the prompt has four styles total: Default, Small Padding, Large Padding, and No Padding.
5. Switch the prompt type to Inline, and confirm there are three styles: Default, Hide Border, and Large Border.
6. Change the styles and switch between the two different prompt types. This is to make sure the style settings still map to their 'counterparts' correctly:
* 'Default' > 'Default' and vis versa.
* 'Small Padding' > 'Hide Border' and vis versa.
* 'Large Padding' > 'Large Border' and vis versa.
* ... and if you pick 'No Padding' for the overlay, it should switch to 'Default' when you switch to an inline prompt since there isn't an equivalent. 
8. Do a visual test of the 'No Padding' option with different block settings:
* Test the different widths (small, large, full-width, etc.) [Here are some screenshots](https://cloudup.com/cNeHoGeVfqS) of the full-width size -- unlike the other placements, I removed the max-width from the content.
* Test the different placement (top, middle, the corners, etc.)
* Test different backgrounds (solid colours); the 'close' button should change to have sufficient contrast against the background, switching to white when the background is dark.
* Test different content (text blocks -- which should touch the sides -- and wrapping things in cover, Media & Text, etc. blocks). The former will look a little janky (like the first screenshot below), but the latter should look good, with the blocks touching the edges of the prompt:

CleanShot 2025-01-20 at 16 00 16

CleanShot 2025-01-20 at 16 03 53

9. Do a visual test of the 'Small Padding' option with different block settings. These should look very similar to the 'Default' padding, except with a bit less padding.
10. Smoke test the other settings -- nothing should have changed for them, but just in case!

### Other information:


thomasguillot (changes requested):
Works as described.

The only thing is I find the preview label and close button slightly too close to the edge. I think, a 4px or 8px margin would improve this. What do you think, @laurelfulford?

laurelfulford:
@thomasguillot I think that makes sense! I got a little severe with cutting how much these things might overlap with popup content, it doesn't need to be that tight 😅 

I removed the padding-specific styles for the preview label so it uses the default 4px margin, and increased the close button from 24px to 32px (still a little smaller than the default 36px). Just let me know if either could use more tweaking! 

CleanShot 2025-01-22 at 09 20 59@2x

thomasguillot (changes requested):
This is much better thank you @laurelfulford.

I didn't realise you resized the close button. I think we should stick to 36px considering in theory it's still too small to tap on mobile.

laurelfulford:
🤦‍♀️ Thanks @thomasguillot -- glad I called that out, I didn't think of that! 

I updated the PR to remove the button size change so it's always 36px again: 

CleanShot 2025-01-23 at 15 24 39

thomasguillot (approved):
It looks great. Thank you very much @laurelfulford! ❤️

laurelfulford:
Thanks for your eyes, @thomasguillot! 🙌 I'm going to shuffle the labels (remove "Needs Design Review" and switch it back to "Needs Review") so it gets a code review, unless you've reviewed that and this review counts for both. Just let me know if that's the case -- thanks! 😄

leogermani (approved):
I'm amazed at how much it takes to add a little option to the admin!

laurelfulford:
Thanks @leogermani!

> I'm amazed at how much it takes to add a little option to the admin!

I know, right?? 😆 @dkoo was a big help the admin part of this PR!

PR #1396, opened 2025-02-14
feat: add control to change close button color

laurelfulford:
### All Submissions:

This PR adds a control to change the overlay prompt close button's background colour. 

~~I made a couple of clunky decisions to try to make this change as non-disruptive as possible:~~
* ~~I opted to only make this control available when the featured image wasn't set, since the button already has a semi-transparent background in that case.~~ 
* ~~I made the button default empty, so it could fallback to the current behaviour of using the prompt's background colour to determine what colour the close X should be.~~

~~I'm happy to be talked out of either of these if they seem like they'll cause future headaches! My goal was to make sure existing prompts won't change, but there may be a better way.~~

**Edited to add:** I added a toggle based on feedback, to the above is moot!

~~(The tests are also cranky that I'm using an empty hex value 😅 If this approach seems okay I think we'll need to make that an acceptable value, but I don't want to start messing with that if it seems like a terrible idea!).~~

**Edited to add:** @dkoo helped me fix this error by removing the `format: hex-color` from our schema for the button.

1. Apply this PR and run `npm run build`.
2. Create a new popup campaign. 
3. Confirm that there's a toggle to "Customize close button background", and that it's off by default:

CleanShot 2025-02-18 at 12 09 49

4. Preview the prompt and confirm that the close button X is black with a transparent background.
5. Try changing the "Content background color" to a dark colour and preview; confirm that the Close button X contrasts against that colour and changes to white:

CleanShot 2025-02-14 at 16 41 44

6. Toggle on "Customize close button background", and confirm you get a colour picker for the "Close button background color" that defaults to `#00000000` (transparent black).
7. Try changing the "Close button background color" colour to a few different values; preview after each change, and confirm your colour is applied and that the Close X contrasts against that colour:

CleanShot 2025-02-14 at 16 42 48

CleanShot 2025-02-14 at 16 43 31

8. Add a featured image to your prompt, and preview it to confirm your customized colours still work (try checking dark and light background colours with the featured image.
10. Toggle off "Customize close button background", and preview your prompt again. Confirm that the default (transparent black and white X) are now used:

CleanShot 2025-02-14 at 16 45 35

11. Toggle back on "Customize close button background".
12. Switch your prompt from 'Overlay' to 'Inline'; confirm that the "Customize close button background" and "Close button background color" picker option disappear. 

### Other information:


laurelfulford:
@thomasguillot Adding you to this in case you have any objections to how things look, or how I've set things up to try to preserve the current behaviour (the empty background colour default, and the featured image behaviour). Thanks!

includes/class-newspack-popups-model.php
```diff
@@ -448,6 +450,7 @@ function( $value, $key ) {
 			$filtered_options,
 			[
 				'background_color'               => '#FFFFFF',
+				'close_button_background_color'  => '',
```
thomasguillot: Instead of an empty colour, I'd suggest using `#00000000` is basically black at 0% opacity.

includes/schemas/class-prompts.php
```diff
@@ -149,6 +149,13 @@ public static function get_schema() {
 							'required' => false,
 							'default'  => '#FFFFFF',
 						],
+						'close_button_background_color'  => [
+							'name'     => 'close_button_background_color',
+							'type'     => 'string',
+							'format'   => 'hex-color',
+							'required' => false,
+							'default'  => '',
```
thomasguillot: `#00000000`

presets/ras-defaults.json
```diff
@@ -13,6 +13,7 @@
             ],
             "options": {
                 "background_color": "#FFFFFF",
+                "close_button_background_color": "",
```
thomasguillot: `#00000000`

presets/ras-defaults.json
```diff
@@ -136,6 +137,7 @@
...
```
thomasguillot: `#00000000`

presets/ras-defaults.json
```diff
@@ -259,6 +261,7 @@
...
```
thomasguillot: `#00000000`

presets/ras-defaults.json
```diff
@@ -378,6 +381,7 @@
...
```
thomasguillot: `#00000000`

presets/ras-defaults.json
```diff
@@ -476,6 +480,7 @@
             "content": "<!-- wp:separator -->\n<hr class=\"wp-block-separator has-alpha-channel-opacity\"\/>\n<!-- \/wp:separator -->\n\n<!-- wp:heading -->\n<h2>{{heading}}<\/h2>\n<!-- \/wp:heading -->\n\n<!-- wp:paragraph -->\n<p>{{body}}<\/p>\n<!-- \/wp:paragraph -->\n\n<!-- wp:newspack-blocks\/donate {\"className\":\"is-style-alternate\",\"amounts\":{\"once\":[9,20,90,20],\"month\":[7,15,30,15],\"year\":[84,180,360,180]},\"disabledFrequencies\":{\"once\":false,\"month\":false,\"year\":false},\"thanksText\":\"{{thanks_message}}\",\"buttonText\": \"{{button_label}}\",\"minimumDonation\":5} \/-->",
             "options": {
                 "background_color": "#FFFFFF",
+                "close_button_background_color": "",
```
thomasguillot: `#00000000`

tests/test-model.php
```diff
@@ -30,6 +30,7 @@ public function test_popup_object_creation() {
 			$popup_object_default['options'],
 			[
 				'background_color'               => '#FFFFFF',
+				'close_button_background_color'  => '',
```
thomasguillot: `#00000000`

thomasguillot (changes requested):

thomasguillot:
The button should always be fully transparent by default, regardless of whether a featured image is present. Having different behaviours adds too much complexity, and featured images aren’t widely used on most sites. And even if they are, the change is minor and can be easily adjusted manually (if it’s even noticed).

laurelfulford:
Thanks @thomasguillot!

> The button should always be fully transparent by default, regardless of whether a featured image is present.

TBH, I'm more worried about how it should work when there's a background colour on the prompt, since the close button contrasts against the background right now. Should it keep doing that or should it now always contrast against the button background? Or should it toggle between the two depending on how transparent the button background is? 

I agree that the empty button colour is a not-great way to go about this. What do you think about a toggle, like how we handle the overlay, and have it off by default?

CleanShot 2025-02-18 at 08 33 17@2x

thomasguillot:
Hmm yes for consistency we should use a toggle like for the overlay, I didn't think of that.

thomasguillot:
The svg should contrast against the button unless the button doesn't have a background-color and in this case it should contrast against the background-color of the modal. Does this make sense? 😅

laurelfulford:
Thanks @thomasguillot! I updated the code, and updated the testing steps:

* There's now a toggle to enable/disable "Customize close button background", and it's off by default.
* When it's off, the close button will choose a contrasting colour to use for the SVG based on the prompt background.
* When it's off, the close button will use the default semi-transparent bg and white SVG when the prompt has a featured image (I know this goes counter to your feedback, but I think it's less hacky with the toggle. Let me know if you disagree though!!). 
* When it's on in all cases, the close button will use whatever the background colour is set to, and contrast the X SVG based on that colour. 

Let me know if anything seems off the mark!

includes/class-newspack-popups-model.php
laurelfulford: Done throughout!

laurelfulford:
I added this in Slack, too, but I can't figure out how to get the tests happy with this PR -- I think at least one issue is the 8-character hex value, but the other's definitely something weird I've done. Any pointers are welcome! 😅

dkoo:
Looks like the schema update in 44f739a fixed the failing unit tests, so this should be ready for a full code review!

includes/class-newspack-popups-model.php
```diff
@@ -1092,7 +1101,7 @@ class="<?php echo esc_attr( implode( ' ', $classes ) ); ?>"
 					<div class="newspack-popup__content">
 						<?php echo do_shortcode( $body ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>
 					</div>
-					<button class="newspack-lightbox__close" style="color: <?php echo esc_attr( $close_button_color ); ?>"aria-label="<?php esc_html_e( 'Close Pop-up', 'newspack-popups' ); // phpcs:ignore WordPressVIPMinimum.Security.ProperEscapingFunction.htmlAttrNotByEscHTML ?>">
+					<button class="newspack-lightbox__close" style="background-color: <?php echo esc_attr( $close_button_background_color ); ?>; color: <?php echo esc_attr( $close_button_color ); ?>"aria-label="<?php esc_html_e( 'Close Pop-up', 'newspack-popups' ); // phpcs:ignore WordPressVIPMinimum.Security.ProperEscapingFunction.htmlAttrNotByEscHTML ?>">
```
dkoo: Nit: is there any reason to render `style="background-color: "` if `$close_button_background_color` is empty?

dkoo (approved):
Works as described—just one nitpicky question below.

includes/class-newspack-popups-model.php
laurelfulford: Absolutely not! 😄 I reworked this a bit in [9e7f727](https://github.com/Automattic/newspack-popups/pull/1396/commits/9e7f72725178c40781eb789c6201a18dcb295bb2) so it only writes the `background-color` if it's needed.

thomasguillot:
Looks good, thank you @laurelfulford!

I do think the prompt sidebar is getting too cramped now. We should moved the Colors to their own sidebar panel like we do with the newsletter. But it can be done in a separate PR.

thomasguillot (approved):

laurelfulford:
Thanks @thomasguillot & @dkoo!

PR #1398, opened 2025-02-18
fix(dev-env): remove test env const check

naxoc:
### All Submissions:

This is a tiny PR to get rid of the check for `IS_TEST_ENV`. There are other ways to suppres the popups, so it is not necessary. It was messing with me because I needed that constant set for testing something else :)

Let's see if tests still work.

1. Not much to test – take a look at the code

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -619,10 +619,6 @@ final class Newspack_Popups_Inserter {
 	 * Enqueue the assets needed to display the popups.
 	 */
 	public static function enqueue_scripts() {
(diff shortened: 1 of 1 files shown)


naxoc:
Hm. Not sure I understand why the tests fail. They don't locally. It looks like the dist-folder is not present on Circle. I'll get back to this.

dkoo:
Strange how the tests run fine locally. Might have something to do with our CI jobs?

naxoc:
This is not super high priority, but I would like to figure out the discrepancy. It looks like the tests are run before the JS stuff is built on Circle. I took a look at the config and it looks very similar to the one on newspack-plugin for instance 🤔

naxoc:
I'ma close this one. It's old and not relevant :)

PR #1192, opened 2023-08-17
feat: campaigns segmentation API rearchitecture

dkoo:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/master/.github/CONTRIBUTING.md)?
* [x] Does your code follow the [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/) and [VIP Go coding standards](https://vip.wordpress.com/documentation/vip-go/code-review-blockers-warnings-notices/)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

### Changes proposed in this Pull Request:

All changes required for the Campaigns Rearchitecture initiative. Also requires https://github.com/Automattic/newspack-plugin/pull/2558 and https://github.com/Automattic/newspack-blocks/pull/1498.

### How to test the changes in this Pull Request:

All changes have already been tested and reviewed in individual PRs, so a check of the changeset should be all that's needed.

### Other information:

* [x] Have you added an explanation of what your changes do and why you'd like us to include them?
* [x] Have you written new tests for your changes, as applicable?
* [x] Have you successfully ran tests with your changes locally?

PR #1399, opened 2025-02-24
fix: restore UTM suppression feature

dkoo:
### All Submissions:

Looks like we never implemented the UTM Suppression feature in the big refactor we did in #1192. This restores it.

Cloes `1200550061930446/1208670219439755`.

1. Publish a new prompt to the "Everyone" segment or any other segment and set it to display on every pageview. Also add a value to the "UTM Suppression" option in the sidebar (make sure to add some spaces which will be URL-encoded), for example `Newspack suppression`:

Screenshot 2025-02-24 at 5 07 16 PM

2. On `release`, in a new reader session, visit the site with a `utm_source` param set to the encoded version of the UTM Suppression value you set in step 1. For example, for `Newspack suppression`, add `?utm_source=Newspack%20suppression` to the URL. Make sure you match the prompt's segment, if you set one.
3. Observe that the prompt is shown.
4. Check out this branch, repeat step 2, and confirm that the prompt is not shown.
5. In the same browser session, visit some other pages on the site and confirm that the prompt continues to be suppressed.

### Other information:

---
  - https://app.asana.com/0/0/1208670219439755

### DIFF
--- a/includes/class-newspack-popups-model.php
+++ b/includes/class-newspack-popups-model.php
@@ -969,6 +969,7 @@ final class Newspack_Popups_Model {
 		$classes              = array_merge( $classes, explode( ' ', $popup['options']['additional_classes'] ) );
 		$assigned_segments    = Newspack_Segments_Model::get_popup_segments_ids_string( $popup['id'] );
 		$frequency_config     = self::get_frequency_config( $popup );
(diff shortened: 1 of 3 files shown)


naxoc (approved):
I tested this and it works 👍

PR #1400, opened 2025-02-25
feat: prompt merge tags

miguelpeixe:
### All Submissions:

Update: After @thomasguillot's [input](https://github.com/Automattic/newspack-popups/pull/1400#issuecomment-2704261457), we're renaming "prompt tags" to "merge tags" along with other changes that are reflected below.

Implements merge tags for dynamic content rendering in campaign prompts.

By default, the `callback` argument will be used to render the tag content. If the tag has a `criteria` set, its content will be rendered in the frontend using the segmentation criteria library to pull the value from the reader data store.

| Editor | Render |
| --- | --- |
| image | image |

In the editor, [text blocks](https://github.com/Automattic/newspack-popups/blob/fe9d47aa0a5d1db0227c0dc4b1fc041904a3d879/src/editor/index.js#L151-L159) render a block control toolbar item:

image

image

image

Eventually, this should be integrated with "bits", a feature to implement inline tokens to Gutenberg's `RichText` currently under discussion: https://make.wordpress.org/core/2024/06/08/proposal-bits-as-dynamic-tokens/

1. Draft a new prompt, write a paragraph, and confirm there's no new item in the block control toolbar
2. Add `define( 'NEWSPACK_MERGE_TAGS', true );` to your `wp-config.php`
3. Refresh the page and confirm the new toolbar item renders in the block control, as in the image above
4. Add a heading, list, quote, pullquote, verse, and preformatted blocks
5. Confirm they all render the toolbar item
6. Add and select multiple paragraphs
7. Confirm the toolbar item does not render with multi-block selection
8. In each block, write some text and use the toolbar dropdown to add a merge tag
9. Confirm the merge tag is added in the correct cursor position and the block is focused with the cursor position at the end of the merge tag
10. For one of the blocks, change the merge tag to lowercase
11. Publish the prompt
12. Visit a page that renders the prompt and confirm all tags render their corresponding content
13. Open the browser console and dispatch an `article_view` activity: `newspackReaderActivation.dispatchActivity( 'article_view', { post_id: 0 } );`
14. Confirm the `{{articles_read}}` and `{{articles_read_in_session}}` are updated

### Other information:


includes/prompt-tags/class-prompt-tags.php
```diff
@@ -0,0 +1,104 @@
...
+					return get_bloginfo( 'description' );
+				},
+			]
+		);
+		self::register_tag(
+			'articles_read',
+			[
+				'title'        => __( 'Articles Read', 'newspack-popups' ),
+				'description'  => __( 'Number of articles read in the last 30 day period.', 'newspack-popups' ),
+				'use_criteria' => true,
```
leogermani: you think it's safe to assume that the tag slug will always match the criteria slug? Or maybe better to declare a string here with the criteria slug you want to use?

includes/prompt-tags/class-prompt-tag.php
```diff
@@ -0,0 +1,78 @@
...
+		$this->use_criteria = $args['use_criteria'] ?? false;
+		$this->callback     = $args['callback'] ?? '__return_empty_string';
+	}
+
+	/**
+	 * Get tag value.
+	 *
+	 * @return string
+	 */
+	public function get_value() {
```
leogermani: since this is a HTML output, and not just a value, maybe a more descriptive method name?

includes/prompt-tags/class-prompt-tags.php
miguelpeixe: It shouldn't be the case, but to be safe I changed it in 7fd317da4adc1aa354c5c81cd47c2c3e81361ed7

includes/prompt-tags/class-prompt-tag.php
miguelpeixe: Good call, renamed it in 206102ca175c47a07c14a897899e10f89a765a89

thomasguillot:
Some recommendations:

* I'd rename this feature "Merge Tags".
* I'd probably use double braces to make it clear in the text that it's a merge tag
* I'd maybe even want the tags to be uppercase (note: if user enters lowercase it should work)
* I'd remove the side panel and add it to the block toolbar and have the available tags as a dropdown so user can just click and insert

image

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24">
  <path d="M9.16601 4H10V5.5H9.16601C8.74401 5.5 8.25901 5.86 8.14001 6.543L7.63601 9.443C7.54684 9.96776 7.31701 10.4585 6.97101 10.863L6.00001 12L6.97301 13.138C7.31601 13.536 7.54301 14.028 7.63601 14.557L8.14001 17.457C8.25901 18.141 8.74401 18.5 9.16601 18.5H10V20H9.16601C7.87201 20 6.87901 18.965 6.66201 17.714L6.15801 14.815C6.11566 14.5585 6.00514 14.3181 5.83801 14.119L5.83701 14.118L3.99701 12L5.83901 9.88C5.99401 9.701 6.10901 9.46 6.15901 9.185L6.66201 6.287C6.87901 5.035 7.87201 4 9.16601 4ZM17.838 9.185L17.334 6.287C17.117 5.035 16.124 4 14.83 4H13.996V5.5H14.83C15.252 5.5 15.738 5.86 15.856 6.543L16.36 9.443C16.452 9.967 16.679 10.465 17.025 10.863L17.996 12L17.023 13.138C16.6782 13.5426 16.4491 14.0329 16.36 14.557L15.856 17.457C15.738 18.141 15.252 18.5 14.83 18.5H13.996V20H14.83C16.125 20 17.117 18.965 17.334 17.714L17.838 14.815C17.886 14.542 18.002 14.301 18.158 14.119L18.159 14.118L20 12L18.157 9.88C17.9902 9.68121 17.88 9.44113 17.838 9.185Z"/>
</svg>
```

miguelpeixe:
Thanks, @thomasguillot! It's much better now.

1. Renamed the feature in a037a32e2f02398cc5ee4d01ddd1497a0f40351a and 10ed7c495f76dde5d491ec150b4a4f537db76320
2. Moved to double curly braces and case insensitive tag in b374656ca36803e7de02de0ce2edc9b441e0ca20 (favoring uppercase)
3. Refactored the editor tool to a block control in a24c6b669c494a7bdeb9bf404c430215fab6ed96

The standard component for a toolbar dropdown doesn't allow components, just a string title, so we're pretty limited on how we render the options:

image

If you feel that's not an option I can dig deeper on a custom dropdown component.

I updated the PR description to match the new approach.

thomasguillot:
Screenshot 2025-03-11 at 10 13 14@2x

I’m not sure if displaying the actual merge tag in the dropdown is necessary, as it may not be meaningful to users. Keeping only the description seems clearer. I’ve updated the dropdown accordingly and added a max-width for better readability.

See: https://github.com/Automattic/newspack-popups/pull/1400/commits/765e9b07b6cf6d881b7d2ade952fd25f38d57293

miguelpeixe:
Thanks, @thomasguillot! Looks good to me

leogermani:
@thomasguillot is it possible to add a heading to the dropdown... I mean, a title to say what this is all about?

thomasguillot:
> @thomasguillot is it possible to add a heading to the dropdown... I mean, a title to say what this is all about?

@leogermani I'm not sure I understand the question correctly. What sort of heading did you have in mind?

From @miguelpeixe:

> The standard component for a toolbar dropdown doesn't allow components, just a string title, so we're pretty limited on how we render the options

leogermani:
Something like this

image

leogermani:
@miguelpeixe Everything working and looking good!

I just saw one little glitch that maybe it's worth investigating. My site has a reg wall, so the number of articles read was always zero, because it doesn't count when you visit a gated post normal.

But then I went throught the registration, and when I got offered to subscribe to newsletters, I closed the modal, and the window did not refresh automatically... then I refreshed it and in this request the value for number of articles read was empty (not zero). No errors in the console. I continued browsing the site and then it started counting normally again.

So it was just this one page load right after logging in to a reg wall that had this empty value

thomasguillot:
> Something like this

Like @miguelpeixe said, this would then require a custom dropdown. And frankly, I'm not convinced. You don't see this sort of behaviour for the other dropdowns.
As long as this is well-documented, I don’t think it’s a major concern.

leogermani (approved):

PR #1414, opened 2025-04-02
feat(segments): add screen size criteria

naxoc:

 This adds a segmentation for device sizes. I used these sizes suggested by Miguel:

image

For sniffing out the device, I was going to use [window.matchMedia()](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia) but since I would have to query multiple times to match all devices chosen, I went with a more hand held version where I get width and height (browser support is the same or better than matchMedia) and return as soon as we have a match.

Note that the labels (e.g. Mobile (large phones) - 480/768 px wide) is a bit long and cryptic, and I wanted `description` an `help` to help explain that. Not sure if it's a bug, but the (i) info icon does nothing on `wp-admin/admin.php?page=newspack-audience-campaigns#/segments` when clicked and the description is not shown on any items.

1. You need this PR from `newspack-plugin` too: https://github.com/Automattic/newspack-plugin/pull/3880
2. Build both repos `npm run build`
3. Go to `wp-admin/admin.php?page=newspack-audience-campaigns#/segments` and add a segment with some devices.
4. Add some prompts to the segment
5. Go to a page that should display the prompt and resize the browser to ensure that the device-sizes you chose in the segment show the prompt and the ones that don't - don't.

### Other information:

---
  - https://app.asana.com/0/0/1209596441047349

### DIFF
--- a/src/criteria/default/devices.js
+++ b/src/criteria/default/devices.js
@@ -1,49 +1,27 @@
-import { setMatchingFunction } from '../utils';
+import {setMatchingFunction} from '../utils';
 
 setMatchingFunction('devices', (config) => {
 	const selectedDevices = Array.isArray(config.value) ? config.value : [];
-
 	if (selectedDevices.length === 0) {
 		return false;
 	}
 
 	const width = window.innerWidth;
-	// On desktop if the screen is wide enough, no need for more checks.
-	if (width > 1280 && selectedDevices.includes('desktop')) {
+	if (width >= 1280 && selectedDevices.includes('desktop')) {
+		return true;
+	}
+	if (width <= 1024 && width > 768 && selectedDevices.includes('laptop')) {
+		return true;
+	}
+	if (width <= 768 && width > 480 && selectedDevices.includes('tablet')) {
+		return true;
+	}
+	if (width <= 480 && width > 360 && selectedDevices.includes('mobile')) {
+		return true;
+	}
+	if (width < 360 && selectedDevices.includes('mobile_small')) {
 		return true;
 	}
 
-	const isPortrait = window.innerHeight > width;
-	const orientationMode = isPortrait ? 'portrait' : 'landscape';
-
-	const breakpoints = {
-		mobile_small: {
-			portrait: { min: 0, max: 360 },
-			landscape: { min: 0, max: 640 },
-		},
-		mobile: {
-			portrait: { min: 361, max: 480 },
-			landscape: { min: 641, max: 768 },
-		},
-		tablet: {
-			portrait: { min: 481, max: 768 },
-			landscape: { min: 769, max: 1024 },
-		},
-		laptop: {
-			portrait: { min: 769, max: 1024 },
-			landscape: { min: 1025, max: 1280 },
-		},
-	};
-
-	for (const device of selectedDevices) {
-		if (!breakpoints[device]) {
-			continue;
-		}
-
-		const { min, max } = breakpoints[device][orientationMode];
-		if (width >= min && width <= max) {
-			return true;
-		}
-	}
 	return false;
 });
--- a/src/criteria/default/devices.js
+++ b/src/criteria/default/devices.js
@@ -10,13 +10,13 @@ setMatchingFunction('devices', (config) => {
 	if (width >= 1280 && selectedDevices.includes('desktop')) {
 		return true;
 	}
-	if (width <= 1024 && width > 768 && selectedDevices.includes('laptop')) {
+	if (width >= 1024 && width < 1280 && selectedDevices.includes('laptop')) {
 		return true;
 	}
-	if (width <= 768 && width > 480 && selectedDevices.includes('tablet')) {
+	if (width >= 768 && width < 1024 && selectedDevices.includes('tablet')) {
 		return true;
 	}
-	if (width <= 480 && width > 360 && selectedDevices.includes('mobile')) {
+	if (width >= 360 && width < 768 && selectedDevices.includes('mobile')) {
 		return true;
 	}
 	if (width < 360 && selectedDevices.includes('mobile_small')) {
--- a/src/criteria/default/devices.js
+++ b/src/criteria/default/devices.js
@@ -1,0 +1,52 @@
+import { setMatchingFunction } from '../utils';
+
+setMatchingFunction('devices', (config) => {
+	const selectedDevices = Array.isArray(config.value) ? config.value : [];
+
+	if (selectedDevices.length === 0) {
+		return false;
+	}
+
+	// Get window dimensions once
+	const width = window.innerWidth;
+	// On desktop if the screen is wide enough, no need for more checks.
+	if (width > 1280 && selectedDevices.includes('desktop')) {
+		return true;
+	}
+
+	const isPortrait = window.innerHeight > width;
+	const orientationMode = isPortrait ? 'portrait' : 'landscape';
+
+	const breakpoints = {
+		mobile_small: {
+			portrait: { min: 0, max: 360 },
+			landscape: { min: 0, max: 640 },
+		},
+		mobile: {
+			portrait: { min: 361, max: 480 },
+			landscape: { min: 641, max: 768 },
+		},
+		tablet: {
+			portrait: { min: 481, max: 768 },
+			landscape: { min: 769, max: 1024 },
+		},
+		laptop: {
+			portrait: { min: 769, max: 1024 },
+			landscape: { min: 1025, max: 1280 },
+		},
+	};
+
+	// Check each requested type and return the matching one
+	for (const device of selectedDevices) {
+		if (!breakpoints[device]) {
+			continue;
+		}
+
+		const { min, max } = breakpoints[device][orientationMode];
+
+		if (width >= min && width <= max) {
+			return true;
+		}
+	}
+	return false;
+});
--- a/src/criteria/default/devices.js
+++ b/src/criteria/default/devices.js
@@ -1,27 +1,20 @@
 import {setMatchingFunction} from '../utils';
 
-setMatchingFunction('devices', (config) => {
+setMatchingFunction('devices', ( config, ras, { optionParams } )  => {
 	const selectedDevices = Array.isArray(config.value) ? config.value : [];
 	if (selectedDevices.length === 0) {
 		return false;
 	}
 
 	const width = window.innerWidth;
-	if (width >= 1280 && selectedDevices.includes('desktop')) {
-		return true;
-	}
-	if (width >= 1024 && width < 1280 && selectedDevices.includes('laptop')) {
-		return true;
-	}
-	if (width >= 768 && width < 1024 && selectedDevices.includes('tablet')) {
-		return true;
-	}
-	if (width >= 360 && width < 768 && selectedDevices.includes('mobile')) {
-		return true;
-	}
-	if (width < 360 && selectedDevices.includes('mobile_small')) {
-		return true;
-	}
 
-	return false;
+	return selectedDevices.some(deviceType => {
+		const device = optionParams[deviceType];
+		if (!device || !device.min_width || !device.max_width) {
+			return false;
+		}
+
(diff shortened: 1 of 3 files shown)


src/criteria/default/devices.js
```diff
@@ -0,0 +1,49 @@
...
+		},
+		laptop: {
+			portrait: { min: 769, max: 1024 },
+			landscape: { min: 1025, max: 1280 },
+		},
+	};
+
+	for (const device of selectedDevices) {
+		if (!breakpoints[device]) {
+			continue;
```
adekbadek: When the `width` is > `laptop.landscape.max`, so above any scale, the loop is skipped here, and `true` can never be returned. This means that the `'desktop'` size is never met.

src/criteria/default/devices.js
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	const width = window.innerWidth;
+	// On desktop if the screen is wide enough, no need for more checks.
+	if (width > 1280 && selectedDevices.includes('desktop')) {
+		return true;
+	}
+
+	const isPortrait = window.innerHeight > width;
+	const orientationMode = isPortrait ? 'portrait' : 'landscape';
```
adekbadek: IMO considering orientation is an overkill. In CSS, breakpoints are usually set on screen width only, and I think this convention should be followed here for simplicity.

src/criteria/default/index.php
```diff
@@ -31,6 +31,34 @@
...
 		'matching_function' => 'list__in',
 	],
+	'devices'                  => [
+		'name'        => __( 'Devices', 'newspack-popups' ),
+		'description' => __( "Segment based on the user's device", 'newspack-popups' ),
+		'help'        => __( 'The device the user is viewing the site on – e.g. mobile, tablet, etc.', 'newspack-popups' ),
+		'category'    => 'reader_engagement',
+		'options'     => [
+			[
+				'label' => __( 'Mobile (small phones) - 360/640 px wide', 'newspack-popups' ),
```
adekbadek: I'm not sure how to read `360/640 px wide`. I know it's about orientation from reading the code, but from the label it's not clear at all. But, I'd ditch the orientation consideration – see the other comment.

src/criteria/default/index.php
```diff
@@ -31,6 +31,34 @@
...
+				'value' => 'tablet',
+			],
+			[
+				'label' => __( 'Laptop - 1024/1280 px wide', 'newspack-popups' ),
+				'value' => 'laptop',
+			],
+			[
+				'label' => __( 'Desktop - 1280 px and above', 'newspack-popups' ),
+				'value' => 'desktop',
+			],
```
adekbadek, line 78: I'm not a fan of duplicating the configuration, already defined in `devices.js`, here. The options should be derived from a single source of truth about the screen sizes.

adekbadek (changes requested):
> Not sure if it's a bug, but the (i) info icon does nothing […] when clicked

It's hover-based, though I agree a click is a reasonable assumption. 

image

src/criteria/default/devices.js
naxoc: Desktops are handled above and returns early because there would be no reason to check more things.
 I'll add a comment here though so it's clearer.
 
Edit: Eh, I refeactored it.

src/criteria/default/index.php
naxoc: I agree. Just not sure how to do that. Would you do an endpoint with the values?

src/criteria/default/index.php
naxoc: Yeah, I think ditching the orientation is a good idea. Will do.

src/criteria/default/index.php
adekbadek: No need for that, you can pass data from PHP to JS with `wp_localize_script` [in here](https://github.com/Automattic/newspack-popups/blob/f18bbe1045a16ce268259a0d44f9c80888ddcf4b/includes/class-newspack-popups-criteria.php#L63-L70). The `newspackPopupsCriteria` becomes a JS global, can be read in `devices.js`.

miguelpeixe: It's standard for the criteria system to have [the configuration in one place](https://github.com/Automattic/newspack-popups/blob/bef21e0babaf2590d28b75177e171397d68d4d4e/src/criteria/default/index.php#L16-L21) and [business logic in another](https://github.com/Automattic/newspack-popups/blob/d554151bfd05d7278d86825385d337847956b037/src/criteria/default/articles-read.js).

I don't agree this should be different unless we change the standard across criteria.

adekbadek: IMO duplication should be avoided, but it's not a strong opinion in this case – the screen sizes list is not likely to change often. The fact that there is duplication somewhere else is not a strong argument for me :/ 
Anyway, I'm fine with keeping things as they are here 👌

src/criteria/default/devices.js
```diff
@@ -1,49 +1,27 @@
...
-	for (const device of selectedDevices) {
-		if (!breakpoints[device]) {
-			continue;
-		}
-
-		const { min, max } = breakpoints[device][orientationMode];
-		if (width >= min && width <= max) {
-			return true;
-		}
+	if (width <= 1024 && width > 768 && selectedDevices.includes('laptop')) {
```
adekbadek: `'laptop'` is defined as range `768-1024`, and `'desktop'` is `1280-∞`. What's between `1024` and `1280`? This case won't be covered so no prompts will be displayed in this range. Also, the new ranges are different than earlier – before, `'laptop'` was `1025-1280`. 

Sidenote: I think this would be less error-prone and readable as a series of if-else statements, to make sure all widths are covered.

adekbadek (changes requested):

src/criteria/default/index.php
miguelpeixe: Let's take this opportunity to improve what we have!

This is a great case of options that can hold more information to be used in the matching function and having it centralized in the criteria registration is ideal.

Instead of another localized variable, in #1421 I propose a standardized way to do that.

With the approach proposed in the PR, the option registration would look like this:

```php
[
	'label'  => __( 'Desktop - 1280 px and above', 'newspack-popups' ),
	'value'  => 'desktop',
	'params' => [
		'threshold' => 1280,
	],
]
```

then in the matching function:

```javascript
setMatchingFunction( 'devices', ( config, ras, { optionParams } ) => {
	console.log( optionParams ); // { desktop: { threshold: 1280 } }
} );
```

Does that make sense?

src/criteria/default/devices.js
naxoc: Fixed that. I'm not sure what you mean by a series of if-else-statements? Do you mean adding `else` to the conditionals in the series of them?

src/criteria/default/index.php
naxoc: Because the settings have human names and the conditionals have values, I'd like to keep this one as is. Other than that I agree that it's best to have config and logic use the same values.

naxoc: Oh, submitted the above without seeing your comment, @miguelpeixe . I think that is a great idea, so I'll wait for that PR.

src/criteria/default/devices.js
miguelpeixe: Do we need it to be multi-select? Treating it as thresholds, like we commonly do with CSS media queries, makes more sense to me and simplifies this a lot.

I don't think there's a use case for a campaign to render on mobile and desktop but not on tablets, which would justify such specific control.

adekbadek: >  what you mean by a series of if-else-statements?

Sorry, I should have provided and example. What I mean is that a series of if-else statements (like a `switch`, which can't be used here because of the condition complexity) would leave no room for such issues as I've described:

```js
if (a) {
    // ...
} else if (b) {
    // ...
} else {
    // ...
}
```

All cases are covered this way. 

> Do we need it to be multi-select? Treating it as thresholds, like we commonly do with CSS media queries, makes more sense to me and simplifies this a lot.

Huh, that's valid. While I agree in principle, I'd be worried about the UI here. How to convey a threshold? A slider comes to mind, but I don't think that's a good idea. In practical terms, I think the UI that's there now is fine. Using a threshold would simplify the code at the expense of added UI complexity.

naxoc: Adding `else` to the `if`s would change nothing logic-wise I think? But readability and preference – yes.

@miguelpeixe has this PR #1421 that would solve the duplicated logic once it's merged. It would also do away with the conditional because we could do something like this instead:
```js
setMatchingFunction('devices', ( config, ras, { optionParams } )  => {
	const selectedDevices = Array.isArray(config.value) ? config.value : [];
	if (selectedDevices.length === 0) {
		return false;
	}

	const width = window.innerWidth;

	return selectedDevices.some(deviceType => {
		const device = optionParams[deviceType];
		if (!device || !device.min_width || !device.max_width) {
			return false;
		}

		return width >= device.min_width && width < device.max_width;
	});

});
```

miguelpeixe: Sounds good to me! #1421 is merged

naxoc:
Alright, with #1421 merged, the JS code here could be simplified a lot.

src/criteria/default/index.php
```diff
@@ -31,6 +31,53 @@
...
+				'params' => [
+					'max_width' => 768,
+					'min_width' => 480,
+				],
+			],
+			[
+				'label'  => __( 'Mobile (small phones) - 360 px wide', 'newspack-popups' ),
+				'value'  => 'Mobile small',
+				'params' => [
+					'max_width' => 360,
```
adekbadek: There's a "hole" between "Mobile" and "Mobile small" -> if the screen size is between 360 and 480, no device will match it. Also, the list is out of sync with the breakpoints defined in the image in the PR description – 

| device | PR desc. | code |  
|--------------|-----------|-----------|
| Small mobile | <=360 | <=360 | 
| Large mobile | <=480 | <=768 | 
| Tablet | <=768 | <=1024 | 
| Laptop | <=1024 | <=1280 | 
| Desktop | >1280 | >1280 | 

So only Small mobile and Desktop match the requirements.

adekbadek (changes requested):

src/criteria/default/index.php
naxoc: You are right. That is off. But am I just confused or will those intervals never work? There will always be a gap on one end or the other? It's probably because I took it from the "Portrait" column and the original table has more context because it has landscape too?

Would this make more sense?
| Device Type   | Width (px) |
|---------------|------------|
| Desktop       | ≥ 1280     |
| Laptop        | 1024-1279  |
| Tablet        | 768-1023   |
| Phone         | 480-767    |
| Small Phone   | < 480      |

adekbadek: Yeah, that makes sense. And you're right about the gap being there anyway – I didn't catch it at first.

naxoc:
OK, that was a lot of confusion on my part with those device widths. Should be making more sense now.

src/criteria/default/devices.js
```diff
@@ -0,0 +1,20 @@
...
+	const selectedDevices = Array.isArray(config.value) ? config.value : [];
+	if (selectedDevices.length === 0) {
+		return false;
+	}
+
+	const width = window.innerWidth;
+
+	return selectedDevices.some(deviceType => {
+		const device = optionParams[deviceType];
+		if (!device || !device.min_width || !device.max_width) {
```
adekbadek: The `Mobile small` device will always fail this check, because its `min_width` is `0` and `0` is falsy. The check should rather be:

```suggestion
		if (!device || isNaN(device.min_width) || isNaN(device.max_width)) {
```

adekbadek (changes requested):

src/criteria/default/devices.js
naxoc: Yeah, good call! Fixed.

adekbadek (approved):

PR #1416, opened 2025-04-04
Fix: prevent date picker from jumping

laurelfulford:
### All Submissions:

This PR fixes an issue where the Expiration Date date picker occasionally jumped forward a day.

Best I can tell, the issue was related to `toISOString()` switching the date to UTC -- every time you pick a date, the time is set based on UTC, and then if you pick another date, the new UTC time is based off of that. So for me at -7:00 UTC, every time I change the date, the time would jump forward 7 hours. Sometimes this still landed on the same day, sometimes it landed on the next day.

You can see this by switching the `DatePicker` component to `DateTimePicker` -- it will be different based on your time zone, but for me it consistently jumped 7 hours each change:

plus-seven

My fix was to replace `toISOString()` with something more manual that also set the time to `23:59:59` but I'm open to suggestions about that -- I'm not sure it's the best approach.

See: 1200550061930446-as-1209825755063362

1. Create a new prompt, and enable Expiration Date.
2. Confirm that the selected default date is tomorrow.
3. Try picking different dates and ensure the date you pick is what ends up selected.
4. Bonus: Swap `DatePicker` for `DateTimePicker` and confirm that the time is always set to `23:59:59` regardless of the date.
5. Try saving the prompt and refreshing, and making sure the expiration date is still accurate.

### Other information:

### DIFF
--- a/src/editor/ExpirationPanel.js
+++ b/src/editor/ExpirationPanel.js
@@ -65,7 +65,7 @@ const ExpirationPanel = ( {
 
 	const defaultExpirationDate = useMemo( () => {
 		const date = new Date();
-		date.setDate(date.getDate() + 1);
+		date.setHours( date.getHours() + 24 );
 		return convertDateToString( date );
 	}, [] );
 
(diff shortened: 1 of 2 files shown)


adekbadek (approved):
Simple and gets the job done!

laurelfulford:
Thanks @adekbadek!

PR #1425, opened 2025-04-16
fix(criteria): filter on segment if passed in "view_as"

naxoc:
### All Submissions:

This adds a filter (and that might be a bit too late in the process) that filters out all prompts that don't have the right segment if the `segment=xx` is in the url. It's used for previewing segments.

Should we filter it earlier? It's pretty rare that the `$segment_id` variable will be not empty, so this is maybe the most performant so we don't have to add a query or similar.

1. Without the branch – in `wp-admin/admin.php?page=newspack-audience-campaigns#/campaigns`, click "Preview segment" on one of the segments. You get _all_ campaigns.
2. Now switch to this branch and click "preview segment" again. You should only see the segment's campaigns.
3. Test that you still get the right campaigns elsewhere on the site.

### Other information:

---
  - https://app.asana.com/0/0/1209918677142618

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -66,14 +66,13 @@ final class Newspack_Popups_Inserter {
 		$view_as_spec        = Newspack_Popups_View_As::parse_view_as();
 		$campaign_id         = isset( $view_as_spec['campaign'] ) ? $view_as_spec['campaign'] : false;
 		$include_unpublished = isset( $view_as_spec['show_unpublished'] ) && 'true' === $view_as_spec['show_unpublished'] ? true : false;
+		$segment_id          = $view_as_spec['segment'] ?? false;
 
 		// Retrieve all prompts eligible for display.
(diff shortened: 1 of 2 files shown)


miguelpeixe (changes requested):
The segmentation doesn't happen here. It's in the js:

https://github.com/Automattic/newspack-popups/blob/351d63afb568f9528619f6ffc7591d8a9639fd41/src/view/utils/segments.js#L89-L93

I'd rather avoid duplicating this logic and fix the bug directly in the js, which should just inject the segment via a query arg and let the logic work like it does for all readers.

As for the bug, I'm unable to reproduce... Maybe we're expecting 2 different behaviors? I can see your PHP will restrict the preview to only render prompts assigned to that segment, which is not how the preview is designed to work.

The preview renders all the prompts a reader matching that segment will see, which include prompts from the segment and prompts for "Everyone" (without a segment). This is working as expected for me.

naxoc:
Makes sense to move it to JS :) 

As for the expected behaviour I think you're right! I was trying to fix: 26890605006346-as-1209918677142618 but I think it's the "Everyone" that is messing with that.

I'll close this one

PR #1428, opened 2025-04-30
fix(segmentation): log page view even if page is not showing prompts

adekbadek:
### All Submissions:

Updating pageview count and donor status would only be handled on posts/pages where prompts are not disabled. If the donor landing page has prompts disabled, no reader would ever be tagged as a donor. 

1. On `trunk`,
2. Set up a segment targeting Donors, create a page, and set is as the donor landing page in Audience -> Campaigns → Settings
3. Publish a prompt and assign it to the segment
4. Observe that the prompt is not displayed after visiting the donor landing page
5. Switch to this branch – be sure to retrigger JS builds – and observe that the prompt is displayed after visiting the donor landing page

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -623,6 +623,20 @@ final class Newspack_Popups_Inserter {
 			return;
 		}
 
+		$script_handle = 'log-page-view';
+		\wp_register_script(
+			$script_handle,
+			plugins_url( '../dist/logPageView.js', __FILE__ ),
+			[],
+			filemtime( dirname( NEWSPACK_POPUPS_PLUGIN_FILE ) . '/dist/logPageView.js' ),
+			[
+				'strategy'  => 'defer',
+				'in_footer' => true,
+			]
+		);
+		\wp_localize_script( $script_handle, 'newspack_popups_view', [ 'donor_landing_page' => Newspack_Popups_Settings::donor_landing_page() ] );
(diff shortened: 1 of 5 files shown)


includes/class-newspack-popups-inserter.php
```diff
@@ -623,6 +623,20 @@ public static function enqueue_scripts() {
...
+			$script_handle,
+			plugins_url( '../dist/logPageView.js', __FILE__ ),
+			[],
+			filemtime( dirname( NEWSPACK_POPUPS_PLUGIN_FILE ) . '/dist/logPageView.js' ),
+			[
+				'strategy'  => 'defer',
+				'in_footer' => true,
+			]
+		);
+		\wp_localize_script( $script_handle, 'newspack_popups_view', [ 'donor_landing_page' => Newspack_Popups_Settings::donor_landing_page() ] );
```
dkoo: This should probably localized with a variable name other than `newspack_popups_view`, otherwise it will be overwritten by the one localized with the same name below (line 710). Probably not safe to assume that the donor landing page will always have campaigns disabled.

dkoo (changes requested):
It works, just one thing makes me nervous, below:

includes/class-newspack-popups-inserter.php
adekbadek: Good catch, thanks! Fixed in 1a0eb2e2236fe2db7f911ae781d78dbfa89db409

dkoo (approved):

PR #1433, opened 2025-05-08
fix(segmentation): "should display prompt" logic

miguelpeixe:
### All Submissions:

The UTM suppression feature implemented in #1399 has a logic error. If a prompt has UTM suppression configured, it will never reach the remaining logic to match frequency, matching segment, etc. Because the UTM suppression logic only sets the `display = false` if it's meant to suppress, the starting `let display = true;` will remain, and the prompt will always render if the reader does not match the configured UTM.

That happens primarily because it's all in `else if` statements.

To fix the logic and improve readability, I propose we place the rules in guard clauses. The only caveat is that we'd lose the ability to know if a prompt would be suppressed by multiple rules, because we'd be returning on the first hit. IMO, it's a good compromise over a safer logic.

If we really want to keep that debugging ability, I think we should refactor this a bit further so that each added rule is a separate function called by `shouldPromptBeDisplayed()` instead of maintaining long conditional statements. This way, we can better manage the rules in a systematic way.

1. To reproduce the bug, create 2 prompts (overlay and inline) with UTM suppression and assign them to a segment that matches users who have an account
2. Make sure you are in the `release` branch and in a fresh incognito session, visit pages that would render those prompts
3. Confirm that both prompts render, even though the reader doesn't match the segment
4. Checkout this branch, refresh the page and confirm the prompts don't render
5. Authenticate and confirm the prompts render
6. Add `?utm_source={value}` to the URL and confirm the prompts don't render
7. In a fresh session, go through the RAS defaults funnel and confirm that the expected prompts render for each step:
   1. First visit (anonymous)
   2. Register
   3. Subscribe
   4. Donate
8. Confirm `src/view/utils/segments.test.js` tests are passing

### Other information:

---
  - https://app.asana.com/0/0/1210191951891711

### DIFF
--- a/src/view/utils/segments.js
+++ b/src/view/utils/segments.js
@@ -122,37 +122,27 @@ export const getBestPrioritySegment = ( segments, viewAsString = null ) => {
  */
 export const shouldPromptBeDisplayed = ( prompt, matchingSegment, ras, override = null ) => {
 	const id = prompt.getAttribute( 'id' );
-	let display = true;
 	const suppression = [];
 	const debugInfo = {
 		element: prompt,
 	};
-	const suppressByUTM = prompt.getAttribute( 'data-suppression' );
 
-	// By override.
-	if ( true === override || false === override ) {
-		debugInfo.override = true;
-		if ( ! override ) {
-			display = false;
-			suppression.push( 'Prompt suppressed by override.' );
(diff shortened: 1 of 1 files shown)


dkoo (approved):
Both segmentation and UTM suppression are working well, now. Thanks for catching the logical error!

PR #1436, opened 2025-05-13
fix(pageviews): enqueue logPageView JS before segmentation JS

dkoo:
### All Submissions:

Strangely, I didn't see this when testing #1428, but after it was merged to trunk I started seeing it happen. It seems like there's a race condition between logging the pageview and checking user pageviews for segmentation purposes, and that sometimes segmentation logic will run before the pageview is logged. This can result in inaccurate pageview counts, or worse, an uncaught exception for new anonymous readers. cc @adekbadek 

1. To make it easier to test, publish a prompt that will be displayed only on the first pageview.
2. On `trunk`, start a new incognito session and view a page that will render prompts. Observe a JS console error: `Uncaught TypeError: Cannot read properties of null (reading 'day')` that prevents segmentation logic from completing, which in turn prevents any prompts from being displayed on this pageview and any subsequent pageviews. (You may need to repeat this a few times to get it to happen since it's a race condition.)
3. Check out this branch and retry with a fresh incognito session. Confirm that the prompt created in step 1 is displayed on the first pageview, then not again afterward, and that no JS console errors occur.

### Other information:


miguelpeixe (changes requested):
I'm throttling my network and still getting the issue with this change. It's not great to have these scripts depend on each other.

We can allow them to run independently:
1. Store a global var stating that the current page view has been logged
2. Handle empty `pageviews` on `shouldPromptBeDisplayed()` and check the global var to account for the current view or not

dkoo:
@miguelpeixe I'm worried that even if we go the route of reading a global var, the race condition might still exist between the two separate files. What if we just allowed the Campaigns view.js to be enqueued on every page, but only run segmentation functions if prompts are enabled? The total JS filesize is small (46kb not minified) and about 25kb of that is boilerplate webpack bundle stuff. I'm not convinced that separating into two files is optimizing performance in a meaningful way.

Here's what that looks like: c22f3f6

Let me know if you'd still like to pursue enqueuing a separate JS file for logging pageviews.

miguelpeixe (changes requested):
I like that better! Just need to remove the unused `log-page-view.js` file

dkoo:
Whoops! e606b8b

miguelpeixe (approved):
💯

PR #1467, opened 2025-08-20
fix: scheduling expiry check for midnight using AS

leogermani:
### All Submissions:

This PR fixes a few things with the scheduled event to check popups expiration.

In the current behavior, the event is not necessarily scheduled to the top of the hour. So even though the check runs hourly, it can end up only executing much later than midnight.

* instead of running it on every hour, schedule it to the next midnight and run it daily
* Make sure to schedule it for the top of the hour at midnight
* Give preference to use Action Scheduler instead of regular wp cron

1. Checkout this branch
2. Go to Tools > Scheduled actions ans search for "expiry"
3. Confirm you see the event scheduled for midnight

### Other information:

### DIFF
--- a/includes/class-newspack-popups-expiry.php
+++ b/includes/class-newspack-popups-expiry.php
@@ -21,9 +21,43 @@ final class Newspack_Popups_Expiry {
 	 */
 	public static function init() {
 		add_action( 'transition_post_status', [ __CLASS__, 'transition_post_status' ], 10, 3 );
+		add_action( 'init', [ __CLASS__, 'register_recurring_event' ] );
 		add_action( self::CRON_HOOK, [ __CLASS__, 'revert_expired_to_draft' ] );
-		if ( ! wp_next_scheduled( self::CRON_HOOK ) ) {
-			wp_schedule_event( time(), 'hourly', self::CRON_HOOK );
+	}
+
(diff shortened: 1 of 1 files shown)


includes/class-newspack-popups-expiry.php
```diff
@@ -21,9 +21,43 @@ final class Newspack_Popups_Expiry {
...
+	}
+
+	/**
+	 * Register the recurring event, prioritizing the use of Action Scheduler.
+	 *
+	 * @return void
+	 */
+	public static function register_recurring_event() {
+		if ( is_admin() && function_exists( 'as_schedule_recurring_action' ) ) {
+			self::register_check_expiry_as_event();
```
chickenn00dle, line 35: Do we need to unschedule the wp cron event if it is scheduled here as well?

chickenn00dle (approved):
Looks good. Left a nonblocking question, but this is good to go 👍

includes/class-newspack-popups-expiry.php
leogermani: My thought is that the event would run one more time and then not be rescheduled... but it wouldn't hurt to unschedule it... what I want to avoid is still keep checking if it's scheduled every time

PR #1480, opened 2025-10-11
fix: don’t add overlays to ras store if locked via front-end metering

dkoo:
### All Submissions:

Avoids adding overlay prompts to the global `overlays` store on pages content-gated via front-end metering.

See https://github.com/Automattic/newspack-plugin/pull/4230.

See https://github.com/Automattic/newspack-plugin/pull/4230 for testing with front-end metering.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -653,10 +653,7 @@ final class Newspack_Popups_Inserter {
 			\wp_enqueue_style( $admin_script_handle );
 		}
 
-		if ( self::assess_has_disabled_popups() ) {
-			return;
-		}
-
+		// Note: always enqueue scripts even when prompts are disabled, to allow page view logging.
 		$script_handle = 'newspack-popups-view';
 
 		if ( ! self::is_amp() ) {
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -653,6 +653,10 @@ final class Newspack_Popups_Inserter {
 			\wp_enqueue_style( $admin_script_handle );
 		}
 
+		if ( self::assess_has_disabled_popups() ) {
+			return;
+		}
+
 		$script_handle = 'newspack-popups-view';
 
 		if ( ! self::is_amp() ) {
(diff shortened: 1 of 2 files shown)


includes/class-newspack-popups-inserter.php
```diff
@@ -653,6 +653,10 @@ public static function enqueue_scripts() {
 			\wp_enqueue_style( $admin_script_handle );
 		}
 
+		if ( self::assess_has_disabled_popups() ) {
+			return;
+		}
+
```
miguelpeixe: I think this breaks the solution from #1436 and prevents pageviews from being logged on pages without prompts.

miguelpeixe (changes requested):
Besides the issue in the inline comment, I still get [criteria.js enqueued](https://github.com/Automattic/newspack-popups/blob/2dcdf7a7f6f9ad96fdc9b77df3bb5c64e6c82abe/includes/class-newspack-popups-criteria.php#L55-L64)

includes/class-newspack-popups-inserter.php
dkoo: Ah yes, you're right, we've been through this many times before

dkoo: Restored and added a reminder note: b57af54

miguelpeixe: I also mentioned that criteria.js is still enqueued: https://github.com/Automattic/newspack-popups/blob/2dcdf7a7f6f9ad96fdc9b77df3bb5c64e6c82abe/includes/class-newspack-popups-criteria.php#L55-L64

dkoo: Is criteria.js being enqueued causing any issues?

dkoo: Decided to remove all of the performance improvements from this PR since it's not really the purpose of the changes here.

miguelpeixe: I was under the impression that `view.js` was the only one required to stay, as it's the script handling pageview logging. As discussed internally, let's keep them all enqueued, as it's not the purpose of this PR.

miguelpeixe (approved):

PR #1499, opened 2025-11-19
fix(apple-news): disable prompts when exporting to apple news

dkoo:
### All Submissions:

Prevents prompts from being injected into post content while the [Apple News plugin](https://github.com/alleyinteractive/apple-news) is exporting posts. Segmentation behavior and prompt styles won't carry over gracefully to Apple News.

Also does a little reorganization to consolidate several callbacks on the `newspack_popups_assess_has_disabled_popups` to a single callback.

Also allows prompts to appear on Product pages, since the original reason for disabling prompts on all Product pages (AMP) is no longer a factor.

If you have an Apple News publishing account, you can test this by [downloading Apple News JSON for a post](https://github.com/alleyinteractive/apple-news/wiki/Usage#downloading-posts) and confirming that prompts that would otherwise be rendered in post content never appear in these exports.

You can also run this code via WP CLI to view the export JSON—just replace `1234` with a post ID to test with (h/t @adekbadek):

```
wp eval "echo (new \Apple_Actions\Index\Export((new Admin_Apple_Settings())->fetch_settings(), 1234))->fetch_exporter()->export();" | jq '.'
```

Additionally:

1. Smoke test the conditions that have been consolidated into the `disable_prompts` method. Prompts should be disabled when:

  - The "Disable prompts on this post or page" toggle is ON for specific posts and pages
  - The `newspack_newsletters_assess_has_disabled_popups` filter returns `true` (temporarily add a filter callback to set this value)

2. View a single product page and confirm that prompts CAN now appear on these pages. Smoke test adding the product to your cart and checkout out via normal Woo checkout.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -31,13 +31,6 @@ final class Newspack_Popups_Inserter {
 	protected static $segments = [];
 
 	/**
-	 * Whether we're exporting to Apple News.
-	 *
-	 * @var boolean
-	 */
-	private static $is_apple_news_exporting = false;
-
-	/**
 	 * Whether we've already inserted prompts into the content.
 	 * If we've already inserted popups into the content, don't try to do it again.
 	 *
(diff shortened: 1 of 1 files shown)


includes/class-newspack-popups-inserter.php
```diff
@@ -38,6 +38,71 @@ final class Newspack_Popups_Inserter {
...
+			}
+		);
+
+		add_action(
+			'newspack_blocks_homepage_posts_after_render',
+			function() {
+				add_filter( 'the_content', [ $this, 'insert_popups_in_content' ], 1 );
+			}
+		);
+	}
```
dkoo, line 79: Just moving the `__construct` method to the top of the class for consistency with our other code

includes/class-newspack-popups-inserter.php
```diff
@@ -469,8 +467,10 @@ public static function insert_popups_in_content( $content = '' ) {
 		$filtered_content = explode( "\n", $content );
 		$post_content     = explode( "\n", $post->post_content );
 		if (
+			// If prompts are disabled for this post.
+			self::assess_has_disabled_popups()
```
dkoo, line 471: Hit the `newspack_popups_assess_has_disabled_popups` filter earlier when deciding whether to insert prompts into post content, to avoid running some unnecessary calculations before it runs in the `popups_for_post()` method (it must still be run in that method as it's called separately for other purposes too).

claudiulodro (approved):
The feature and refactoring all work well in my testing and code looks solid. Thanks!

dkoo:
Thanks for the review, @claudiulodro!

PR #1500, opened 2025-11-20
fix: make sure custom placements follow post type restrictions

laurelfulford:
### All Submissions:

This PR makes the the Manual and Custom prompt placements treat the post type restrictions the same way, but I'm less sure that this is the way it was designed to work 😅 

Prior to this change, the Manual placement respected the post type setting, and the Custom placement did not. This PR fixes an error in the _Manual_ placement code to make them act the same way, and removes a check for either in the `should_display()` function. This implies to me that the Custom Placements were acting correctly, but given how both Custom and Manual placements can be added to widget areas, it's helpful to be able to control their visibility like this.

(Widget visibility would also work with the classic theme, but not if we wanted to do a similar quasi-in-template placement in the block theme, since they wouldn't be widgets there).

Note that if you limit these prompts to just posts, they show up on the archives; the same thing happens with other placements, like above header. It's not a post type, but should we add an additional option to that list for archive pages?

See NPPM-2314.

1. Set up three prompts: a Custom placement, Manual placement, and Inline (to be placed in the above-header spot for comparison).
2. Set all three prompts to only show for Posts.
3. Set up the Custom placement and Manual placement prompts in the Above footer widget area.
4. View on the front-end; note that the Manual placement and Above Header prompts only show on posts (and archives), but the Custom placement ignores this setting.
5. Apply this PR.
6. Repeat step 4; confirm all the prompts are only showing up for the same post type. 
7. Change the criteria of all three prompts (to a different post type, or to a particular tag or category, or add to a segment), and confirm they still respect it.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-custom-placements.php
+++ b/includes/class-newspack-popups-custom-placements.php
@@ -258,7 +258,7 @@ final class Newspack_Popups_Custom_Placements {
 	 */
 	public static function is_custom_placement_or_manual( $prompt ) {
 		return (
-			'manual' === $prompt['options']['frequency'] ||
+			'manual' === $prompt['options']['placement'] ||
 			in_array( $prompt['options']['placement'], self::get_custom_placement_values() )
 		);
 	}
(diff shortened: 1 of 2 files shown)


dkoo (approved):
I'm pretty sure we did the override because some publisher expected it to work that way in the past, but this seems like a more logical behavior to me.

laurelfulford:
Thanks @dkoo!

It just dawned on me that if a publisher doesn't want these placements to follow post/segment/whatever restrictions, they can just make them visible to everyone in all cases 😄 I'll make sure this gets a blurb in our release notes, though!

PR #1507, opened 2026-01-05
fix(prompt-expiration): evaluate expiration timestamp in site timezone

dkoo:
### All Submissions:

Ensures that when evaluating whether a prompt's `expiration_date` timestamp has passed, the comparison is done with the next occurring midnight in the site's timezone, rather than in UTC. This should prevent premature prompt expiration for sites west of UTC, and delays in expiration for sites in timezones east of UTC.

Closes NPPM-2503.

Kind of hard to test this without waiting 24 hours, but for clarity's sake:

1. In Settings > General, set your site's timezone to a large difference from UTC, such as UTC-8.
2. Publish a prompt with an expiration date set to tomorrow.
3. Wait until UTC midnight has passed, but it's still the prior day in your site's timezone. On `release`, observe that the prompt is unpublished even though there are still 8 hours left in the day in your site's timezone.
4. Check out this branch, repeat, and confirm that the prompt isn't unpublished until midnight of the site's timezone, not UTC.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-expiry.php
+++ b/includes/class-newspack-popups-expiry.php
@@ -81,7 +81,7 @@ final class Newspack_Popups_Expiry {
 		$now      = current_datetime(); // Site timezone aware.
 		$tomorrow = $now->setTime( 0, 0 )->modify( '+1 day' ); // Next occurring midnight in the site's timezone.
 
-		return $expiration_date <= $tomorrow;
+		return $expiration_date < $tomorrow;
 	}
 
 	/**
--- a/includes/class-newspack-popups-expiry.php
+++ b/includes/class-newspack-popups-expiry.php
@@ -70,11 +70,18 @@ final class Newspack_Popups_Expiry {
 	 * @return bool True if the date is expired, false otherwise.
 	 */
 	public static function is_expired( $date ) {
-		$expiration_date = strtotime( $date );
-		if ( $expiration_date && $expiration_date <= strtotime( 'tomorrow' ) ) {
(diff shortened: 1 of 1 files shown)


includes/class-newspack-popups-expiry.php
```diff
@@ -70,11 +70,18 @@ public static function register_check_expiry_as_event() {
...
-		return false;
+		$expiration_date = date_create_immutable( $date, wp_timezone() );
+		if ( false === $expiration_date ) {
+			return false;
+		}
+
+		$now      = current_datetime(); // Site timezone aware.
+		$tomorrow = $now->setTime( 0, 0 )->modify( '+1 day' ); // Next occurring midnight in the site's timezone.
+
+		return $expiration_date <= $tomorrow;
```
miguelpeixe: This works well, except when the date argument is the next day without any time set (so it matches midnight):

```php
Newspack_Popups_Expiry::is_expired( gmdate( 'F j, Y', strtotime( '+1 day' ) ) )
```

To fix it, we must tweak this comparison:

```suggestion
		return $expiration_date < $tomorrow;
```

miguelpeixe (changes requested):

includes/class-newspack-popups-expiry.php
dkoo: Nice catch! I couldn't replicate saving that meta value without a time, but this should make the method even more bug-proof.

miguelpeixe (approved):

PR #1509, opened 2026-01-06
fix: rename prompt_id to newspack_popup_id for consistency

laurelfulford:
### All Submissions:

Along with https://github.com/Automattic/newspack-plugin/pull/4378, this PR renames `prompt_id` to `newspack_popup_id` in Data Events for consistency based on feedback.

Question (I'll ask this in Linear, too!): do we need to do anything to make sure existing data that's been captured in GA4 "understands" this change? Or is it basically a breaking change?

See NPPM-2486

1. Apply this PR and https://github.com/Automattic/newspack-plugin/pull/4378 (it's just a readme update, but both PRs need to go together)
2. Set up some prompts (inline and popup).
3. Either using GA4 or an extension like [this](https://chromewebstore.google.com/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna), check the information coming with the `np_prompt_interaction` event as you interact with the site. Ensure the prompt's ID is being tracked as `newspack_popup_id`.
4. Confirm whether the readme change in https://github.com/Automattic/newspack-plugin/pull/4378 looks okay.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-data-api.php
+++ b/includes/class-newspack-popups-data-api.php
@@ -105,8 +105,8 @@ final class Newspack_Popups_Data_Api {
 			return $data;
 		}
 
-		$data['prompt_id']    = $popup['id'];
-		$data['prompt_title'] = $popup['title'];
+		$data['newspack_popup_id'] = $popup['id'];
+		$data['prompt_title']      = $popup['title'];
 
 		if ( isset( $popup['options'] ) ) {
 			$data['prompt_frequency'] = self::get_frequency_summary( $popup );
@@ -186,8 +186,8 @@ final class Newspack_Popups_Data_Api {
 	 */
 	public static function get_rendered_popups( $popup ) {
 		$data = self::get_popup_metadata( $popup );
-		if ( ! empty( $data['prompt_id'] ) ) {
-			self::$popups[ $data['prompt_id'] ] = $data;
+		if ( ! empty( $data['newspack_popup_id'] ) ) {
+			self::$popups[ $data['newspack_popup_id'] ] = $data;
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


dkoo (commented):
Question that I don't know the answer to, maybe @miguelpeixe has an opinion on this? I still see a few instances of `prompt_id` being set on the Data Events activities that are being dispatched by this plugin:

1. https://github.com/Automattic/newspack-popups/blob/fix/rename-prompt-id-to-newspack-popup-id/src/view/analytics/seen.js#L11
2. https://github.com/Automattic/newspack-popups/blob/fix/rename-prompt-id-to-newspack-popup-id/src/view/utils/index.js#L77
3. https://github.com/Automattic/newspack-popups/blob/fix/rename-prompt-id-to-newspack-popup-id/src/view/utils/segments.js#L170
4. https://github.com/Automattic/newspack-popups/blob/fix/rename-prompt-id-to-newspack-popup-id/src/view/utils/segments.test.js#L114

These aren't strictly related to the GA4 events, but do we want to standardize around the `newspack_popup_id` name? If so, we'll probably want to make sure that any segmentation-related logic (such as in `#3` above) handles both names to avoid breaking making this a breaking change.

laurelfulford:
Eep, good catch - thanks @dkoo! Let me know if it'd make sense to raise this in Linear, too, and flag for @kmwilkerson!

miguelpeixe:
Given that we use `prompt_seen` as the event name, I don't mind keeping `prompt_id` for the data events schema. We don't really prefix properties with `newspack_` for data events either, so that seems more like a GA requirement. Not a super strong opinion, though.

dkoo (approved):
> Given that we use `prompt_seen` as the event name, I don't mind keeping `prompt_id` for the data events schema. We don't really prefix properties with `newspack_` for data events either, so that seems more like a GA requirement. Not a super strong opinion, though.

Sounds good to me—it doesn't look like we're directly using the name `prompt_id` anywhere else outside of the Data Events activities, so let's keep it as-is for now.

laurelfulford:
Thanks @dkoo and @miguelpeixe!

PR #1511, opened 2026-01-08
fix: ignore corrections when validating content for inline prompts

chickenn00dle:
### All Submissions:

* [ ] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-plugin/blob/trunk/.github/CONTRIBUTING.md)?

Closes # https://linear.app/a8c/issue/NPPM-2507/high-priority-corrections-hide-inline-campaign-prompts

This PR also requires https://github.com/Automattic/newspack-plugin/pull/4385

This PR ignores corrections when validating content before rendering inline prompts. This is because corrections are always added to the start or end of the content and should not impact whether an inline prompt renders.

Screenshot 2026-01-08 at 15 10 07

Screenshot 2026-01-08 at 15 10 12

1. Create an inline campaign prompt set to display at 0% of article content (top)
2. Create another inline campaign prompt set to display at 100% of article content (bottom)
3. Create a post and observe that both display on the frontend
4. Add a correction to the post and set Priority to "High" (top)
5. Add another correction to the post and set Priority to "Low" (bottom)
6. On `release`, observe that the inline prompts don't render on the frontend 
7. On this branch, observe that the top inline prompt appears BEFORE the top correction and the bottom inline prompt appears AFTER the bottom correction (Be sure to also check out https://github.com/Automattic/newspack-plugin/pull/4385)

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -466,7 +466,7 @@ final class Newspack_Popups_Inserter {
 			return $content;
 		}
 
-		$filtered_content = explode( "\n", $content );
+		$filtered_content = explode( "\n", self::filter_content( $content ) );
 		$post_content     = explode( "\n", $post->post_content );
 		if (
 			// Avoid duplicate execution.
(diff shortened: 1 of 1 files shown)


includes/class-newspack-popups-inserter.php
```diff
@@ -606,6 +606,34 @@ public static function should_show_admin_ui() {
...
 	}
 
+	/**
+	 * Filter specific blocks to be ignored when validating content.
+	 *
+	 * @param string $content The content.
+	 *
+	 * @return string Filtered content.
+	 */
+	private static function filter_content( $content ) {
```
miguelpeixe: Non-blocking, but I think we should rename this to `get_validation_content` or at least `get_filtered_content`

miguelpeixe (approved):
Works!

includes/class-newspack-popups-inserter.php
chickenn00dle: Good call. Fixed in https://github.com/Automattic/newspack-popups/pull/1511/changes/9a6b10ad9c48dc83675f34d86dd2762a35fb0311

miguelpeixe (approved):

PR #1526, opened 2026-02-06
fix: run do_shortcode on manual, custom placements for the block theme

laurelfulford:
### All Submissions:

While Single Prompt and Custom Placements still work in the post/page content using the block theme, it absolutely does not want to do anything with the shortcode these placements use when placed in a template/template part via the site editor. 

This PR adds a check for the block theme and the content, and runs `do_shortcode()` when you're using the block theme and not in the content, rather than inserting the shortcode block code directly. Both Copilot and Cursor suggested using a filter to make it easier to swap between these two "themes" for testing, so I went with that, too 😅 

I think this is kind of a temporary measure; NPPD-1216 proposes some changes to campaigns to make them more "block-like".

Closes NPPD-1212

1. Set up a couple prompts, one manual and one custom placement.
2. If it's not already running the block theme, switch your site to the block theme. 
3. Edit the Site, and add a Single Placement and Custom Placement block, and add your campaign prompts to them.
4. View on the front-end:

CleanShot 2026-02-06 at 15 52 23

5. Apply this PR.
6. Refresh the page and confirm you're getting prompts:

CleanShot 2026-02-06 at 15 51 47

7. Try adding the two manual placements in post/page content, and ensure they still work.
8. Switch to the Classic theme and try adding both placements to the content and widget areas, respectively.

### Other information:

### DIFF
--- a/src/blocks/custom-placement/view.php
+++ b/src/blocks/custom-placement/view.php
@@ -62,7 +62,8 @@ function render_block( $attributes ) {
 		}
 		foreach ( $prompts as $prompt_id ) {
 			$shortcode = '[newspack-popup id="' . $prompt_id . '"' . $class_names . ']';
-			$content  .= $is_block_theme
+			$render_shortcode = apply_filters( 'newspack_popups_render_custom_placement_shortcode', $is_block_theme, $attributes, $custom_placement_id, $prompt_id );
+			$content  .= $render_shortcode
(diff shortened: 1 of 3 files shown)


miguelpeixe (approved):
Good solution and tests well!

laurelfulford:
Yay! Thanks @miguelpeixe! 🙌 😄

PR #1533, opened 2026-03-04
fix(block-theme): move where above header campaign is inserted

laurelfulford:
### All Submissions:

An alternative approach to https://github.com/Automattic/newspack-popups/pull/1527. 

Out of the box, the above-header campaigns were missing block support CSS (like layout, spacing, and alignment styles) when used with the block theme. The styles need to be registered before `wp_head()` fires, but they're not; this PR moves the insertion point of the above header prompt for the block theme to fix this.

(Unfortunately it moves the prompt _inside_ of the header; if we do end up reworking the prompts to be more block-like themselves like in NPPD-1216, we should be able to use the block hooks API to move this above the header [in theory]).

Closes NPPD-1199

1. Switch to a block theme.
2. Set up an above-header campaign with a block that generates CSS, like adding a row block and setting it to spread the inside blocks out.
3. View on the front end; note that you're losing your row alignment styles (the campaign by default also doesn't have spacing, but that matches with how it works with the classic theme). The blocks also lose their 'content width' -- by default, the block content should be narrower: 632px wide normally, or 1296px wide if you're using a wide block.

In my example, I have a row block set to wide with padding on the top and bottom, and a background set for the entire campaign itself:

CleanShot 2026-03-03 at 17 01 57

5. Apply this PR.
6. Confirm that your above header campaign's row block now acts like a row block:

CleanShot 2026-03-03 at 17 00 39

Smoke testing:

1. Double check that the inline/overlay prompts (both timed and scroll depth) look okay with the same kinds of blocks. 
2. Try switching to the classic theme and double check the above header, inline, and overlay prompts.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -644,7 +644,7 @@ final class Newspack_Popups_Inserter {
 			[ 'Newspack_Popups_Model', 'is_inline' ]
 		);
 		if ( ! empty( $overlay_before_header_popups ) ) {
-			self::sort_overlays_by_specificity( $overlay_before_header_popups );
+			$overlay_before_header_popups = self::sort_overlays_by_specificity( array_values( $overlay_before_header_popups ) );
 		}
 		$popup_markup = '';
 		// Render overlays first, in specificity order.
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -611,7 +611,7 @@ final class Newspack_Popups_Inserter {
 	 */
 	public static function insert_before_header() {
 		// In block themes, prompts are inserted via the header template-part render filter.
-		if ( function_exists( 'wp_is_block_theme' ) && wp_is_block_theme() ) {
+		if ( Newspack_Popups_Model::is_block_theme() ) {
 			return;
 		}
 
@@ -632,7 +632,7 @@ final class Newspack_Popups_Inserter {
 	 * @return string Rendered content with campaign markup prepended when applicable.
 	 */
 	public static function insert_before_header_in_template_part( $block_content, $block, $instance ) {
-		if ( ! function_exists( 'wp_is_block_theme' ) || ! wp_is_block_theme() || is_admin() || self::$header_template_part_has_rendered ) {
+		if ( ! Newspack_Popups_Model::is_block_theme() || is_admin() || self::$header_template_part_has_rendered ) {
 			return $block_content;
 		}
 
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -570,7 +570,40 @@ final class Newspack_Popups_Inserter {
 	}
 
 	/**
-	 * Insert popups markup before header.
+	 * Build combined markup for all above-header prompts: overlays (sorted by specificity)
+	 * first, then inline prompts in their original order.
+	 *
+	 * Both above-header inline prompts and time-triggered overlay prompts are included,
+	 * as both must appear before the header so they are visible without scrolling.
+	 *
+	 * @return string HTML markup, or empty string if there are no above-header prompts.
+	 */
+	private static function get_before_header_markup() {
+		$before_header_popups = array_filter( self::popups_for_post(), [ 'Newspack_Popups_Model', 'should_be_inserted_above_page_header' ] );
+		if ( empty( $before_header_popups ) ) {
+			return '';
+		}
+
+		// Sort only the overlay subset by specificity — above-header inline prompts are
+		// not subject to the single visible overlay slot constraint and are left in their
+		// original order.
+		$overlay_popups = self::sort_overlays_by_specificity(
+			array_values( array_filter( $before_header_popups, [ 'Newspack_Popups_Model', 'is_overlay' ] ) )
+		);
+		$inline_popups  = array_values( array_filter( $before_header_popups, [ 'Newspack_Popups_Model', 'is_inline' ] ) );
+
+		$markup = '';
+		foreach ( $overlay_popups as $popup ) {
+			$markup .= Newspack_Popups_Model::generate_popup( $popup );
+		}
+		foreach ( $inline_popups as $popup ) {
+			$markup .= Newspack_Popups_Model::generate_popup( $popup );
+		}
+		return $markup;
+	}
+
+	/**
+	 * Insert popups markup before header (classic themes via wp_body_open).
 	 */
 	public static function insert_before_header() {
 		if ( self::$before_header_has_rendered ) {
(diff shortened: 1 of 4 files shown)


src/view/style.scss
```diff
@@ -492,6 +492,11 @@ $width__tablet: 782px;
 	&.hidden {
 		display: none;
 	}
+
+	// Remove space below the above header campaign in block themes.
```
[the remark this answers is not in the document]
laurelfulford: Fixed in [b5969ed](https://github.com/Automattic/newspack-popups/pull/1533/commits/b5969edb49085ede9de9695ebe361881ce8598be).

includes/class-newspack-popups-model.php
```diff
@@ -1006,6 +1006,7 @@ public static function generate_inline_popup( $popup ) {
 		$is_newsletter_prompt = self::has_newsletter_prompt( $popup );
 		$classes              = [ 'newspack-popup-container', 'newspack-popup', 'hidden' ];
 		$classes[]            = 'above_header' === $popup['options']['placement'] ? 'newspack-above-header-popup' : null;
+		$classes[]            = 'above_header' && self::is_block_theme() ? 'has-global-padding' : null;
```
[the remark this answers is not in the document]
laurelfulford: Fixed in [3c5b300](https://github.com/Automattic/newspack-popups/pull/1533/commits/3c5b3004c50cde84e36615ca1706074f2aafb330).

includes/class-newspack-popups-inserter.php
```diff
@@ -588,6 +611,67 @@ function( $popup ) {
...
+		$before_header_popups = array_filter( self::popups_for_post(), [ 'Newspack_Popups_Model', 'should_be_inserted_above_page_header' ] );
+		if ( empty( $before_header_popups ) ) {
+			return $block_content;
+		}
+
+		$popup_markup = '';
+		foreach ( $before_header_popups as $popup ) {
+			$popup_markup .= Newspack_Popups_Model::generate_popup( $popup );
+		}
+
```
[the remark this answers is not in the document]
laurelfulford: Fixed in [49bedea](https://github.com/Automattic/newspack-popups/pull/1533/commits/49bedea171f13ad894a9fba418f9b434d43d5740).

includes/class-newspack-popups-inserter.php
```diff
@@ -53,6 +60,7 @@ public function __construct() {
 		add_shortcode( 'newspack-popup', [ $this, 'popup_shortcode' ] );
 		add_action( 'after_header', [ $this, 'insert_popups_after_header' ] ); // This is a Newspack theme hook. When used with other themes, popups won't be inserted on archive pages.
 		add_action( 'wp_body_open', [ $this, 'insert_before_header' ] );
+		add_filter( 'render_block_core/template-part', [ $this, 'insert_before_header_in_template_part' ], 10, 3 );
```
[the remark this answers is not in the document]
laurelfulford: Added in [a9583e2](https://github.com/Automattic/newspack-popups/pull/1533/commits/a9583e2adc196942ec0fddb53ef234f8b9ec8941).

src/view/style.scss
```diff
@@ -492,6 +492,11 @@ $width__tablet: 782px;
 	&.hidden {
 		display: none;
 	}
+
+	// Remove top margin from block immediately following the prompt in block themes.
+	.wp-site-blocks & > * {
```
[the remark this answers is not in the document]
laurelfulford: Fixed in [355ef3c](https://github.com/Automattic/newspack-popups/pull/1533/commits/355ef3c06b683e22d6bed68c0daa90e4a91d35f8)

includes/class-newspack-popups-inserter.php
```diff
@@ -580,12 +595,87 @@ function( $popup ) {
...
+		}
+
+		$markup = self::get_before_header_markup();
+		if ( empty( $markup ) ) {
+			return $block_content;
+		}
+
+		self::$header_template_part_has_rendered = true;
+		return $markup . $block_content;
+	}
```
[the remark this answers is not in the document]
laurelfulford: Fixed in [4359654](https://github.com/Automattic/newspack-popups/pull/1533/commits/4359654155ca62ecd08eea6d2aee1d79160f08a7).

src/view/style.scss
```diff
@@ -492,6 +492,11 @@ $width__tablet: 782px;
 	&.hidden {
 		display: none;
 	}
+
+	// Remove top margin from block immediately following the prompt in block themes.
+	.wp-site-blocks & + * {
+		margin-block-start: 0;
+	}
```
rbcorrales, line 499: This matches even when the prompt has `display: none` (the `hidden` class), so the next sibling loses its top margin before JS evaluates segment matching. Would `:not(.hidden)` be safer here? Something like:

```css
.wp-site-blocks &:not(.hidden) + * {
```
This is still a subtle edge case (possible brief flash of missing margin before JS runs).

includes/class-newspack-popups-inserter.php
```diff
@@ -580,12 +595,89 @@ function( $popup ) {
...
+		}
+		return $markup;
+	}
+
+	/**
+	 * Insert popups markup before header (classic themes via wp_body_open).
+	 */
+	public static function insert_before_header() {
+		// In block themes, prompts are inserted via the header template-part render filter.
+		if ( function_exists( 'wp_is_block_theme' ) && wp_is_block_theme() ) {
```
rbcorrales: Minor consistency issue: this calls `wp_is_block_theme()` directly, but the `Model` class has a private `is_block_theme()` wrapper with a test override. Not a problem today since the tests use `switch_theme()`, but could become a maintenance pitfall if someone tries to use the override for `Inserter` tests. Might be worth sharing the `Model`'s method.

includes/class-newspack-popups-inserter.php
```diff
@@ -580,12 +595,89 @@ function( $popup ) {
...
+
+		$attrs = isset( $block['attrs'] ) ? $block['attrs'] : [];
+
+		// Most reliable signal when present.
+		if ( isset( $attrs['area'] ) && 'header' === $attrs['area'] ) {
+			return true;
+		}
+
+		// Many themes use non-exact slugs such as "header-post" or "site-header".
+		if ( isset( $attrs['slug'] ) && preg_match( '/(^|[-_])header([-_]|$)/', $attrs['slug'] ) ) {
```
rbcorrales, line 675: The area: 'header' path is covered by the tests, but this regex fallback isn't. Might be worth adding a couple of cases for slug-only matching (header-post, site-header) and a non-match (my-headers-archive).

includes/class-newspack-popups-inserter.php
```diff
@@ -580,12 +595,89 @@ function( $popup ) {
...
+
+	/**
+	 * Insert popups markup before the header template part in block themes.
+	 *
+	 * @param string   $block_content The rendered block content.
+	 * @param array    $block         The full block.
+	 * @param WP_Block $instance      The block instance.
+	 * @return string Rendered content with campaign markup prepended when applicable.
+	 */
+	public static function insert_before_header_in_template_part( $block_content, $block, $instance ) {
```
rbcorrales: Nit: `$instance` isn't used in the callback. Could change the hook to `10, 2` and drop the third param from the method signature.

rbcorrales (approved):
Tested locally, and it works correctly for above-header campaigns. Inline prompts also look fine. Left a few inline suggestions, nothing blocking 👍

includes/class-newspack-popups-inserter.php
laurelfulford: Good call! Added in [dcc7d99](https://github.com/Automattic/newspack-popups/pull/1533/commits/dcc7d99d63d315c1c44131bc6ae51cb917e30a2e).

includes/class-newspack-popups-inserter.php
laurelfulford: 🤦‍♀️ Thanks Ramon! Removed in [3dcd0ea](https://github.com/Automattic/newspack-popups/pull/1533/commits/3dcd0ea18699bbd48bf4ddf2d75b99c8459a8a03).

includes/class-newspack-popups-inserter.php
laurelfulford: Thanks! Added in [fd15ce6](https://github.com/Automattic/newspack-popups/pull/1533/commits/fd15ce65f83baf413450850e20099f4582c95acf).

src/view/style.scss
laurelfulford: Good call! Added in [dcc7d99](https://github.com/Automattic/newspack-popups/pull/1533/commits/dcc7d99d63d315c1c44131bc6ae51cb917e30a2e).

laurelfulford:
Thanks @rbcorrales! 🙌

PR #1544, opened 2026-04-02
fix(inserter): trim leading whitespace before content validation comparison

adekbadek:
### All Submissions:

The guard clause in `insert_popups_in_content()` compares the first line of `get_validation_content($content)` (which `trim()`s its output) against the first line of raw `$post->post_content`. When `post_content` starts with `\n`, `explode("\n", ...)` produces an empty first element on the raw side but not the validated side, causing a mismatch that silently skips all inline prompt insertion. Overlays are unaffected (they render via `wp_body_open`).

Fix: `ltrim()` the raw `post_content` before exploding, matching the behavior of `get_validation_content()`.

Closes [NPPM-2716](https://linear.app/a8c/issue/NPPM-2716).

1. Create a post with leading `\n` characters in `post_content` (via direct DB write or WP-CLI, not the block editor which normalizes them away)
2. Verify inline prompts render on the post
3. Verify inline prompts still render on normal posts without leading whitespace
4. Verify overlay prompts are unaffected

### Other information:

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -504,7 +504,7 @@ final class Newspack_Popups_Inserter {
 		}
 
 		$filtered_content = explode( "\n", self::get_validation_content( $content ) );
-		$post_content     = explode( "\n", $post->post_content );
+		$post_content     = explode( "\n", ltrim( $post->post_content ) );
 		if (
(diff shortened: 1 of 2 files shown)


dkoo (approved):
Works!

PR #1545, opened 2026-04-03
feat: add campaigns to block theme archives

laurelfulford:
### All Submissions:

This PR adds support for the campaigns in the block theme, both the popup ones (which are inserted in the classic theme using a custom hook), and inline between archive posts. 

**Edited to add:**
This PR also addresses an issue unrelated to block themes: The `archive_page_types` gate in `insert_inline_prompt_in_archive_pages` previously used `return`, which stopped running through all remaining prompts once one was skipped. This has been switched to `continue` in this PR, so each prompt is evaluated independently. This will change how this works for Classic themes, too: if multiple archive prompts have different page type restrictions, they will now each be evaluated rather than the first skip aborting the entire loop.

Closes NPPD-1125

1. On a site running the block theme, set up a few different campaign prompts: an inline one for archives that will show after every X posts, and one that will show as a popup.
2. View your archive page; ensure the prompts display as expected.
3. For the inline prompt, try changing the settings: have it not repeat, change the conditions of the archive it should show on, etc. Test as you change things and make sure it's acting as expected.
4. Try changing the archive to use a grid layout; confirm that the campaign spans the whole available width (possibly breaking the grid, depending where the campaign is inserted).
5. Switch to the Classic theme and smoke test campaigns in archives, just in case. Note that the prompt doesn't span the full width of the grid there - the block theme uses a widely used block so it made sense to me to add the CSS to the plugin, but the Classic theme is quite custom. I think it'd make more sense to add the CSS there.

### Other information:

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -62,7 +62,7 @@ final class Newspack_Popups_Inserter {
 		add_action( 'wp_body_open', [ $this, 'insert_before_header' ] );
 		add_filter( 'render_block_core/template-part', [ $this, 'insert_before_header_in_template_part' ], 10, 2 );
 		add_action( 'after_archive_post', [ $this, 'insert_inline_prompt_in_archive_pages' ] );
-		add_filter( 'render_block', [ $this, 'insert_inline_prompt_in_block_theme_archives' ], 10, 2 );
+		add_filter( 'render_block', [ $this, 'insert_inline_prompt_in_block_theme_archives' ], 10, 3 );
 		add_action( 'wp_before_admin_bar_render', [ $this, 'add_preview_toggle' ] );
 
 		// Always enqueue scripts, since this plugin's scripts are handling pageview sending via GTAG.
@@ -736,19 +736,33 @@ final class Newspack_Popups_Inserter {
 	/**
 	 * Insert inline prompts into a rendered core/post-template block on archive/home pages.
 	 *
-	 * Uses the render_block filter so this works with block themes.
+	 * Uses the render_block filter so this works with block themes. Only targets
+	 * the primary Query Loop (the one that inherits the global/main query) so
+	 * that secondary loops like "featured posts" or "you may also like" don't
+	 * receive prompt injection.
 	 *
-	 * @param string $block_content Rendered block HTML.
-	 * @param array  $block         Block data array, including 'blockName'.
+	 * @param string   $block_content Rendered block HTML.
+	 * @param array    $block         Block data array, including 'blockName'.
+	 * @param WP_Block $instance      The block instance (available since WP 5.9).
 	 * @return string Filtered block HTML.
 	 */
-	public static function insert_inline_prompt_in_block_theme_archives( $block_content, $block ) {
+	public static function insert_inline_prompt_in_block_theme_archives( $block_content, $block, $instance = null ) {
 		if ( 'core/post-template' !== $block['blockName'] ) {
 			return $block_content;
 		}
 
 		if ( ! is_archive() && ! is_home() ) {
 			return $block_content;
+		}
+
+		// Only inject into the Query Loop that inherits the main/global query.
+		// Secondary loops (custom queryId, inherit=false) should not get prompts.
+		if ( $instance instanceof WP_Block ) {
+			$query_context = $instance->context['query'] ?? [];
+			$inherits_main = ! empty( $query_context['inherit'] );
+			if ( ! $inherits_main ) {
+				return $block_content;
+			}
 		}
 
 		$archives_popups = array_filter( self::popups_for_post(), [ 'Newspack_Popups_Model', 'should_be_inserted_in_archive_pages' ] );
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -713,7 +713,7 @@ final class Newspack_Popups_Inserter {
 				|| ( is_post_type_archive() && ! in_array( 'post-type', $popup['options']['archive_page_types'] ) )
 				|| ( is_tax() && ! in_array( 'taxonomy', $popup['options']['archive_page_types'] ) )
 			) {
-				return '';
+				continue;
 			}
 
 			$archive_insertion_posts_count = intval( $popup['options']['archive_insertion_posts_count'] );
@@ -762,12 +762,16 @@ final class Newspack_Popups_Inserter {
 		}
 
 		// $parts[0] is the opening wrapper (e.g. <ul ...>); $parts[1..n] are the post items.
+		// Detect the post item tag from the first item so the campaign wrapper matches:
+		// <li> for list layouts (valid inside <ul>), <div> for grid layouts.
+		preg_match( '/^<(li|div)/', $parts[1], $tag_match );
+		$item_tag   = $tag_match[1] ?? 'li';
 		$post_count = count( $parts ) - 1;
 		$output     = $parts[0];
 
 		for ( $i = 1; $i <= $post_count; $i++ ) {
 			$output .= $parts[ $i ];
-			$output .= self::get_inline_prompt_html_for_archive_pages( $i, 'aside' );
+			$output .= self::get_inline_prompt_html_for_archive_pages( $i, $item_tag );
 		}
 
 		return $output;
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -764,12 +764,25 @@ final class Newspack_Popups_Inserter {
 
 		// $parts[0] is the opening <ul>; $parts[1..n] are the post items.
 		$post_count = count( $parts ) - 1;
-		$output     = $parts[0];
+
+		// The closing </ul> (and any content that follows it) lives in the last part because
+		// preg_split's lookahead puts everything after the final post item there. Strip it off
+		// so that any injected prompt <li> elements are inserted before </ul>, not after it.
+		$trailing = '';
+		$close_ul = strripos( $parts[ $post_count ], '</ul' );
+		if ( false !== $close_ul ) {
+			$trailing            = substr( $parts[ $post_count ], $close_ul );
+			$parts[ $post_count ] = substr( $parts[ $post_count ], 0, $close_ul );
+		}
+
+		$output = $parts[0];
 
 		for ( $i = 1; $i <= $post_count; $i++ ) {
 			$output .= $parts[ $i ];
 			$output .= self::get_inline_prompt_html_for_archive_pages( $i, 'li', $archives_popups );
 		}
+
+		$output .= $trailing;
 
 		return $output;
 	}
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -62,6 +62,7 @@ final class Newspack_Popups_Inserter {
 		add_action( 'wp_body_open', [ $this, 'insert_before_header' ] );
 		add_filter( 'render_block_core/template-part', [ $this, 'insert_before_header_in_template_part' ], 10, 2 );
 		add_action( 'after_archive_post', [ $this, 'insert_inline_prompt_in_archive_pages' ] );
+		add_filter( 'render_block', [ $this, 'insert_inline_prompt_in_block_theme_archives' ], 10, 2 );
 		add_action( 'wp_before_admin_bar_render', [ $this, 'add_preview_toggle' ] );
 
 		// Always enqueue scripts, since this plugin's scripts are handling pageview sending via GTAG.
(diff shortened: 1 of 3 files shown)


includes/class-newspack-popups-inserter.php
```diff
@@ -699,7 +713,7 @@ public static function insert_inline_prompt_in_archive_pages( $post_count ) {
 				|| ( is_post_type_archive() && ! in_array( 'post-type', $popup['options']['archive_page_types'] ) )
 				|| ( is_tax() && ! in_array( 'taxonomy', $popup['options']['archive_page_types'] ) )
 			) {
-					return;
+				return '';
```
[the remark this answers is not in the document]
laurelfulford: Fixed in [717682b](https://github.com/Automattic/newspack-popups/pull/1545/commits/717682b3007508b4dc4f6a4750cf0818e8821d32).

includes/class-newspack-popups-inserter.php
```diff
@@ -710,10 +724,53 @@ public static function insert_inline_prompt_in_archive_pages( $post_count ) {
...
+			return $block_content;
+		}
+
+		// $parts[0] is the opening wrapper (e.g. <ul ...>); $parts[1..n] are the post items.
+		$post_count = count( $parts ) - 1;
+		$output     = $parts[0];
+
+		for ( $i = 1; $i <= $post_count; $i++ ) {
+			$output .= $parts[ $i ];
+			$output .= self::get_inline_prompt_html_for_archive_pages( $i, 'aside' );
```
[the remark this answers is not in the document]
laurelfulford: Fixed in [717682b](https://github.com/Automattic/newspack-popups/pull/1545/commits/717682b3007508b4dc4f6a4750cf0818e8821d32) and tweaked in [43513a4](https://github.com/Automattic/newspack-popups/pull/1545/commits/43513a4cb2779e0a17301b1b10b4deddb9454efc)

includes/class-newspack-popups-inserter.php
```diff
@@ -710,10 +724,53 @@ public static function insert_inline_prompt_in_archive_pages( $post_count ) {
...
+
+		if ( ! is_archive() && ! is_home() ) {
+			return $block_content;
+		}
+
+		$archives_popups = array_filter( self::popups_for_post(), [ 'Newspack_Popups_Model', 'should_be_inserted_in_archive_pages' ] );
+		if ( empty( $archives_popups ) ) {
+			return $block_content;
+		}
+
```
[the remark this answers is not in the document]
laurelfulford: Fixed in [5eea1b3](https://github.com/Automattic/newspack-popups/pull/1545/commits/5eea1b3062b7ede9114ca2519b5f5b5dd5c74a24).

tests/test-insertion.php
```diff
@@ -437,6 +437,128 @@ public function test_tags_exclusion() {
...
+			]
+		);
+
+		$block_content = '
+			<ul class="wp-block-post-template">
+				<li class="wp-block-post post-type-post">Post 1 content</li>
+				<li class="wp-block-post post-type-post">Post 2 content</li>
+				<li class="wp-block-post post-type-post">Post 3 content</li>
+			</ul>';
+
```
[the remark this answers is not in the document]
laurelfulford: Fixed in [aa152eb](https://github.com/Automattic/newspack-popups/pull/1545/commits/aa152eb95ee4fe56e6da980ac87a34951c684b53), then removed in [43513a4](https://github.com/Automattic/newspack-popups/pull/1545/commits/43513a4cb2779e0a17301b1b10b4deddb9454efc) when the HTML was simplified.

includes/class-newspack-popups-inserter.php
```diff
@@ -710,10 +727,51 @@ public static function insert_inline_prompt_in_archive_pages( $post_count ) {
...
+		if ( ! $parts || count( $parts ) < 2 ) {
+			return $block_content;
+		}
+
+		// $parts[0] is the opening <ul>; $parts[1..n] are the post items.
+		$post_count = count( $parts ) - 1;
+		$output     = $parts[0];
+
+		for ( $i = 1; $i <= $post_count; $i++ ) {
+			$output .= $parts[ $i ];
```
adekbadek, line 796: 🟠 **Orphan `<li>` outside the `<ul>`.** `preg_split` with a lookahead on `<li[^>]*\bwp-block-post[\s"\']` produces parts where the **last** element contains both the final post item *and* the closing `</ul>`. The loop then concatenates that final part and appends the inline-prompt `<li>` right after – so the prompt `<li>` ends up **after** `</ul>`, as a stray list item outside any list:

```html
<ul class="wp-block-post-template">
  <li class="wp-block-post">…post 1…</li>
  <li>[PROMPT]</li>          ← correct, between posts
  <li class="wp-block-post">…post N…</li>
</ul>
<li>[PROMPT]</li>             ← stray, after </ul>
```

This fires whenever an insertion condition matches on the final post, which includes the fallback path `$archive_insertion_posts_count >= $wp_query->post_count && $post_count === $wp_query->post_count` – that branch is specifically designed to insert at the end of short lists, so every single-page archive with a small post count hits it.

Consequences:
- Invalid HTML – a stray `<li>` outside any `<ul>`/`<ol>`.
- The new `:has()` grid rule in `style.scss` doesn't apply to the orphan (it targets `.wp-block-post-template-is-layout-grid > li`). In the grid case I tested it happens to render full-width because it's a block-level sibling of the ul, but that's accidental – it won't honor grid `gap`, won't receive theme CSS targeting `.wp-block-post-template > li`, and will misplace if the parent layout changes.
- Possible duplication: when `is_repeating=true` and `(last_post % N === 0)`, the "between-posts" insertion and the "end-of-list" fallback can both fire for the last position, emitting two identical prompts in a row.

Simplest fix: peel `</ul>` (and any trailing content after the last post item) off the final part before the loop, then re-append after the loop finishes. Cleaner fix: use `WP_HTML_Tag_Processor` (WP 6.2+) to walk the rendered block and insert structurally, which also addresses the regex-fragility concern.

Repro: create a prompt with `placement=archives`, `archive_insertion_posts_count=1`, `is_repeating=true`, view any block-theme archive – DOM inspector shows the trailing `<li>` is a sibling of `<ul.wp-block-post-template>` rather than a child.

pr1545-flow-layout-orphan-li

includes/class-newspack-popups-inserter.php
```diff
@@ -710,10 +727,51 @@ public static function insert_inline_prompt_in_archive_pages( $post_count ) {
...
+	/**
+	 * Insert inline prompts into a rendered core/post-template block on archive/home pages.
+	 *
+	 * Uses the render_block filter so this works with block themes.
+	 *
+	 * @param string $block_content Rendered block HTML.
+	 * @param array  $block         Block data array, including 'blockName'.
+	 * @return string Filtered block HTML.
+	 */
+	public static function insert_inline_prompt_in_block_theme_archives( $block_content, $block ) {
```
adekbadek: 🟡 **`render_block` fires on every Query Loop.** The early-return on `blockName` and `is_archive()/is_home()` is cheap, but once they pass, `popups_for_post()` runs once per `core/post-template` render. A block-theme home or archive that contains multiple Query Loops (e.g., featured + recent + "you may also like") will inject prompts into every one of them, not just the main archive loop. Worth gating on `is_main_query()` or checking `$block['attrs']['queryId']` so only the main archive loop gets prompts.

Also: the `archive_page_types` gate inside `get_inline_prompt_html_for_archive_pages` has no `is_home()` branch – on the blog posts index, `is_category()`/`is_tag()`/etc. are all false, so a prompt restricted to a specific archive type still inserts. Pre-existing behavior on the classic-theme path, but this function extends it to block themes too. Worth a follow-up.

includes/class-newspack-popups-inserter.php
```diff
@@ -687,9 +688,25 @@ private static function is_header_template_part_block( $block ) {
...
+	/**
+	 * Get the HTML for an inline prompt on archive pages.
+	 *
+	 * @param integer $post_count      Order of the post in the posts loop.
+	 * @param string  $wrapper         Opening tag (with any attributes) to wrap the prompt in. Defaults to 'article class="entry"' to match classic theme; pass the post item tag for block themes.
+	 * @param array   $archives_popups Pre-filtered list of archive popups. If null, popups_for_post() is filtered internally.
+	 * @return string HTML output, or empty string.
+	 */
+	public static function get_inline_prompt_html_for_archive_pages( $post_count, $wrapper = 'article class="entry"', $archives_popups = null ) {
+		$tag = strtok( $wrapper, ' ' );
```
adekbadek: 🔵 **Nit: `strtok` on raw HTML wrapper is opaque.** `strtok` maintains hidden global state and the caller is effectively passing raw HTML (`'article class="entry"'`, `'li'`) to a function that parses it back out. Cleaner API: split into `$tag` and `$attrs` params, or use a context enum (`'classic' | 'block'`) and have the helper build the wrapper internally. As-is, a future caller could pass `'div onclick="…"'` and the escaping context is unclear.

tests/test-insertion.php
```diff
@@ -437,6 +437,128 @@ public function test_tags_exclusion() {
 		);
 	}
 
+	/**
+	 * Block theme archive insertion — inserts after Nth post item.
+	 */
+	public function test_block_theme_archive_insertion_basic() {
```
adekbadek, line 443: 🟡 **Tests miss the orphan-`<li>` bug.** The assertions use `strpos`/`substr_count` on the returned string, so they pass even when the prompt ends up outside the `<ul>`. A structural assertion would catch it:

```php
$pos_prompt     = strpos( $result, self::$popup_content );
$pos_closing_ul = strpos( $result, '</ul>' );
self::assertLessThan( $pos_closing_ul, $pos_prompt, 'Inline prompt is inserted before </ul>.' );
```

Two other cases worth covering while you're here:
1. The `return` → `continue` fix in `get_inline_prompt_html_for_archive_pages`: two prompts with different `archive_page_types` restrictions, assert the second still renders (under the old `return` it would have been suppressed once the first hit the skip condition). Worth a sentence in the PR description too, since it's a classic-theme behavior change.
2. The end-of-list fallback path (`$archive_insertion_posts_count >= $wp_query->post_count`): the current tests create enough posts to avoid it, but that path is the common real-world case and has no coverage on the block-theme entry point.

adekbadek (changes requested):
Thanks Laurel! Tested this in an isolated env on the PR branch, installed the block theme with 11 posts in a category, created an archive-placement inline prompt, and viewed the category archive in both the default flow layout and a 3-column grid variant. 

Requesting changes on one structural issue around the inline-prompt placement and a couple of test coverage gaps. Details in the inline comments.

includes/class-newspack-popups-inserter.php
laurelfulford: Thanks Adam! Best I could determine, `WP_HTML_Tag_Processor` doesn't let you insert HTML _yet_ so I went with the 'peel off' approach in [180b1e7](https://github.com/Automattic/newspack-popups/pull/1545/commits/180b1e759d1ec7653c0273d61e6b8e8536b733a2).

includes/class-newspack-popups-inserter.php
laurelfulford: Should be fixed in [e1ff5d6](https://github.com/Automattic/newspack-popups/pull/1545/commits/e1ff5d6d9f689ed49a3acb42ee9b8fcf7c6b8655)!

laurelfulford: I also spun up https://github.com/Automattic/newspack-popups/pull/1551 to address the second bit (no `is_home()` check). I ended up adding an option for that archive type, too, but I'm not sure if that muddies things more... 😅 Either way, it's in another PR, nothing to worry about here!

includes/class-newspack-popups-inserter.php
laurelfulford: Cleaned up in [87354ba](https://github.com/Automattic/newspack-popups/pull/1545/commits/87354ba280dad7fefe9eae9ab70be8c43de06620)!

tests/test-insertion.php
laurelfulford: Tests added in [dccb8a4](https://github.com/Automattic/newspack-popups/pull/1545/commits/dccb8a4815e7af809b20adac61e544970efe26cd), and I added a note to the PR description to highlight the `return` -> `continue` change (thanks for flagging that, I'll make sure it gets into the Upcoming Changes/Release Notes doc when this lands, too!).

laurelfulford:
Thanks for the review, @adekbadek! 🙌 I think I got everything sorted, but just let me know if that's not the case!

adekbadek (approved):
All four pieces of feedback addressed cleanly. Verified the orphan `<li>` is gone via browser DOM inspection on the block-theme category archive, and the new regression tests lock in the structural invariant. The `inherit` context check is a nice minimal gate for the main-query targeting. Nice work, and thanks for splitting the `is_home()` + archive type follow-up to #1551.

LGTM

laurelfulford:
Thanks @adekbadek! I somehow missed this in all the meetup-related stuff. Mergin'!

PR #1551, opened 2026-04-16
fix: add is_home() gate to archive_page_types and fix return to continue

laurelfulford:
### All Submissions:

This came out of feedback from https://github.com/Automattic/newspack-popups/pull/1545: 

When you set a prompt to just appear in one archive type (category, tag...), it will also appear if you have a page set as your 'posts' page. This is because there isn't a check for `is_home()` in the loop, so that passes.

This PR adds a check, as well as makes the posts page an option for showing a prompt on. 

Open question: by adding the option, I created a weird situation where the current behaviour is bubbled up in the UI, but it may not be what a publisher already has set. If someone has a prompt set to only show on category archive pages and they edit it, suddenly they will have a new option that's already checked (you can see this in the two before and after screenshots in the testing steps). I'm not sure the best way to work around that:
1. Leave it? It just reflects the way it was already working, except now you can turn it off
2. Uncheck it by default?
3. Just add the check for `is_home()` and don't actually add anything to the UI?

I'm leaning towards option 1, but I'm open to feedback on that! 

This PR also includes the return/continue switch from #1545 just so this works as expected and prompts beyond the first skipped one  😅 

1. If you don't have one set already, under Customizer > Homepage Settings, set a Posts Page to display all your latest posts.
2. In Campaigns, create a prompt that will only show in the archives.
3. Set the prompt to only show on the 'Category' archive type (the smooshed checkboxes are a known WP 7.0 RC bug, just ignore 'em in this screenshot!):

CleanShot 2026-04-16 at 12 49 58

4. Go to your posts page and note that you can see the prompts.
5. Apply this PR.
6. Edit your prompt and confirm you now can control whether it shows on the Posts page:

CleanShot 2026-04-16 at 12 55 03

7. Update the options to only show on the category archives again, and confirm the prompt no longer shows on the posts page. 
8. Test out the other options -- including showing a prompt only on the blog posts page -- to make sure it works as expected.

### Other information:


includes/class-newspack-popups-model.php
```diff
@@ -577,6 +577,11 @@ public static function get_available_archive_page_types() {
 				/* translators: archive page */
 				'label' => __( 'Taxonomies' ),
 			],
+			[
+				'name'  => 'home',
+				/* translators: archive page */
+				'label' => __( 'Posts Page' ),
+			],
```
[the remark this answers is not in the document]
laurelfulford: Fixed in [f09baed](https://github.com/Automattic/newspack-popups/pull/1551/commits/f09baedf3b14fead5a8e2028a6a4f0e871d7c562) and added text domains to the rest in the list.

tests/test-insertion.php
```diff
@@ -491,4 +491,98 @@ public function test_insertion_with_leading_whitespace_in_post_content() {
...
+				'frequency'                      => 'always',
+				'archive_insertion_posts_count'  => 1,
+				'archive_insertion_is_repeating' => false,
+				'archive_page_types'             => [ 'category' ],
+			]
+		);
+
+		self::factory()->post->create_many( 3 );
+		$this->go_to( home_url() );
+
```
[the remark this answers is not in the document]
laurelfulford: Fixed in [46a8096](https://github.com/Automattic/newspack-popups/pull/1551/commits/46a8096bbbe4f636571fde1f8c4fa8661f8e99ff).

tests/test-insertion.php
```diff
@@ -491,4 +491,98 @@ public function test_insertion_with_leading_whitespace_in_post_content() {
...
+		// Prompt B: restricted to 'category' — should render on a category archive.
+		$popup_b_content = 'Prompt-B-category';
+		self::createPopup(
+			$popup_b_content,
+			[
+				'placement'                      => 'archives',
+				'frequency'                      => 'always',
+				'archive_insertion_posts_count'  => 1,
+				'archive_insertion_is_repeating' => false,
+				'archive_page_types'             => [ 'category' ],
```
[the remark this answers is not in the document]
laurelfulford: Removed in [2f688d5](https://github.com/Automattic/newspack-popups/pull/1551/commits/2f688d54c5f91e1fa2961a9520cecafa75afbd97) -- the `return` -> `continue` change is also in #1545, it was just added here to make sure it didn't interfere with testing.

adekbadek:
On the open question: my vote is **Option 2** (`home` unchecked for existing prompts).

The `is_home()` leak was a bug, not a feature – a publisher who restricted a prompt to "Categories" and left everything else unchecked didn't consent to the home-page display. Option 1 frames it as "you always had Posts Page checked," which is a bit of a revisionist history. If someone actually wanted it on the home page, they'll find the new checkbox and opt in; if they didn't notice the leak (most won't), nothing visibly changes for them after upgrade.

Option 3 is strictly worse than Option 2 in my view: same behavior change, but no UI to opt back in.

A couple of smaller notes from skimming the diff:

- **Sequencing with #1545**: the `return` → `continue` change is duplicated. Once #1545 lands, this one needs a rebase – no big deal, just a heads-up.
- **Default for new prompts**: `home` being in `get_default_popup_archive_page_types()` is fine – new prompts get all types pre-checked, which is a reasonable onboarding default. The important part is that existing prompts' saved `archive_page_types` meta isn't retroactively backfilled to include `home`. Worth a quick check that nothing reads the defaults and merges them into stored meta on load/save.
- **Test coverage**: the new test covers the skip case. A positive test (prompt with `['home']` in `archive_page_types` does render on the blog home) would be a nice symmetric addition, and so would one asserting an existing prompt whose meta doesn't include `home` is correctly hidden after the fix.

Happy to QA in the browser once the design direction is settled.

laurelfulford:
Thanks @adekbadek! 

> Option 1 frames it as "you always had Posts Page checked," which is a bit of a revisionist history. 

I hadn't actually thought about it that way; that does seem kind of yucky 🙂 

I tweaked it so it should keep the option unchecked for prompts that already exist, while checking it for new prompts. 

I think I've covered the rest but let me know if anything looks weird once you start diving into it!

includes/class-newspack-popups-model.php
```diff
@@ -627,6 +632,15 @@ public static function get_default_popup_post_types() {
...
 	public static function get_default_popup_archive_page_types() {
+		return [ 'home', 'category', 'tag', 'author', 'date', 'post-type', 'taxonomy' ];
+	}
+
+	/**
+	 * Get the fallback archive page types for prompts saved before newer
+	 * options (like 'home') were introduced. Used when a prompt has no
+	 * archive_page_types meta, so it retains its pre-existing behavior.
+	 */
+	public static function get_legacy_default_archive_page_types() {
```
adekbadek: Two small suggestions:

1. The full list (`get_default_popup_archive_page_types`) and this one will drift the moment a new page type is added. Consider deriving one from the other, e.g. `return array_diff( self::get_default_popup_archive_page_types(), [ 'home' ] );`, or a name more descriptive than "legacy" (which ages poorly – imagine `search` being added next).
2. The docblock could call out that this fallback is load-bearing in two places: the `register_meta` `default` (above in `class-newspack-popups.php`) **and** the `wp_parse_args` fallback in `create_popup_options()` at line 486 of this file. Future maintainers touching one may not realize the other needs to match.

tests/test-insertion.php
```diff
@@ -491,4 +491,140 @@ public function test_insertion_with_leading_whitespace_in_post_content() {
 			'Inline prompt renders even when post_content has leading newlines.'
 		);
 	}
+
+	/**
+	 * Prompt restricted to categories does not appear on blog home (is_home()).
+	 */
+	public function test_archive_prompt_skipped_on_home_when_not_in_page_types() {
```
adekbadek: The three new tests each repeat ~20 lines of posts-page boilerplate (front-page creation, `show_on_front`/`page_on_front`/`page_for_posts`, `go_to`). Worth extracting into a helper – either locally or in `WP_UnitTestCase_PageWithPopups` if reusable elsewhere.

Also, for consistency with other tests in this file, consider `self::createPopup()` rather than reusing `self::$popup_id` via `Newspack_Popups_Model::set_popup_options()` – gives each test its own fixture instead of mutating shared state.

includes/class-newspack-popups-model.php
```diff
@@ -550,32 +550,37 @@ public static function get_available_archive_page_types() {
...
 			],
 			[
 				'name'  => 'taxonomy',
 				/* translators: archive page */
-				'label' => __( 'Taxonomies' ),
+				'label' => __( 'Taxonomies', 'newspack-popups' ),
+			],
+			[
+				'name'  => 'home',
+				/* translators: archive page */
```
adekbadek: Nit: the translator comment here repeats `/* translators: archive page */` from the entries above, but "Posts Page" is WP-specific terminology. A more descriptive hint like `/* translators: the "Posts page" configured under Settings > Reading when a static front page is set. */` would help translators.

adekbadek (approved):
Follow-up to my earlier review. QA'd end-to-end against `newspack-theme` with four popup fixtures (new defaults, legacy meta = `['category']`, legacy no meta row, only `['home']`) on both the blog posts page and a category archive. All scenarios render as expected – the `register_meta` + `wp_parse_args` legacy fallback kicks in correctly for prompts with no saved meta row.

Approving, but flagging inline 1 as a pre-merge ask – worth hardening the two-default relationship before this ships.

includes/class-newspack-popups-model.php
laurelfulford: Good call! Updated in [350bcad](https://github.com/Automattic/newspack-popups/pull/1551/commits/350bcad3c6a536f29fa766d723769050482ebccb).

tests/test-insertion.php
laurelfulford: Thanks Adam! Both changes have been applied to tests in [7caeeed](https://github.com/Automattic/newspack-popups/pull/1551/commits/7caeeedbe56ab02658cacf1ef486abcde02f55cf).

includes/class-newspack-popups-model.php
laurelfulford: Thanks Adam! Both have been updated in [05d79de](https://github.com/Automattic/newspack-popups/pull/1551/commits/05d79dec6c10f3f6624b0d6611871d2fc9734187), with an extra inline comment snuck in [e63f736](https://github.com/Automattic/newspack-popups/pull/1551/commits/e63f73602a20cfd91503cde05346e46f01ee2b2b) just in case (since it's a different file than the new `get_archive_page_types_meta_default()` function is defined in).

laurelfulford:
Thanks for the review, @adekbadek! All feedback has been applied - I'll go ahead and merge since it's been approved, but let me know if anything else needs tweaking!

PR #1561, opened 2026-05-19
fix(popups): portal overlay lightboxes to wp_footer to escape stacking contexts

adekbadek:
## What

Overlay prompts are emitted inside `.entry-content` via `the_content`. Any ancestor that establishes a stacking context (transform, position+z-index, isolation, will-change, contain — common with custom theme CSS or ad-management plugins) traps the lightbox's `z-index: 99999` local to that context, so siblings can paint over the close button. Symptom observed in the wild: scroll-triggered Center Overlay prompts on customer sites end up partially obscured by the sticky right-rail.

## How

- Queue overlay lightboxes on `the_content` instead of inlining them. Flush at `wp_footer` priority 9 (before `wp_print_footer_scripts` at 20) so they render as direct children of `<body>` and any assets enqueued during overlay-content render still make it into the footer. Inline / above-header / archive placements unchanged.
- Scroll-triggered overlays keep their `#page-position-marker_*` inline at the content position; only the lightbox itself moves. The marker still drives `IntersectionObserver` reveal as before.

## How to test

1. Activate Newspack Campaigns; create a Center Overlay prompt with **scroll** trigger at 0%.
2. Add a temporary mu-plugin that fakes the stacking trap:

   ```php
   <?php
   add_action( 'wp_head', function() { ?>
   <style>
     .entry-content { transform: translateZ(0); position: relative; z-index: 1; }
     .demo-rail { position: fixed; top: 0; right: 0; width: 380px; height: 100vh;
                  background: rgba(220,38,38,.85); color: #fff; padding: 24px;
                  z-index: 2; pointer-events: none; font: 700 24px system-ui; }
   </style>
   <?php } );
   add_action( 'wp_body_open', function() {
     echo '<aside class="demo-rail">Demo rail z-index: 2</aside>';
   } );
   ```

3. Visit any post. Without the fix the rail paints over the prompt's right edge (close X hidden). With the fix the lightbox is a direct `<body>` child at z=99999 and renders above the rail.
4. Devtools sanity check: `document.querySelector('.newspack-lightbox').parentElement.tagName === 'BODY'` and `document.querySelector('[id^=page-position-marker_]').closest('.entry-content') !== null`.

## Tests

197/197 PHPUnit pass. Dedicated tests in `tests/test-overlay-queueing.php`: queueing, cross-injection-point dedupe, marker stays inline, archive scroll-trigger marker emission, flush semantics.

## Notes for review

- Hotfix off `release`; needs forward-port to `trunk` once merged.
- Behavior preserved: dedupe by popup ID across multi-emission paths; specificity-sorted queue order (most specific wins the single visible slot).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/includes/class-newspack-popups-inserter.php
+++ b/includes/class-newspack-popups-inserter.php
@@ -53,6 +53,15 @@ final class Newspack_Popups_Inserter {
 	private static $header_template_part_has_rendered = false;
 
 	/**
+	 * Overlay prompts queued for rendering at wp_footer, keyed by popup ID so a
+	 * popup that would otherwise be emitted from multiple injection points
+	 * (singular content, archive header, block-theme header) only renders once.
+	 *
+	 * @var array<int|string, array>
+	 */
+	private static $queued_overlays = [];
+
+	/**
 	 * Constructor.
 	 */
 	public function __construct() {
@@ -64,6 +73,11 @@ final class Newspack_Popups_Inserter {
 		add_action( 'after_archive_post', [ $this, 'insert_inline_prompt_in_archive_pages' ] );
 		add_filter( 'render_block', [ $this, 'insert_inline_prompt_in_block_theme_archives' ], 10, 3 );
 		add_action( 'wp_before_admin_bar_render', [ $this, 'add_preview_toggle' ] );
+
+		// Render overlay prompts as direct children of <body> so they escape any
+		// ancestor stacking context (transformed parents, ad wrappers, etc.) that
+		// would otherwise trap their z-index below sibling content.
+		add_action( 'wp_footer', [ $this, 'print_queued_overlays' ], 100 );
 
 		// Always enqueue scripts, since this plugin's scripts are handling pageview sending via GTAG.
 		add_action( 'wp_enqueue_scripts', [ $this, 'enqueue_scripts' ] );
@@ -456,13 +470,20 @@ final class Newspack_Popups_Inserter {
 			}
 		}
 
-		// 4. Insert overlay prompts at the top of content.
-		// To leave the existing behavior (prepending each overlay) in place,
-		// we reverse our sorted overlays to ensure the most specific appear
-		// first in the DOM, and get priority for the single available slot.
-		$overlay_popups = array_reverse( self::sort_overlays_by_specificity( $overlay_popups ) );
-		foreach ( $overlay_popups as $overlay_popup ) {
-			$output = '<!-- wp:html -->' . Newspack_Popups_Model::generate_popup( $overlay_popup ) . '<!-- /wp:html -->' . $output;
+		// 4. Queue overlay prompts for footer rendering. Reverse the specificity
+		// sort so the most specific overlay is queued first and wins the single
+		// visible slot, matching the previous "prepend to content" DOM order.
+		// Scroll-triggered overlays carry a page-position marker which must
+		// remain inline in `.entry-content` so its percentage offset resolves
+		// against the article column (the marker drives IntersectionObserver
+		// reveal of the lightbox). Emit the marker inline and queue only the
+		// lightbox itself for the footer.
+		foreach ( array_reverse( self::sort_overlays_by_specificity( $overlay_popups ) ) as $overlay_popup ) {
+			self::queue_overlay( $overlay_popup );
+			$marker = Newspack_Popups_Model::generate_position_marker( $overlay_popup );
+			if ( '' !== $marker ) {
+				$output = '<!-- wp:html -->' . $marker . '<!-- /wp:html -->' . $output;
+			}
 		}
 		return $output;
 	}
@@ -564,10 +585,54 @@ final class Newspack_Popups_Inserter {
 				return Newspack_Popups_Model::should_be_inserted_in_page_content( $popup ) && Newspack_Popups_Model::is_overlay( $popup );
 			}
 		);
-		$popups = self::sort_overlays_by_specificity( array_values( $popups ) );
-		foreach ( $popups as $popup ) {
-			echo Newspack_Popups_Model::generate_popup( $popup ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+		foreach ( self::sort_overlays_by_specificity( array_values( $popups ) ) as $popup ) {
+			self::queue_overlay( $popup );
 		}
+	}
+
+	/**
+	 * Queue an overlay popup for rendering at wp_footer.
+	 *
+	 * Overlays are deduped by popup ID, so the same overlay queued from multiple
+	 * injection points (singular content + above-header, archive header, etc.)
+	 * only renders once. Queue order is preserved, so the existing specificity
+	 * sort done by callers still controls which overlay wins the single visible
+	 * slot.
+	 *
+	 * @param array $popup Popup data as returned by Newspack_Popups_Model.
+	 */
+	private static function queue_overlay( $popup ) {
+		if ( empty( $popup['id'] ) ) {
+			return;
+		}
+		$id = $popup['id'];
+		if ( ! isset( self::$queued_overlays[ $id ] ) ) {
+			self::$queued_overlays[ $id ] = $popup;
+		}
+	}
+
+	/**
+	 * Print all queued overlay popups as direct children of <body>, via wp_footer.
+	 *
+	 * This decouples overlay output from any ancestor stacking context that the
+	 * popup would otherwise inherit when emitted inside post content or above
+	 * the site header — for example a transformed/scaled wrapper, a sticky ad
+	 * container, or any element with isolation:isolate. Inline popups remain
+	 * inline; only the overlay-typed placements are portaled.
+	 */
+	public static function print_queued_overlays() {
+		if ( empty( self::$queued_overlays ) ) {
+			return;
+		}
+		foreach ( self::$queued_overlays as $popup ) {
+			// Suppress the scroll-trigger page-position marker: for scroll-triggered
+			// overlays it was emitted inline at the content position so its
+			// percentage offset still resolves against `.entry-content`. For time-
+			// triggered overlays there is no marker to begin with. Emitting the
+			// marker here (at body level) would break scroll-trigger detection.
+			echo Newspack_Popups_Model::generate_popup( $popup, false ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+		}
+		self::$queued_overlays = [];
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


includes/class-newspack-popups-inserter.php
```diff
@@ -65,6 +87,15 @@ public function __construct() {
...
 
+		// Flush queued overlay prompts at wp_footer. Echoing from a wp_footer
+		// callback (with no surrounding container in the callback itself) lands
+		// the markup as a direct child of <body>, which is what lets it escape
+		// any ancestor stacking context that would otherwise trap its z-index.
+		// Priority 9 runs *before* wp_print_footer_scripts (priority 20), so any
+		// asset that an overlay's content block / shortcode enqueues at render
+		// time still makes it into the page's footer scripts.
+		add_action( 'wp_footer', [ __CLASS__, 'print_queued_overlays' ], 9 );
+
```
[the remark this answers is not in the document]
adekbadek: Good catch – the PR description was stale from the earlier draft. Updated the body to reflect priority 9 (intentionally before `wp_print_footer_scripts` at 20 so any asset that an overlay's content block / shortcode enqueues at render time still makes it into the footer).

includes/class-newspack-popups-inserter.php
```diff
@@ -564,12 +602,77 @@ function ( $popup ) {
 				return Newspack_Popups_Model::should_be_inserted_in_page_content( $popup ) && Newspack_Popups_Model::is_overlay( $popup );
 			}
 		);
-		$popups = self::sort_overlays_by_specificity( array_values( $popups ) );
-		foreach ( $popups as $popup ) {
-			echo Newspack_Popups_Model::generate_popup( $popup ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+		foreach ( self::sort_overlays_by_specificity( array_values( $popups ) ) as $popup ) {
+			self::queue_overlay( $popup );
+			echo self::emit_position_marker_inline( $popup ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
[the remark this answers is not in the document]
adekbadek: Fixed in 86a26b1 – `emit_position_marker_inline()` now returns raw HTML, and only the singular-content callsite (where the result is prepended to `the_content` output that may go through a downstream block parser) wraps in `<!-- wp:html -->`. The echoed archive and block-theme-header paths no longer leak the block-serialization comments.

src/view/style.scss
```diff
@@ -395,43 +395,42 @@ $width__tablet: 782px;
...
+		}
+
+		@media only screen and ( min-width: $width__mobile ) {
+			.alignfull.wp-block-columns {
+				margin-left: -16px;
+				margin-right: -16px;
+				max-width: calc(100% + 32px);
+				width: calc(100% + 32px);
+			}
 		}
```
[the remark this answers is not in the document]
adekbadek: Fixed in 86a26b1 – dropped the standalone `@media` block. The inner `&.wp-block-columns` rule under `.alignfull, .alignwide` already produces `.alignfull.wp-block-columns` (and `.alignwide.wp-block-columns`).

tests/test-overlay-queueing.php
```diff
@@ -0,0 +1,386 @@
...
+	 * The dedupe map is shared across injection points: queueing the same
+	 * overlay via `insert_popups_in_post_content` (singular content path) AND
+	 * `insert_before_header_in_template_part` (block-theme above-header path)
+	 * must still result in a single emission.
+	 */
+	public function test_dedupe_across_injection_points() {
+		$overlay_popup = self::create_overlay_popup_object( 'Cross-path overlay', 'time' );
+		// Mark the popup as "above_page_header" so the block-theme path queues it.
+		Newspack_Popups_Model::set_popup_options( $overlay_popup['id'], [ 'trigger_type' => 'time' ] );
+		$expected_id = 'id="' . Newspack_Popups_Model::canonize_popup_id( $overlay_popup['id'] ) . '"';
```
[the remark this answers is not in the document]
adekbadek: Fixed in 86a26b1 – dropped the redundant `set_popup_options` call and rewrote the docblock to match what the test actually exercises (queueing the popup from two paths, asserting the dedupe contract on `queue_overlay()` itself).

includes/class-newspack-popups-inserter.php
```diff
@@ -564,10 +606,76 @@ function ( $popup ) {
...
+	 * position (see {@see emit_position_marker_inline()}).
+	 */
+	public static function print_queued_overlays(): void {
+		if ( empty( self::$queued_overlays ) ) {
+			return;
+		}
+		foreach ( self::$queued_overlays as $popup ) {
+			echo Newspack_Popups_Model::generate_popup( $popup, false ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+		}
+		self::$queued_overlays = [];
```
dkoo, line 684: ### 🔴 `$emitted_markers` is never reset, while `$queued_overlays` is

`print_queued_overlays()` does `self::$queued_overlays = []` after flushing but never clears `self::$emitted_markers`, despite the docblock calling the two maps "parallel." If the same popup ID is re-queued after a flush in one request (preview/tooling, a manual `apply_filters('the_content', …)`, or certain render flows), the overlay re-queues and re-flushes but its scroll marker is suppressed (`emit_position_marker_inline()` returns `''`) → **a lightbox with no marker → scroll-triggered overlay never reveals** (`segmentation.js` finds no `page-position-marker_*` to observe). 

Suggested fix: drain `$emitted_markers` in `print_queued_overlays()` too (or reset both wherever `$the_content_has_rendered` resets).

includes/class-newspack-popups-inserter.php
```diff
@@ -65,6 +87,15 @@ public function __construct() {
...
 		add_action( 'wp_before_admin_bar_render', [ $this, 'add_preview_toggle' ] );
 
+		// Flush queued overlay prompts at wp_footer. Echoing from a wp_footer
+		// callback (with no surrounding container in the callback itself) lands
+		// the markup as a direct child of <body>, which is what lets it escape
+		// any ancestor stacking context that would otherwise trap its z-index.
+		// Priority 9 runs *before* wp_print_footer_scripts (priority 20), so any
+		// asset that an overlay's content block / shortcode enqueues at render
+		// time still makes it into the page's footer scripts.
+		add_action( 'wp_footer', [ __CLASS__, 'print_queued_overlays' ], 9 );
```
dkoo, line 97: ### 🟡 Non-blocking edge case regressions: rendering now depends on `wp_footer` firing after `the_content`

Previously the overlay markup lived in the returned content string, so it rendered wherever the content did. Now the content pass only queues; the lightbox renders only if `wp_footer` fires later. For a standard singular template this is always safe. Risk cases: a theme that omits `wp_footer()`, a path that `exit`s after content but before footer, and the archive (`after_header`) / above-header (`wp_body_open`) paths that queue independently of the `$the_content_has_rendered` guard. A vanished overlay is a worse, *silent* regression than the bug being fixed. Worth confirming (or a defensive `shutdown` fallback flush).

Also affects overlays rendered in archive pages: overlay content used to render at `after_header` (before the loop); it now renders at `wp_footer` (after the loop), so global `$post` / `in_the_loop()` differ. Dynamic, post-context-dependent blocks inside an overlay could bind to a different post. Low practical impact (marketing/donation overlays rarely embed post-context dynamic blocks), and for singular posts global `$post` is unchanged — but worth a sanity check if any publisher puts dynamic blocks in overlay prompts.

Flagging because this change introduces potential regressions in the above scenarios, but considering non-blocking since they're edge cases.

includes/class-newspack-popups-model.php
```diff
@@ -1085,13 +1085,60 @@ public static function get_current_popup() {
...
+	public static function generate_position_marker( array $popup ): string {
+		if ( ! self::is_overlay( $popup ) ) {
+			return '';
+		}
+		$trigger_type = $popup['options']['trigger_type'] ?? '';
+		if ( 'scroll' !== $trigger_type ) {
+			return '';
+		}
+		$element_id = self::canonize_popup_id( $popup['id'] );
+		$progress   = absint( $popup['options']['trigger_scroll_progress'] ?? 0 );
```
dkoo, line 1114: ### 🟢 Non-blocking nit: Two divergent marker generators

`generate_position_marker()` uses `absint()`+`%d`; the legacy inline branch in `generate_popup()` ([`model.php:1255`](repos/newspack-popups/includes/class-newspack-popups-model.php:1255)) uses raw `esc_attr( …trigger_scroll_progress )`. Same output for valid ints, but two sources of truth — consider having the inline branch call the new helper.

dkoo (changes requested):
@adekbadek looks good, passes manual testing. Just one blocker and two non-blocking comments inline.

includes/class-newspack-popups-inserter.php
adekbadek: Fixed in 74892c9 – `print_queued_overlays()` now drains `$emitted_markers` alongside `$queued_overlays` so a re-queue after a flush re-emits both the lightbox and the inline marker. Added a regression test that fails without the marker drain (the assertion the comment described: scroll overlay re-queued post-flush ends up marker-less in the returned content).

includes/class-newspack-popups-inserter.php
adekbadek: Acknowledged – leaving as-is for this hotfix. For the singular path the dependency on `wp_footer` is bounded by the standard template lifecycle, and the archive (`after_header`) / above-header (`wp_body_open`) paths now both queue and rely on the same flush, so a theme that doesn't fire `wp_footer` would have been broken under the inline rendering too (admin-bar, scripts, the existing scroll-tracking JS all depend on it). Worth revisiting with a `shutdown` fallback if a publisher reports a vanished overlay, but adding it speculatively here would be belt-and-suspenders for a case the existing footer-dependent surface doesn't already cover. Re: dynamic post-context blocks on archive overlays – will keep an eye on it; not aware of any publisher embedding context-bound dynamic blocks in overlay prompts today.

includes/class-newspack-popups-model.php
adekbadek: Fixed in 74892c9 – the inline branch in `generate_popup()` now calls `generate_position_marker()`, giving a single source of truth for the marker markup (and incidentally swapping the raw `esc_attr` on `trigger_scroll_progress` for the helper's `absint`/`%d` formatting).

adekbadek:
@dkoo Thanks for the review – addressed the three inline threads in 74892c9:

- **🔴 `$emitted_markers` is never reset** – `print_queued_overlays()` now drains both maps together, so a re-queue after a flush re-emits both the lightbox and the inline marker. Added a regression test (`test_marker_re_emits_after_flush_and_requeue`) that fails without the marker drain.
- **🟡 `wp_footer`-dependent rendering edge cases** – left as-is for this hotfix. Reasoning in-thread: the existing footer-dependent surfaces (admin-bar, scripts, scroll-tracking JS) all rely on `wp_footer` already, so a theme that omits it would have been broken before this change too; happy to revisit with a `shutdown` fallback if a real regression surfaces.
- **🟢 Two divergent marker generators** – inline branch in `generate_popup()` now calls `generate_position_marker()` for a single source of truth.

dkoo (approved):

adekbadek:
Superseded by the monorepo hotfix: Automattic/newspack-workspace#194.

Newspack releases now come from the monorepo, and the daily legacy sync only mirrors `trunk` – so merging here wouldn't ship as a hotfix. The change has been ported verbatim under `plugins/newspack-popups/` and opened against `release` there, which cuts the out-of-schedule `newspack-popups` patch and back-merges to `main`/`alpha` automatically (no separate trunk forward-port needed).

